PR #5, opened 2016-12-15
Add support for chome scopes

jasonLaster:
related to:
https://github.com/devtools-html/debugger.html/pull/1435


jasonLaster:
This also fixes support for chrome-remote-interface

Issue #1, opened 2016-12-15
[Local Toolbox] Rename to Launch Pad


jasonLaster:
* rename directory
* rename package (can we change it on npm)?

PR #6, opened 2016-12-16
rename devtools-local-toolbox to devtools-launchpad

juliandescottes:
cc @jasonLaster 

We discussed about renaming devtools-local-toobox to launchpad, so here's a first PR that does just that.

Not sure if we want to reset the versioning or keep going. We'll need to publish the package on npm and update the debugger afterwards of course.


juliandescottes:
cf issue #1

juliandescottes:
> rename package (can we change it on npm)?

From what I've seen (http://stackoverflow.com/questions/28371669/renaming-a-published-npm-module) we can only deprecate and not rename.

PR #12, opened 2016-12-16
add back eslint to launchpad

jasonLaster:


juliandescottes:
I still find it weird to define these dependencies in launchpad since launchpad does not provide any eslintrc using the plugins added, nor has an eslint command in its package.json :)

jasonLaster:
Yea true.

Hmm, I wanted to make the debugger work, but I don't know if we want all
tools to share the same config.

I guess it's safe to assume they will likely share these plugins.

Let's talk next week

PR #13, opened 2016-12-16
Add storybook to test LandingPage component

juliandescottes:
I started a storybook integration but faced two issues so let's discuss.

1. had to move the Redux related code out of the LandingApp component. Using storybook with Redux seems feasible but complicated, not sure it's worth the investment
2. feature.js would fail on Chrome when trying to export its methods (getValue is undefined). Not sure what's up with that, might be linked to the default webpack config used by storybook

cc @jasonLaster ; will try to investigate more the last issue mentioned above

### DIFF
--- a/packages/devtools-launchpad/src/components/LandingPage.js
+++ b/packages/devtools-launchpad/src/components/LandingPage.js
@@ -1,8 +1,5 @@
 const React = require("react");
-const { connect } = require("react-redux");
 const classnames = require("classnames");
-const { getTabs } = require("../selectors");
-const { getValue } = require("devtools-config");
 
 require("./LandingPage.css");
 const { DOM: dom } = React;
@@ -149,11 +146,4 @@ const LandingPage = React.createClass({
   }
 });
 
-module.exports = connect(
-  state => ({
-    tabs: getTabs(state),
-    supportsFirefox: getValue("firefox"),
-    supportsChrome: getValue("chrome"),
-    title: getValue("title"),
-  })
-)(LandingPage);
+module.exports = LandingPage;
--- a/packages/devtools-launchpad/src/components/LandingPage.js
+++ b/packages/devtools-launchpad/src/components/LandingPage.js
@@ -31,7 +31,10 @@ function getTabURL(tab, paramName) {
 
 const LandingPage = React.createClass({
   propTypes: {
-    tabs: ImPropTypes.map.isRequired
+    tabs: ImPropTypes.map.isRequired,
+    supportsFirefox: React.PropTypes.bool.isRequired,
+    supportsChrome: React.PropTypes.bool.isRequired,
+    title: React.PropTypes.string.isRequired,
   },
 
   displayName: "LandingPage",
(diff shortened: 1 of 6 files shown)


jasonLaster:
Yup. That sounds frustrating.

I think the key is to just show the landing page with storybook.

We should pull feature out of the component and pass the values in as props
so it's testable.

Best thing to do with storybook is to imagine your writing a test

juliandescottes:
I pulled the dependency on devtools-config/getValue out of the landing page component. Let me know what you think. 

We could add a "LaunchpadApp" intermediary between index.js and LandingPage.js, that would basically just call connect on the LandingPage component and pass the props, but I feel like it'd be overkill here.

packages/devtools-launchpad/src/components/LandingPage.js
```diff
@@ -154,6 +152,4 @@ const LandingPage = React.createClass({
   }
 });
 
-module.exports = connect(
-  state => ({ tabs: getTabs(state) })
-)(LandingPage);
+module.exports = LandingPage;
```
jasonLaster: You should be able to keep this here and reference the wrappedCompondnt or something like that

juliandescottes:
> I pulled the dependency on devtools-config/getValue out of the landing page component. Let me know what you think.

Would have been nice if I actually had pushed my changes ... Just updated the PR, sorry for pinging you too early.

packages/devtools-launchpad/src/index.js
```diff
@@ -104,7 +105,19 @@ function bootstrap(React, ReactDOM, App, appActions, appStore) {
   }
 
   const { store, actions, LandingPage } = initApp();
-  renderRoot(React, ReactDOM, LandingPage, store);
+
+  let landingPage = connect(
```
jasonLaster: this should still be in the `LandingPage` component.

you can do something like 
`const { wrappedComponent: LandingPage } = require("components/LandingPage")`

juliandescottes: Thanks! Indeed it does fix my issue with connect!

juliandescottes:
Adding back the dependency on "devtools-config" in LandingPage.js means the issue with feature.js on Chrome is back.

So if we land this as is, storybook only works in Firefox. I'll take a look at the generated bundle later to see if I can understand what's wrong here.

juliandescottes:
@jasonLaster sorry about the delay, it got a bit busy during the previous days. 
I added a LaunchpadApp component to cleanup the LandingPage. It can now be tested properly on Firefox and Chrome.

jasonLaster:
@juliandescottes I made some small tweaks so that it would be more react idiomatic.

Feel free to merge and or make other tweaks.

This is pretty great.

screen shot 2016-12-24 at 11 16 21 am

Lets also look at adding some automated testing here as well
https://getstorybook.io/docs/react-storybook/testing/react-ui-testing

The biggest advantage is that we don't run the risk of regressing on these stories

juliandescottes:
Looks great @jasonLaster thanks! Merging.

nchevobbe:
@juliandescottes do you have things to do here or can we merge the PR ?

juliandescottes:
@nchevobbe no sorry ... pretty sure I clicked on merge, but looks like I missed the button ...

nchevobbe:
No harms, the change looks great and will make it easy to hack on the landing page

PR #14, opened 2016-12-16
use smaller footer for landing page

juliandescottes:
The current footer of the landing page is too big and makes it hard to use when there are many tabs.

Before: 

image

After 

image


jasonLaster:
💥

PR #28, opened 2016-12-23
Better styling on the landing page

nchevobbe:
We simplify the rendered markup and some of the logic of the render function.
This allows us to also simplify the CSS, by mostly using the capabilities of
flexbox to have the same UI as we had before (no absolute positionning).
We also add some custom properties so the styling is more consistent.

### DIFF
--- a/packages/devtools-launchpad/src/components/LandingPage.js
+++ b/packages/devtools-launchpad/src/components/LandingPage.js
@@ -42,7 +42,7 @@ const LandingPage = React.createClass({
     };
   },
 
-  renderTabs(tabTitle, tabs, paramName) {
+  renderTabs(tabs, paramName) {
     if (!tabs || tabs.count() == 0) {
       return dom.div({}, "");
     }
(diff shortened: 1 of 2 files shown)


packages/devtools-launchpad/src/components/LandingPage.js
```diff
@@ -65,49 +65,41 @@ const LandingPage = React.createClass({
...
-      firstTimeMessage("Node", "node")
-    );
-  },
-
   renderPanel() {
-    const panels = {
-      Firefox: this.renderFirefoxPanel,
-      Chrome: this.renderChromePanel,
-      Node: this.renderNodePanel
+    const configMap = {
```
jasonLaster: lets move the configMap to the module context

packages/devtools-launchpad/src/components/LandingPage.js
```diff
@@ -122,7 +114,7 @@ const LandingPage = React.createClass({
       connections.push("Chrome", "Node");
     }
 
-    return dom.div(
+    return dom.aside(
```
jasonLaster: 👍

packages/devtools-launchpad/src/components/LandingPage.js
nchevobbe: Sure

nchevobbe:
Could you add me as a contributor on the project so I can merge PRs when you're not around ? Thanks !

packages/devtools-launchpad/src/components/LandingPage.js
nchevobbe: @jasonLaster I forgot to do this. I'll put it in a PR with other some small improvements. 
What do you mean by module context here ? this file https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/index.js ?

jasonLaster:
Hmm not sure

PR #30, opened 2016-12-25
Fix styling issues

nchevobbe:
There was still some issues when : 

- Tabs title were too long
- There wasn't any tabs to show

This PR should fix that.

I removed the 500px style height on the container since it could add a scrollbar that may be confusing to the dev (I originally thought something was wrong with the component because my window was smaller than 500px height). 
Feel free to tell me if I'm wrong on this and if there was a good reason to have it :).

Also, I updated yarn.lock, as recommended [here](https://yarnpkg.com/en/docs/yarn-lock), but I don't know if it's what we want.


juliandescottes:
Looks great!  I initially added the 500px to reproduce a style issue, no problem to remove it now!

PR #31, opened 2016-12-27
Add collapse logic to vertical splitbox.

wldcordeiro:
This adds the collapse logic to the vertical `Splitbox`

### DIFF
--- a/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
+++ b/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
@@ -176,15 +176,18 @@ const SplitBox = React.createClass({
         width: endPanelCollapsed ? 0 : endWidth
       };
     } else {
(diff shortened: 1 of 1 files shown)


wldcordeiro:
Weird that we don't require review here, also I seem to lack team access?

juliandescottes:
@wldcordeiro added mandatory reviews and added you as a collaborator.

wldcordeiro:
@jasonLaster I'm still trying to figure out this collapse. Height is weird, setting it to zero doesn't do the trick since the children elements have height.

jasonLaster:
@wldcordeiro I think I figured out what to do here. 

Let's punt! I don't think we need to vertically collapse the bottom Panè honestly. 

There's plenty of wins without it :)

jasonLaster:
closing this, we can always come back to it

PR #33, opened 2016-12-27
Add a tab filter input

### LINKED ISSUE #25: Add a way to filter tabs in landing page
If you have a lot of tabs (like I do), it could be difficult to find the one you want in the list

Bonus point:
- <kbd>Enter</kbd>keydown automatically select the tab if there is only one (so you can debug/interact with the tab you want really quickly)

nchevobbe:
Fixes #25 

This patch adds an input aside to the "target" label and allow the user to filter out the tabs shown.
That's done the Redux way, hence the patch adds constant, selector and reducer to handle the filtering.
It uses the fuzzaldrin-plus package to handle the filtering (same as the debugger).

This is how it looks : 
capture du 2016-12-27 12-03-57

The thing I'm not confident with is the change in stories. I added a way to create the redux state as the app does and get the tabs and the filteringText from there. I did this to make sure the filtering was working, but maybe there's a better way I didn't thought about ?

### DIFF
--- a/packages/devtools-launchpad/src/components/LaunchpadApp.js
+++ b/packages/devtools-launchpad/src/components/LaunchpadApp.js
@@ -1,13 +1,16 @@
 const React = require("react");
 const { PropTypes } = React;
 const { connect } = require("react-redux");
-const { getTabs } = require("../selectors");
+const { bindActionCreators } = require("redux");
+const { getTabs, getFilterString } = require("../selectors");
 const { getValue } = require("devtools-config");
 const LandingPage = React.createFactory(require("./LandingPage"));
 
 const LaunchpadApp = React.createClass({
   propTypes: {
-    tabs: PropTypes.array
+    tabs: PropTypes.array,
+    filterString: PropTypes.string,
+    actions: PropTypes.array
   },
 
   displayName: "LaunchpadApp",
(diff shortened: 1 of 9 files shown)


packages/devtools-launchpad/src/components/LaunchpadApp.js
```diff
@@ -1,13 +1,16 @@
...
-const { getTabs } = require("../selectors");
+const { bindActionCreators } = require("redux");
+const { getTabs, getFilterString } = require("../selectors");
 const { getValue } = require("devtools-config");
 const LandingPage = React.createFactory(require("./LandingPage"));
 
 const LaunchpadApp = React.createClass({
   propTypes: {
-    tabs: PropTypes.array
+    tabs: PropTypes.array,
```
nchevobbe: Good catch

nchevobbe:
> One UI detail. The header looks off with the filter bar to me. Perhaps we just have the filter bar with a placeholder...

I agree. Maybe we can change the style of the active item in the sidebar, so we don't have to put a label on the header, like this : 

capture d ecran 2016-12-27 a 17 02 21

jasonLaster:
Good call. Not sure about the blue border tho :)

nchevobbe:
here's without it

capture d ecran 2016-12-27 a 17 06 55

I had the feeling it gives us a link between the tabs on the left and the content on the right (like a tab bar), but I'm not strongly opinionated about this

juliandescottes:
I like it! 
Maybe a smaller font for the filter input now that there's no title next to it ? 
(same font size as  the tabs for instance?)

nchevobbe:
> I like it!
> Maybe a smaller font for the filter input now that there's no title next to it ?
> (same font size as the tabs for instance?)

Sure. I also copied the style we have in the debugger when we have a selected file. The background of the file is blue, the border between the file list and the main content is grey. I think it works pretty good

capture d ecran 2016-12-27 a 17 23 32

jasonLaster:
💃 

Really cool

PR #34, opened 2016-12-27
Multiple style bundles

juliandescottes:
cf. https://github.com/devtools-html/debugger.html/pull/1445#issuecomment-269310411 

If a tool using launchpad is requiring css files from several bundles, only one styles.css will be created, containing the CSS files from the last entry that got loaded.

I think we should both
* allow tools to provide an exclude regexp, to avoid including unnecessary styles in the bundle
* add the bundle name to the styles bundle, to avoid losing time investigating this kind of issues

cc @jasonLaster

### DIFF
--- a/packages/devtools-launchpad/webpack.config.js
+++ b/packages/devtools-launchpad/webpack.config.js
@@ -113,7 +113,7 @@ module.exports = (webpackConfig, envConfig) => {
       loader: ExtractTextPlugin.extract("style-loader", "css-loader", "postcss-loader")
     });
 
-    webpackConfig.plugins.push(new ExtractTextPlugin("styles.css"));
+    webpackConfig.plugins.push(new ExtractTextPlugin("[name].css"));
   }
 
   webpackConfig.postcss = () => [require("postcss-bidirection")];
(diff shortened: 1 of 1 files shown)


packages/devtools-launchpad/webpack.config.js
```diff
@@ -106,10 +106,14 @@ module.exports = (webpackConfig, envConfig) => {
...
       test: /\.css$/,
+      exclude: request => {
+        // If the tool defines an exclude regexp for CSS files.
+        return webpackConfig.cssExcludes && request.match(webpackConfig.cssExcludes);
+      },
       loader: ExtractTextPlugin.extract("style-loader", "css-loader", "postcss-loader")
     });
 
-    webpackConfig.plugins.push(new ExtractTextPlugin("styles.css"));
+    webpackConfig.plugins.push(new ExtractTextPlugin("styles-[name].css"));
```
jasonLaster: Why not just use the bundle name? CSS kinda makes it clear :)

juliandescottes:
Thanks for the feedback! Merging.

PR #40, opened 2016-12-28
Better keyboard navigation

nchevobbe:
Add autoFocus on filter tab to make it faster to search for a specific tab.
Add tabIndex and :focus styling  on the relevant items (sidebar items, tab items) to make it easy to navigate with the keyboard in the app.

If there is only one tab (either because there is really only one, or because we filtered them out, pressing
<kbd>Enter</kbd> starts the debugging on the tab.

I'm not sure about https://github.com/devtools-html/devtools-core/commit/761e75789efcdd796bc86862a0944a7b3963c871 , which automatically focus the filter bar when clicking on a sidebar element, tell me what you think about this @jasonLaster & @juliandescottes 

Here's how keyboard navigation looks : 

kbd-nav

And navigationg when there is only one tab : 

kbd-nav-single-tab

### DIFF
--- a/packages/devtools-launchpad/src/components/LandingPage.js
+++ b/packages/devtools-launchpad/src/components/LandingPage.js
@@ -63,13 +63,20 @@ const LandingPage = React.createClass({
       return dom.div({}, "");
     }
 
+    let tabClassNames = ["tab"];
+    if (tabs.size === 1) {
+      tabClassNames.push("active");
+    }
+
     return dom.div(
       { className: "tab-group" },
       dom.ul(
         { className: "tab-list" },
         tabs.valueSeq().map(
           tab => dom.li({
-            className: "tab",
+            className: classnames("tab", {
+              active: tabs.size === 1
+            }),
             key: tab.get("id"),
             tabIndex: 0,
             role: "link",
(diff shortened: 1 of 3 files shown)


packages/devtools-launchpad/src/components/LandingPage.js
```diff
@@ -45,32 +45,59 @@ const LandingPage = React.createClass({
     };
   },
 
+  componentDidUpdate() {
```
jasonLaster: this might be better as `componentDidMount` but it's fine for now

packages/devtools-launchpad/src/components/LandingPage.js
```diff
@@ -45,32 +45,59 @@ const LandingPage = React.createClass({
...
     return dom.div(
       { className: "tab-group" },
       dom.ul(
         { className: "tab-list" },
-        tabs.valueSeq().map(tab => dom.li(
-          { "className": "tab",
-            "key": tab.get("id"),
-            "onClick": () => this.props.onTabClick(getTabURL(tab, paramName))
+        tabs.valueSeq().map(
+          tab => dom.li({
```
jasonLaster: at this point, i'd generally move this to a new function `renderTab`

packages/devtools-launchpad/src/components/LandingPage.js
```diff
@@ -45,32 +45,59 @@ const LandingPage = React.createClass({
...
+            className: classnames({
```
jasonLaster: classnames supports this syntax: `classnames("tabs", {active: tabs.size === 1})`

packages/devtools-launchpad/src/components/LandingPage.js
nchevobbe: The first render is handled by the `autoFocus` attribute on the input. I thought `componentDidMount` was only fired after the first render. also, this wouldn't pick the "update" of the state we do when we change of panel.

packages/devtools-launchpad/src/components/LandingPage.js
nchevobbe: cool :+1:

PR #44, opened 2016-12-28
This makes DIR configurable

jasonLaster:
#43

Issue #27, opened 2016-12-23
[LaunchPad] Make the launchpad title configurable


nchevobbe:
It is hardcoded to "Firefox Debugger" ATM (https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/index.html#L4), which can be a bit weird for other tool using it (like devtools-reps)

jasonLaster:
Yup! Great point, perhaps we also want to make the favicon configurable too!

juliandescottes:
Two small things:

1.  The inspector is not using the index.html from launchpad at the moment. Instead it sets `development.customIndex = true` in its launchpad config and defines its own route to serve "/". One option would be to say that each tool is responsible providing its own index.html and define a route for it.

2. After merging #34, the style bundle is no longer styles.css but [bundle-name].css which means that the following snippet should also be customized for each tool: 
```
{{^isDevelopment}}
<link rel="stylesheet" href="assets/build/styles.css" type="text/css" />
{{/isDevelopment}}
```

PR #45, opened 2016-12-29
[WIP] Make title and favicon configurable

nchevobbe:
Also changed the launchpad default favicon (to the unicode rocket emoji).
This works for the title, but not for the favicon yet.

I don't know jack about Express so I just hacked a little to see how I can done that.
I was hoping to declare the containing folder of the favicon as a static assets so I can use the favicon name directly but I face some problems:

1. I don't see how I can declare a relative path to the favicon in the `development.json` file (in Reps for example), since express uses the `devtools-launchpad` path.
2. Even if I use an absolute path in the config, I can't get the favicon to be displayed. The weird thing is that I can see it in `localhost:8000/favicon.png` though :confused: 

I'll dig this more, just wanted to know if there were better strategies here.

Also this would conflict with #44 , so I'm waiting it to be merged so I can rebase on top of it.


packages/devtools-launchpad/bin/development-server.js
```diff
@@ -35,7 +35,13 @@ function httpGet(url, onResponse) {
 function serveRoot(req, res) {
   const tplPath = path.join(__dirname, "../index.html");
   const tplFile = fs.readFileSync(tplPath, "utf8");
-  res.send(Mustache.render(tplFile, { isDevelopment: isDevelopment() }));
+  res.send(Mustache.render(tplFile, {
+    isDevelopment: isDevelopment(),
+    pageTitle: getValue("pageTitle") || getValue("title"),
```
jasonLaster: Why two titles?

packages/devtools-launchpad/bin/development-server.js
```diff
@@ -35,7 +35,13 @@ function httpGet(url, onResponse) {
...
+    favicon: getValue("assets.favicon")
+      ? path.basename(getValue("assets.favicon"))
+      : "launchpad-favicon.png"
```
jasonLaster: Hmm not sure about this one. Perhaps tools add their favs in the same launchpad dir :)

packages/devtools-launchpad/bin/development-server.js
```diff
@@ -91,6 +97,14 @@ function startDevServer(devConfig, webpackConfig) {
   const app = express();
   app.use(express.static("assets/build"));
   app.use(express.static(path.join(__dirname, '../assets')));
+
+  let assets = getValue("assets");
+  if (assets) {
+    Object.keys(assets).forEach(key => {
+      app.use(express.static(path.dirname( assets[key])));
+    });
```
jasonLaster: Hmm. What does this do?

packages/devtools-launchpad/bin/development-server.js
juliandescottes: Or we keep href="favicon.png" in index.html and the tools just define custom routes to point to their local favicon ?

nchevobbe:
@juliandescottes makes some good points in https://github.com/devtools-html/devtools-core/issues/27#issuecomment-269775976 .

packages/devtools-launchpad/bin/development-server.js
jasonLaster: yea - that's fair

jasonLaster:
Hmm, that *is* a good point. I think I prefer adding more config variables for the template. Either way, that probably helps us as we're treating the index as a mustache template.

This stuff is never great :P

jasonLaster:
ping @nchevobbe lets try and get this in soon.

nchevobbe:
@jasonLaster sure, thanks for the reminder

nchevobbe:
@juliandescottes @jasonLaster I just pushed an update to this PR. Since I wanted the favicon to be fully configurable (and not dependant of where the devtools-launchpad modules is in the tree-dependency), I added a `rootDir` argument to the `startDevServer` function, so we can access it in devtools-launchpad. Which makes possible to declare in the configuration file a path to a favicon, wherever it is in the repo.

I successfully tested it in devtools-rep (with yarn link), with and without favicon configuration specified.

Here's what the configuration looks like : 
```json
  "title": "Firefox Devtools Reps",
  "favicon": "src/launchpad/assets/reps-favicon.png",
```

Feel free to tell me if you think that it's a bad idea :) I'm not used to things like that.

PR #49, opened 2016-12-31
[WIP] Flow types

clarkbw:
I'll be adding more to this soon, but wanted to get the initial layout up here.


packages/devtools-client-adapters/src/firefox/commands.js
```diff
@@ -71,38 +98,39 @@ function onNewBreakpoint(location, res) {
...
-  });
+    actualLocation
+  };
 }
 
-function removeBreakpoint(breakpointId) {
+function removeBreakpoint(breakpointId: BreakpointId) {
   const bpClient = bpClients[breakpointId];
-  bpClients[breakpointId] = null;
+  delete bpClients[breakpointId];
```
clarkbw: flow did not like setting this to null so I changed to delete, we should really move to using a `Set` here though.

packages/devtools-client-adapters/src/firefox/commands.js
```diff
@@ -71,38 +98,39 @@ function onNewBreakpoint(location, res) {
...
 }
 
-function evaluate(script, { frameId }) {
-  const params = frameId ? { frameActor: frameId } : {};
-  const deferred = defer();
-
-  tabTarget.activeConsole.evaluateJS(script, (result) => {
-    deferred.resolve(result);
-  }, params);
+type EvaluateParam = {
```
clarkbw: I wasn't sure what this object really was so I just gave it this, but we should fill this in later.

packages/devtools-client-adapters/src/firefox/commands.js
```diff
@@ -113,46 +141,46 @@ function debuggeeCommand(script) {
   return Promise.resolve();
 }
 
-function navigate(url) {
+function navigate(url: string): Promise<*> {
```
clarkbw: In the fx/types I created a `URL` type that was `export type URL = string` and I might try doing that here just because it looks better.  Not sure its really worth the effort.

packages/devtools-client-adapters/src/firefox/types.js
```diff
@@ -0,0 +1,285 @@
...
+ * @memberof firefox
+ * @static
+ */
+export type ThreadClient = {
+  resume: (Function) => Promise<*>,
+  stepIn: (Function) => Promise<*>,
+  stepOver: (Function) => Promise<*>,
+  stepOut: (Function) => Promise<*>,
+  breakOnNext: () => Promise<*>,
+  // FIXME: unclear if SourceId or ActorId here
```
clarkbw: These are both just strings, so it doesn't really matter too much.

packages/devtools-client-adapters/src/firefox/types.js
```diff
@@ -0,0 +1,285 @@
...
+  source: ({ actor: SourceId }) => SourceClient,
+  pauseGrip: (Grip|Function) => ObjectClient,
+  pauseOnExceptions: (boolean, boolean) => Promise<*>,
+  interrupt: () => Promise<*>,
+  eventListeners: () => Promise<*>,
+  getFrames: (number, number) => FramesResponse
```
clarkbw: In general I tried to stick to a `$XPacket` for information coming in as a notification or in response to a request.  And then `$XResponse` for the return value of a function call.

packages/devtools-client-adapters/src/types.js
```diff
@@ -0,0 +1,162 @@
...
+  condition: ?string,
+};
+
+/**
+ * Breakpoint Result is the return from an add/modify Breakpoint request
+ *
+ * @memberof types
+ * @static
+ */
+export type BreakpointResult = {
```
clarkbw: This is fairly Firefox specific, but also seems to be used in other places so I kept it here for now.  I wanted to have these types become much more of a general case, losing the term `actor` and moving toward `id` to bring Chrome and Firefox together; but that should be done when working on the Chrome side of things.

packages/devtools-client-adapters/src/types.js
```diff
@@ -0,0 +1,162 @@
...
+  sourceMapURL?: string
+};
+
+/**
+ * Frame ID
+ *
+ * @memberof types
+ * @static
+ */
+export type FrameId = string;
```
clarkbw: I pulled out a lot of these `id` types because we use them to reference things later so it felt better to have a specific identifier for each object so we could type the calls that look like `get(FrameId)`.

packages/devtools-client-adapters/src/types.js
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * Why is the Debugger Paused?
+ * This is the generic state handling the reason the debugger is paused.
+ * Reasons are usually related to "breakpoint" or "debuggerStatement"
+ * and should eventually be specified here as an enum.  For now we will
+ * just offer it as a string.
+ * @memberof types
+ * @static
+ */
+export type WhyPaused = {
```
clarkbw: These kinds of types will help us get to a more generic version of a feature like this.

jasonLaster:
@clarkbw just approved the PR with some small tweaks of my own. Feel free to change any of my tweaks or merge whenever you want.

I'll get circle running tomorrow so that we don't regress on flow, or the other linters.

PR #50, opened 2017-01-02
Adds firefox and chrome drivers

jasonLaster:
This adds the firefox and chrome drivers

goals:
* new contributors can run `yarn run firefox`
* launchpad will be able to have a button `launch firefox`


packages/devtools-launchpad/bin/chrome-driver.js
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env node
+
+const spawn = require('child_process').spawn;
+
+const chromeBinary = "/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome"
```
wldcordeiro: This seems to be a OS X specific solution. Is there a way this can be done in `local.config` or overridden there?

jasonLaster: nope - we should fix this separately. In fact, if you could fix it as I think you have linux that would be awesome

PR #54, opened 2017-01-04
close connection with beforeunload handler

clarkbw:
This should fix https://github.com/devtools-html/debugger.html/issues/502


packages/devtools-client-adapters/src/chrome.js
```diff
@@ -78,6 +78,10 @@ function connectNode(tab) {
   return CDP({ tab: tab.webSocketDebuggerUrl })
     .then(conn => {
       connection = conn;
+      window.addEventListener("beforeunload", () => {
+        connection.onclose = function disable() {};
```
clarkbw: In researching this I found that others hadn't disabled the `onclose` handler and then other function would try to run, but in this case we are out of time and the window is closing.

PR #63, opened 2017-01-05
  [DarkMode] fix text selection #1622 from debugger.html

andreicristianpetcu:
Fixing this issue https://github.com/devtools-html/debugger.html/issues/1622

before focused
before_focused

before unfocused
before_unfocused

after focused or unfocused
after_focused_or_unfocused

I think both focused or unfocused should be the same


andreicristianpetcu:
I shamelessly stole the current color from the current debugger :D

Issue #64, opened 2017-01-05
LaunchPad should sync tabs


jasonLaster:
It would be nice for the launch pad to check every second or so for new tabs on the open browsers. 

Here's a [patch](https://gist.github.com/854a5f3a2f0db707522c830ea906ab27) to help get started

```bash
curl https://gist.githubusercontent.com/jasonLaster/854a5f3a2f0db707522c830ea906ab27/raw/3cbeec0d10a15147ca3562479e2f21ea5f653b5b/foo.diff | git apply
```

bomsy:
@wldcordeiro Are you still looking into this?

bomsy:
This should be fixed!

PR #66, opened 2017-01-05
Sync client tabs on an interval

wldcordeiro:
This closes issue #64 making the tabs sync themselves on a 1 second interval. Most credit to @jasonLaster who live coded it during our community call today 😄

### DIFF
--- a/packages/devtools-launchpad/src/index.js
+++ b/packages/devtools-launchpad/src/index.js
@@ -104,7 +104,27 @@ function getTargetFromQuery() {
   return null;
 }
 
-async function bootstrap(React, ReactDOM, App, appActions, appStore) {
+function getTabs(actions) {
+  chrome.connectClient().then(tabs => {
+    actions.newTabs(tabs);
+  }).catch(e => {
+    console.log("Connect to chrome:");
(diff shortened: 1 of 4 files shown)


packages/devtools-launchpad/src/index.js
```diff
@@ -93,19 +93,7 @@ function getTargetFromQuery() {
   return null;
 }
 
-function bootstrap(React, ReactDOM, App, appActions, appStore) {
-  const connTarget = getTargetFromQuery();
-  if (connTarget) {
```
jasonLaster: we should update this function so that it is *just* getting tabs. 

currently we're also bootstrapping here, which is silly. We only want to establish the connection once

wldcordeiro: Yeah good idea. Just did a first pass based on the patch but it'd make more sense to have it do just one thing.

wldcordeiro: Actually I just double checked and we do just get the tabs and then in bootstrap we call `getTabs` https://github.com/devtools-html/devtools-core/pull/66/files#diff-4070cf78e7d2f907263e33fe7190d02dR127

PR #67, opened 2017-01-06
Fix tab filter background and text color

wldcordeiro:
The tab filter background wasn't set so it remained white. It now respects the theme and the text color was set to match the colors used in the EditorSearch in the debugger.

Screenshot
screenshot_20170105_182033

### DIFF
--- a/packages/devtools-launchpad/src/components/LandingPage.css
+++ b/packages/devtools-launchpad/src/components/LandingPage.css
@@ -84,12 +84,17 @@
 
 .landing-page .panel header input {
   flex: 1;
-  color: var(--theme-body-color);
+  background-color: var(--theme-tab-toolbar-background);
+  color: var(--theme-comment);
   font-size: var(--ui-element-font-size);
   border: 1px solid var(--theme-splitter-color);
   padding: calc(var(--base-spacing) / 2);
   margin: 0 var(--base-spacing);
   transition: var(--base-transition);
+}
+
+.landing-page .panel header input::placeholder {
+  color: var(--theme-body-color-inactive);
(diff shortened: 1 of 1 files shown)


packages/devtools-launchpad/src/components/LandingPage.css
```diff
@@ -84,14 +84,19 @@
 
 .landing-page .panel header input {
   flex: 1;
-  color: var(--theme-body-color);
+  background-color: var(--theme-body-background);
+  color: var(--theme-comment);
```
nchevobbe: I'm afraid this color looks too similar to the one in the placeholder 

capture du 2017-01-06 08-15-31

Can we try something more contrasting ?

wldcordeiro: I agree it's a bit lacking in contrast. I chose to use the same style as the search inputs in the debugger. Maybe worth changing here and in the debugger.


https://github.com/devtools-html/debugger.html/blob/master/src/components/Autocomplete.css#L55
https://github.com/devtools-html/debugger.html/blob/master/src/components/EditorSearchBar.css#L23

jasonLaster:
haha - definitely lacking. And yes, we could use @nchevobbe's design eye on the debugger too!

How about we go with the the sidebar background?

nchevobbe:
@wldcordeiro would you have some time to finish this ? If you don't, I can do it :)

wldcordeiro:
Whoops, had forgotten about this PR will update it shortly.

jasonLaster:
👏  Thanks!

PR #74, opened 2017-01-09
Cleanup build options

jasonLaster:
* switch to using an implicit bundleName over just using "bundle" or passing in an explicit one
* drop library name 
* cleanup some code

### DIFF
--- a/packages/devtools-launchpad/bin/development-server.js
+++ b/packages/devtools-launchpad/bin/development-server.js
@@ -35,18 +35,28 @@ function httpOrHttpsGet(url, onResponse) {
   });
 }
 
+function getFavicon() {
+  let favicon = getValue("favicon");
+
+  if (!favicon) {
+    return "launchpad-favicon.png";
+  }
+
+  return path.basename(favicon);
+}
+
 function serveRoot(req, res) {
   const tplPath = path.join(__dirname, "../index.html");
   const tplFile = fs.readFileSync(tplPath, "utf8");
+  const bundleName = getValue("title") ?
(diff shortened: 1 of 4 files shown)


packages/devtools-launchpad/bin/development-server.js
```diff
@@ -35,18 +35,26 @@ function httpOrHttpsGet(url, onResponse) {
...
+
 function serveRoot(req, res) {
   const tplPath = path.join(__dirname, "../index.html");
   const tplFile = fs.readFileSync(tplPath, "utf8");
 
-  let favicon = getValue("favicon");
   res.send(Mustache.render(tplFile, {
     isDevelopment: isDevelopment(),
     dir: getValue("dir") || "ltr",
+    bundleName: getValue("title").toLocaleLowerCase() || "bundle",
```
nchevobbe: wouldn't it break if the config does not specify a title ?

wldcordeiro: @nchevobbe The or should fall back to bundle.

nchevobbe: I was thinking `getValue("title")` is undefined if there isn't a title entry in the configuration, hence, doing `.toLocaleLowerCase()` on undefined would have throw.
I'll check and open a PR if there's such issue

jasonLaster:
sorry if i missed something @nchevobbe

nchevobbe:
> sorry if i missed something @nchevobbe

nope, sounds good, I didn't saw that you fixed it :)

Issue #69, opened 2017-01-06
Linking is not working


jasonLaster:
We've seen several issues with linking the debugger to devtools-core packages

* `yarn link client-devtools-adapters`
* `yarn link devtools-modules`
* `yarn link devtools-launchpad`

Some of the reported issues are
* sources do not show up (likely an issue with client adapters)
* prefs is not defined (likely a devtools-modules issue)

It's possible to link multiple repos i.e. run `yarn link client-devtools-adapters` and `yarn link devtools-launchpad` which might be causing issues. 

Is it possible to run `yarn unlink client-devtools-adapters`? If so we might want to document that.
If not we should document the good old `rm -r node_modules` approach as devs should know how to uninstall.

tromey:
I tried linking as suggested in the "Local Development" doc, and when I tried loading the debugger from the development server, I got:

```ERROR in ../devtools-core/packages/devtools-client-adapters/index.js
Module build failed: ReferenceError: Unknown plugin "transform-flow-strip-types" specified in "base" at 0, attempted to resolve relative to "/home/tromey/ff-github/devtools-core/packages/devtools-client-adapters"
    at /home/tromey/ff-github/debugger.html/node_modules/babel-core/lib/transformation/file/options/option-manager.js:176:17
    at Array.map (native)
    at Function.normalisePlugins (/home/tromey/ff-github/debugger.html/node_modules/babel-core/lib/transformation/file/options/option-manager.js:154:20)
    at OptionManager.mergeOptions (/home/tromey/ff-github/debugger.html/node_modules/babel-core/lib/transformation/file/options/option-manager.js:229:36)
    at OptionManager.init (/home/tromey/ff-github/debugger.html/node_modules/babel-core/lib/transformation/file/options/option-manager.js:374:12)
    at File.initOptions (/home/tromey/ff-github/debugger.html/node_modules/babel-core/lib/transformation/file/index.js:216:65)
    at new File (/home/tromey/ff-github/debugger.html/node_modules/babel-core/lib/transformation/file/index.js:139:24)
    at Pipeline.transform (/home/tromey/ff-github/debugger.html/node_modules/babel-core/lib/transformation/pipeline.js:46:16)
    at transpile (/home/tromey/ff-github/debugger.html/node_modules/babel-loader/lib/index.js:38:20)
    at Object.module.exports (/home/tromey/ff-github/debugger.html/node_modules/babel-loader/lib/index.js:133:12)
[...]

PR #78, opened 2017-01-11
support linking for all devtools-* modules

juliandescottes:
Fix for #69 

With this simple change I am able to use linking for all devtools-core packages with the debugger.
In the long run,  webpack2 should also improve the situation here.

Please note that when using linking, the "debugger.js" bundle goes from ~700 files to ~1200 files.

cc @jasonLaster


jasonLaster:
very cool

PR #90, opened 2017-01-16
Fix devtools-core for reps (linking & ff client)

juliandescottes:
This should fix the issues mentioned in https://github.com/devtools-html/devtools-reps/issues/43

The webpack config now explicitly excludes all devtools-core module (except for launchpad & client-adapters, which were "unexcluded" afterwards).

Also fixed a small bug with the ff client adapters which makes the actions argument of the launchpad bootstrap method mandatory, while reps is not providing any action.

### DIFF
--- a/packages/devtools-launchpad/webpack.config.devtools.js
+++ b/packages/devtools-launchpad/webpack.config.devtools.js
@@ -42,11 +42,15 @@ module.exports = (webpackConfig, envConfig) => {
       return;
     } else if (nativeMapping[mod]) {
       const mapping = nativeMapping[mod];
-      if (Array.isArray(mapping)) {
-        callback(null, `var devtoolsRequire("${mapping[0]}")["${mapping[1]}"]`);
-      }
-      else {
-        callback(null, `var devtoolsRequire("${mapping}")`);
+
+      if (webpackConfig.externalsRequire) {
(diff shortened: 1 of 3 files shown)


juliandescottes:
I tested the following use cases: 
- [x] reps using links to devtools-* packages
- [x] debugger.html using links to devtools-* packages

Both work fine with this PR.

juliandescottes:
cc @jasonLaster this should now be ready to review.

Last commit fixes #91, but requires to update debugger.html when we will bump devtools-launchpad there.

packages/devtools-launchpad/webpack.config.devtools.js
```diff
@@ -42,11 +42,15 @@ module.exports = (webpackConfig, envConfig) => {
...
-        callback(null, `var devtoolsRequire("${mapping[0]}")["${mapping[1]}"]`);
-      }
-      else {
-        callback(null, `var devtoolsRequire("${mapping}")`);
+
+      if (webpackConfig.externalsRequire) {
+        // If the tool defines "externalsRequire" in the webpack config, wrap
+        // all require to external dependencies with a call to the tool's
+        // externalRequire method.
+        let reqMethod = webpackConfig.externalsRequire;
```
jasonLaster: I think we still need to handle the mapping array case

Issue #85, opened 2017-01-13
Splitter not resizable when Debugger is docked to side of browser window


jasonLaster:
Splitter not resizable when Debugger is docked to side of browser window

BUGS
* https://github.com/devtools-html/debugger.html/issues/1378
* https://bugzilla.mozilla.org/show_bug.cgi?id=1322062

NOTES
* This is caused by some simplifications in how we calculate screen width and mouse position in order to work with chromium.
* This is complicated in part because we want to support RTL out of the box. I wonder if we can handle this concern in JS and check the root `dir` attribute and do the switch in JS.

PR #94, opened 2017-01-17
fix wrong calculation when panels are open

amitzur:
In split box, we calculate screenX and screenY with a consideration of the window's border. This is done by comparing window.outerWidth and window.innerWidth (or height).
When there are panels (e.g. devtools :) or bookmarks) the difference between outer and inner size is very large. So the split box breaks.

Neglecting the window's border is a small price to pay since it's around 5px (in Windows) compared to not being about to dock right (or dock bottom with a vertical flow).
The ideal solution would be to not use screenX and screenY, but I realize the tradeoff with not having to calculate relative coordinates of iframes.

### DIFF
--- a/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
+++ b/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
@@ -101,13 +101,13 @@ const SplitBox = React.createClass({
   },
 
   screenX() {
-    const borderWidth = (window.outerWidth - window.innerWidth) / 2;
-    return window.screenX + borderWidth;
+    // NOTE: in practice the window might have a border which calls for comparing window.outerWidth and window.innerWidth
+    return window.screenX;
   },
 
   screenY() {
-    const borderHeignt = (window.outerHeight - window.innerHeight);
-    return window.screenY + borderHeignt;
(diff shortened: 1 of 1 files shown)


amitzur:
related to #85 and to https://github.com/devtools-html/debugger.html/issues/1378

amitzur:
I just want to add that I brought up the iframes consideration since that was the rationale depicted in Draggable.js for using screen coords over page coords.

https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-modules/client/shared/components/splitter/Draggable.js#L30

jasonLaster:
Thanks. Yeah, honestly I think we'll have to revisit this in the future.

The version of splitter and draggable in m-c have diverged and we should
try to fix that and add good coverage for all of these cases.

juliandescottes:
> I just want to add that I brought up the iframes consideration since that was the rationale depicted in Draggable.js for using screen coords over page coords.

Heads up: we now use clientX/clientY in draggable.js in devtools. Computation was off when zoomed in with screenX / screenY

See http://searchfox.org/mozilla-central/source/devtools/client/shared/components/splitter/draggable.js#34

PR #95, opened 2017-01-18
Move lerna to dependency instead of dev dependency

gregtatum:
I just tried to add `devtools-core` to perf.html via `yarn add devtools-core -D`, but it failed with:

```
error /Users/gregorytatum/dev/cleopatra/node_modules/devtools-core: Command failed.
Exit code: 127
Command: sh
Arguments: -c lerna bootstrap
Directory: /Users/gregorytatum/dev/cleopatra/node_modules/devtools-core
Output:
sh: lerna: command not found
```

I globally installed lerna, re-ran the command, and it worked. I believe this is because the dependency was a dev dependency and not included with `devtools-core` when I tried to install.


jasonLaster:
ahh, @gregtatum try following the https://github.com/devtools-html/hello-world

PR #103, opened 2017-01-20
add unmountRoot in devtools-launchpad

juliandescottes:
Related to the memory leak we found yesterday, I think it makes sense to leave the responsibility to launchpad to unmount the proper node, since launchpad is the one that decides to create a child in #mount.


jasonLaster:
tested it locally!

PR #112, opened 2017-01-20
Use webpack Node api over creating system process

Garbee:
This allows the script to be ran in Windows where it currently is failing for whatever reason.

**Warning** I've only done the most superficial test of using yarn to link to reps and running reps install to see if I got the site to run after doing `yarn start`. Beyond that, I really have no idea what to look for to verify this is working as it was before.

### DIFF
--- a/packages/devtools-launchpad/src/tools/mc/make-bundle.js
+++ b/packages/devtools-launchpad/src/tools/mc/make-bundle.js
@@ -1,27 +1,30 @@
 const path = require("path");
-const ps = require("child_process");
-const processHandler = require("../utils/process-handler");
+const webpack = require("webpack");
+const process = require("process");
 
 function makeBundle({ outputPath, projectPath, watch = false }) {
-  const webpackConfigPath = path.resolve(projectPath, "webpack.config.js");
-
-  let args = [`--config=${webpackConfigPath}`];
-  if (watch) {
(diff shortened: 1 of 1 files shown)


janodvarko:
I am able to successfully run `yarn install` for devtools-core but, not for `devtools-reps`. Is this pull request supposed to fix also `devtools-reps` module?

Honza

jasonLaster:
@janodvarko you can comment out the prepublish line in the reps package.Json file and it should  work.

Garbee:
I'll be looking at this again in the morning. Just got home and am extremely tired.

Garbee:
@janodvarko Give `yarn upgrade` a shot. I think we just need to update the yarn lock as it works after that here in Windows 7. (PRing the lock update now.)

Garbee:
And ignore that. `upgrade` works since it isn't triggering the webpack build. Investigating more...

Garbee:
So, the problem is the last version bump was `0.0.29` which was just before this PR was merged. So even a fully updated `launchpad` version doesn't have the fix. Once we get a `0.0.30` tagged and we upgrade the dep in reps it will work just fine.

Thanks for catching that @janodvarko!

janodvarko:
Works for me now, thanks!

Honza

Issue #70, opened 2017-01-06
Add a browser launch button


jasonLaster:
The launchpad establishes a connection with firefox and chrome. It also has browser drivers for launching them with a debugger port.

It would be pretty cool if the launchpad provided a button for launching chrome or firefox if one of the browsers wasn't already running. 

Here's a patch:

```patch

diff --git a/packages/devtools-launchpad/bin/development-server.js b/packages/devtools-launchpad/bin/development-server.js
index 0825b80..2f8153c 100755
--- a/packages/devtools-launchpad/bin/development-server.js
+++ b/packages/devtools-launchpad/bin/development-server.js
@@ -64,6 +64,19 @@ function handleNetworkRequest(req, res) {
   }
 }

+function handleLaunch(req, res) {
+  const browser = req.query.browser;
+
+  const location = "https://devtools-html.github.io/debugger-examples/"
+  if (browser == "firefox") {
+    return ps.spawn(`./firefox-driver --start --location ${location}`)
+  }
+
+  if (browser == "chrome") {
+    return ps.spawn(`./chrome-driver`)
+  }
+}
+
 function onRequest(err, result) {
   const serverPort = getValue("development.serverPort");

@@ -98,6 +111,9 @@ function startDevServer(devConfig, webpackConfig) {
     app.get("/", serveRoot);
   }
   app.get("/get", handleNetworkRequest);
+
+  app.post("/launch", handleLaunch)
+
   const serverPort = getValue("development.serverPort");
   app.listen(serverPort, "0.0.0.0", onRequest);

diff --git a/packages/devtools-launchpad/src/components/LandingPage.js b/packages/devtools-launchpad/src/components/LandingPage.js
index af85435..e2b574c 100644
--- a/packages/devtools-launchpad/src/components/LandingPage.js
+++ b/packages/devtools-launchpad/src/components/LandingPage.js
@@ -23,6 +23,13 @@ function firstTimeMessage(title, urlPart) {
   );
 }

+function launchBrowser(browser) {
+  fetch("/launch", { // this is p
+    params: { browser },
+    method: "post"
+  });
+}
+
 function getTabURL(tab, paramName) {
   const tabID = tab.get("id");
   return `/?${paramName}=${tabID}`;
@@ -68,7 +75,15 @@ const LandingPage = React.createClass({

   renderTabs(tabs, paramName) {
     if (!tabs || tabs.count() == 0) {
-      return dom.div({}, "");
+      // show the button in here
[...]

Issue #116, opened 2017-01-23
Yarn run firefox fails when a tool is symlinked to the launchpad


jasonLaster:
This happens because `start-firefox` points to the launchpad script, but `geckodriver` is installed locally. 

```
../../../devtools-core/packages/devtools-launchpad/bin/firefox-driver.js
```

```
➜  debugger.html yarn run firefox
yarn run v0.18.1
$ ./node_modules/.bin/start-firefox --start --location https://devtools-html.github.io/debugger-examples/
/Users/jlaster/src/mozilla/devtools-core/packages/devtools-launchpad/node_modules/selenium-webdriver/firefox/index.js:298
    throw Error(
    ^

Error: The geckodriver executable could not be found on the current PATH. Please download the latest version from https://github.com/mozilla/geckodriver/releases/WebDriver and ensure it can be found on your PATH.
    at findGeckoDriver (/Users/jlaster/src/mozilla/devtools-core/packages/devtools-launchpad/node_modules/selenium-webdriver/firefox/index.js:298:11)
    at ServiceBuilder (/Users/jlaster/src/mozilla/devtools-core/packages/devtools-launchpad/node_modules/selenium-webdriver/firefox/index.js:410:22)
    at createGeckoDriver (/Users/jlaster/src/mozilla/devtools-core/packages/devtools-launchpad/node_modules/selenium-webdriver/firefox/index.js:515:19)
    at Function.createSession (/Users/jlaster/src/mozilla/devtools-core/packages/devtools-launchpad/node_modules/selenium-webdriver/firefox/index.js:631:14)
    at createDriver (/Users/jlaster/src/mozilla/devtools-core/packages/devtools-launchpad/node_modules/selenium-webdriver/index.js:167:33)
    at Builder.build (/Users/jlaster/src/mozilla/devtools-core/packages/devtools-launchpad/node_modules/selenium-webdriver/index.js:642:16)
    at start (/Users/jlaster/src/mozilla/devtools-core/packages/devtools-launchpad/bin/firefox-driver.js:70:6)
    at Object.<anonymous> (/Users/jlaster/src/mozilla/devtools-core/packages/devtools-launchpad/bin/firefox-driver.js:76:18)
    at Module._compile (module.js:573:32)
    at Object.Module._extensions..js (module.js:582:10)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this
```

bomsy:
i noticed this late yesterday!

PR #117, opened 2017-01-23
Browser launch button

bomsy:
Associate Issue: #70 
### Summary of Changes

* Added launch buttons for firefox and chrome
*

### Test Plan

Example test plan:

- [x] Start the debugger
- [x] Click the `Launch Chrome` button
- [x] It should start the chrome browser
- [x] Do the same to test firefox

browser_launch_button

### DIFF
--- a/packages/devtools-launchpad/bin/development-server.js
+++ b/packages/devtools-launchpad/bin/development-server.js
@@ -19,6 +19,7 @@ const checkNode = require("check-node-version");
 const getValue = require("devtools-config").getValue;
 const setConfig = require("devtools-config").setConfig;
 const isDevelopment = require("devtools-config").isDevelopment;
+const firefoxDriver = require("./firefox-driver");
 
 function httpOrHttpsGet(url, onResponse) {
   let protocol = url.startsWith("https:") ? https : http;
@@ -91,11 +92,11 @@ function handleLaunchRequest(req, res) {
 
   process.env.PATH += `:${__dirname}`;
   if (browser == "Firefox") {
-    ps.spawn("firefox-driver.js", ["--start", "--location", `${location}`])
+    firefoxDriver.start();
   }
 
   if (browser == "Chrome") {
-    ps.spawn("chrome-driver.js", ["--location", `${location}`]);
+    ps.spawn("chrome-driver.js", ["--location", location]);
   }
 
 }
--- a/packages/devtools-launchpad/bin/development-server.js
+++ b/packages/devtools-launchpad/bin/development-server.js
@@ -93,17 +93,13 @@ function handleLaunchRequest(req, res) {
 
   process.env.PATH += ':' + __dirname;
   if (browser == "Firefox") {
-    console.log(browser)
-    var b = ps.spawn("firefox-driver.js", ["--start", "--location", `${location}`]);
+    ps.spawn("firefox-driver.js", ["--start", "--location", `${location}`])
   }
 
   if (browser == "Chrome") {
     ps.spawn("chrome-driver.js", ["--location", `${location}`]);
   }
 
-  b.stdout.on("data", data =>  {
-    console.log(data);
-  });
 }
 
 function onRequest(err, result) {
--- a/packages/devtools-launchpad/bin/development-server.js
+++ b/packages/devtools-launchpad/bin/development-server.js
@@ -20,8 +20,6 @@ const getValue = require("devtools-config").getValue;
 const setConfig = require("devtools-config").setConfig;
 const isDevelopment = require("devtools-config").isDevelopment;
 
-const _env = process.env;
-
 function httpOrHttpsGet(url, onResponse) {
   let protocol = url.startsWith("https:") ? https : http;
 
@@ -91,7 +89,7 @@ function handleLaunchRequest(req, res) {
   const browser = req.body.browser;
   const location = "https://devtools-html.github.io/debugger-examples/";
 
-  process.env.PATH += ':' + __dirname;
+  process.env.PATH += `:${__dirname}`;
   if (browser == "Firefox") {
     ps.spawn("firefox-driver.js", ["--start", "--location", `${location}`])
   }
--- a/packages/devtools-launchpad/bin/development-server.js
+++ b/packages/devtools-launchpad/bin/development-server.js
@@ -10,6 +10,7 @@ const ps = require("child_process");
 const Mustache = require("mustache");
 const webpack = require("webpack");
 const express = require("express");
+const bodyParser = require("body-parser");
 const webpackDevMiddleware = require("webpack-dev-middleware");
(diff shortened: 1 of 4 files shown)


bomsy:
The `Launch Firefox` doesn't seem to work, i think it's related to #116

nchevobbe:
This looks great :)
I think it would make more sense if the buttons where aside (or in) the tabs in the left sidebar but we can tackle this in an other issue

packages/devtools-launchpad/src/components/LandingPage.css
```diff
@@ -93,6 +93,10 @@
   transition: var(--base-transition);
 }
 
+.landing-page .panel header input[type=button] {
+  flex: 0.25;
```
nchevobbe: I'm not sure about this, maybe we could only have some padding on the button and let the search input take all the remaining space ?
What do you think about it ?

packages/devtools-launchpad/src/components/LandingPage.js
```diff
@@ -137,8 +137,20 @@ const LandingPage = React.createClass({
       filterString = ""
     } = this.props;
 
+    let { selectedPane } = this.state;
+
     const targets = getTabsByClientType(tabs, clientType);
 
+    function launchBrowser(browser) {
+      fetch("/launch", {
+        body: JSON.stringify({ browser: browser }),
```
nchevobbe: we can use ES6 shorthand property here  `JSON.stringify({ browser })`

packages/devtools-launchpad/src/components/LandingPage.js
```diff
@@ -137,8 +137,20 @@ const LandingPage = React.createClass({
       filterString = ""
     } = this.props;
 
+    let { selectedPane } = this.state;
+
     const targets = getTabsByClientType(tabs, clientType);
 
+    function launchBrowser(browser) {
+      fetch("/launch", {
```
nchevobbe: could this possibly fail ( e.g. for example if the node process was killed) ? I f so, we should handle the case where it fails (show a message, an alert, or something better :) )

packages/devtools-launchpad/bin/development-server.js
```diff
@@ -83,6 +85,21 @@ function handleNetworkRequest(req, res) {
   }
 }
 
+function handleLaunchRequest(req, res) {
+  const browser = req.body.browser;
+  const location = "https://devtools-html.github.io/debugger-examples/";
+
+  process.env.PATH += `:${__dirname}`;
+  if (browser == "Firefox") {
+    ps.spawn("firefox-driver.js", ["--start", "--location", `${location}`])
```
nchevobbe: `'${location}'` can be simplified in `location` :)

packages/devtools-launchpad/bin/development-server.js
```diff
@@ -83,6 +85,21 @@ function handleNetworkRequest(req, res) {
...
+  }
+
+  if (browser == "Chrome") {
+    ps.spawn("chrome-driver.js", ["--location", `${location}`]);
```
nchevobbe: same here

packages/devtools-launchpad/bin/development-server.js
bomsy: haha 👍

packages/devtools-launchpad/bin/development-server.js
bomsy: 👍

packages/devtools-launchpad/src/components/LandingPage.css
bomsy: i'm fine with that. i'm not particular on styling, lets go with what looks best!

bomsy: hmm.... on second thought. we would need to have some sort of fixed size on the input ... since `flex: 1` is applied to all inputs.

packages/devtools-launchpad/bin/development-server.js
jasonLaster: I think you could probably call it directly :)

it exports a `start` function 
https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/bin/firefox-driver.js#L105

packages/devtools-launchpad/src/components/LandingPage.js
```diff
@@ -154,8 +166,14 @@ const LandingPage = React.createClass({
               this.onTabClick(targets.first(), paramName);
             }
           }
-        })
-      ),
+        }),
+        dom.input({
+          type: "button",
+          value: `Launch ${selectedPane}`,
+          onClick: e => {
```
jasonLaster: `() => launchBrowser(selectedPane)`

packages/devtools-launchpad/bin/development-server.js
bomsy: ahh ... cool!

packages/devtools-launchpad/src/components/LandingPage.js
bomsy: maybe we can show alert, for now and change it later.

packages/devtools-launchpad/src/components/LandingPage.css
nchevobbe: We shouldn't. Flexbox allocate the space left after the elements were assigned a size. So if an element has no size, it will keep its original one, and elements with flex will grow to fill up the remaining space. 
In this case, we can either have a fixed width (which I don't love because the text might change, user might zoom in page, ...), or either having no width nor flex, only a simple padding so it doesn't look too small.

bomsy: I totally agree with you. was just noting that removing the flex and just adding padding does not just work, because the button inherits the flex size from the inputs styling.

Might have to set the `flex: none` to override.

nchevobbe: urgh, sorry about that
Only the search input should have the `flex: 1` property, so we may target it specifically with `input[type="search"]`

packages/devtools-launchpad/src/components/LandingPage.css
```diff
@@ -93,6 +93,10 @@
   transition: var(--base-transition);
 }
 
+.landing-page .panel header input[type=button] {
+  flex: none;
+}
```
bomsy: @nchevobbe updated! The padding is inherited and looks ok, just had to override the inherited flex size.

nchevobbe: Sorry to bother you with this, but instead of this change, could we instead rewrite the rule in l.85 from `.landing-page .panel header input` to `.landing-page .panel header input[type="search"]` ?
I prefer having a more specific rule for the search input and not having one for the button you added.

bomsy: np! kk ... will do.

PR #119, opened 2017-01-23
fix process race condition

jasonLaster:
There's a strange race condition with `process.env.TARGET`, which causes the wrong config to be used.

CC @juliandescottes


juliandescottes:
Thanks for the fix!

jasonLaster:
@juliandescottes no problem. I recently merged the firefox.mcPath fix so we should be good

jasonLaster:
ironically, this doesn't require a version bump because the debugger will work with the raciness

PR #134, opened 2017-01-28
add common IDEs to gitignore

jbhoosreddy:
if it's alright, I'd like to add common IDEs to gitignore as we did in debugger.html repo.


jasonLaster:
ofcourse! Good idea

PR #136, opened 2017-01-28
[Settings] add features

jbhoosreddy:
Associated Issue: #133 

devtools-launchpad: Adds features in Settings Page.
devtools-config: Modified logic for feature checking

PR #138, opened 2017-01-29
version bump for schema change

jbhoosreddy:
Associated PR: #136 

Need to bump version to enable reading on new schema.


jasonLaster:
Hmm, @jbhoosreddy could you also bump `client-adapters` and `launchpad` with launchpad you'll need to bump the dependencies too.

I just wrote some docs for versioning to help out! https://github.com/devtools-html/devtools-core/blob/ce6b737c14a147eddf17f3042ac185cead2e1045/docs/versioning.md

Thanks for helping here.

PR #139, opened 2017-01-29
Docs

jasonLaster:
* Organize the docs a bit
* add some content on versioning

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -12,15 +12,15 @@ There are a couple strategies for this: link, local install, and package install
 #### Linking
 
 Linking is the most common approach.
-Here, yarn creates a symlink from the debugger's `node_modules/devtools-client-adapters` directory to the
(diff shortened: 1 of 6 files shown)


docs/local-development.md
```diff
@@ -0,0 +1,53 @@
...
+It's very common to work on a package and test it on a tool like the Debugger.
+There are a couple strategies for this: link, local install, and package install.
+
+#### Linking
+
+Linking is the most common approach.
+Here, yarn creates a symlink from the debugger's `node_modules/devtools-client-adapters` directory to the
+devtools-core `devtools-client-adapters` package directory.
+
+* Once the debugger is linked then any change in the code will be reflected in the debugger.
```
nchevobbe: not only the debugger, but I don't know how we can say that better

jasonLaster: true.

Issue #148, opened 2017-01-31
Auto-restart development server on config setting changes


bomsy:
Building of work on #130 
We would like auto-restart the server so that changes to settings get picked up.

````javascript
// inside handleSetConfig
app.close('restart');

// outside
app.listen(serverPort, "0.0.0.0", onRequest);
app.on('close', data=> data == "restart" ? app.listen(serverPort, "0.0.0.0", onRequest) : null;
````

we can have `startServer` and `restartServer` functions

Issue #142, opened 2017-01-30
[Landing Page] Pull out Settings Pane into separate component


jbhoosreddy:
Tracking Issue: #129

Part of Landing Page Component refactor.

jasonLaster:
We should use the setValue action and move the API call there (similar to debugger w/ async await actions)

Issue #129, opened 2017-01-26
Refactor LandingPage


jasonLaster:
We should have many smaller components here:

* Tabs
* Settings Pane
* the local config should be put in the redux store

PR #152, opened 2017-02-01
Launchpad Refactor (#129)

jbhoosreddy:
Associated Issue: #129

This work contains some code refactor for the landing page component.

- Create Components `Sidebar`, `Tabs`
- Add actions and reducers for `config`

The more challenging part of figuring out redux-heavy logic was figured out during the community call with @jasonLaster @bomsy @Garbee and @wldcordeiro.

I suspect the config/setValue action/reducer might not work. But I would rather punt it for now. maybe @bomsy can play with it within his PR for updating the config/features programmatically #148.

This also leaves final bits of Launch Pad refactor like Panel and SettingsPane Component tracked in #142.


packages/devtools-launchpad/src/index.js
```diff
@@ -45,7 +45,13 @@ function initApp() {
 
   const store = createStore(combineReducers(reducers));
   const actions = bindActionCreators(require("./actions"), store.dispatch);
-
+  await fetch("/getconfig", {
```
jasonLaster: hmm, lets flatten this out with `await`

PR #155, opened 2017-02-01
set the javascript.options.asyncstack preference to true

clarkbw:


clarkbw:
Until https://github.com/devtools-html/debugger.html/issues/1972 is fixed this change won't do anything for the debugger but does make async stacks available to the console.

PR #158, opened 2017-02-02
Settings component

bomsy:
Associate Issue: #142

The settings pane is now a component on its own.
All updates to all the existing settings (Configurations & Features) 
now use redux, and update`local.json`

@jasonLaster @jbhoosreddy


jasonLaster:
Wow! This is great work @bomsy.

PR #160, opened 2017-02-02
Fix storage bug

jasonLaster:
Looks like we're not stripping the prefixes


clarkbw:
Failed on a lerna issue...

PR #171, opened 2017-02-07
Fix to enable tracking of navigation events in chrome

bomsy:
Associate issue #168

#### Testing Steps
- Select chrome as the debuggee
- Set breakpoints in the `willNavigate()` and `navigated()` actions
- Go back and forward using the using the back and forward buttons in chrome
- Watch for the breakpoints hit

willnavigate


jasonLaster:
Looks great!

PR #172, opened 2017-02-08
Drop SingleModulePlugin

jasonLaster:
This removes the need for SMP, which was keeping us from using `babel-traverse`.

SMP was needed before because it helps deal with duplicate versions of `devtools-config` or `react`. If we take on the responsibility of making sure that we don't do this, we are okay.

I tested this version in the debugger and hot reloading worked.

### DIFF
--- a/packages/devtools-launchpad/webpack.config.devtools.js
+++ b/packages/devtools-launchpad/webpack.config.devtools.js
@@ -4,20 +4,12 @@ const webpack = require("webpack");
 
 const { DefinePlugin } = webpack;
 
-const ignoreRegexes = [
-  /(chrome-remote-debugging-protocol)/
-];
-
 const nativeMapping = {
   "../utils/source-editor": "devtools/client/sourceeditor/editor",
   "./test-flag": "devtools/shared/flags",
   "./client/shared/shim/Services": "Services",
-
-  // React can be required a few different ways, make sure they are
-  // all mapped.
   "react": "devtools/client/shared/vendor/react",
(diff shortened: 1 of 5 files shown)


packages/devtools-launchpad/webpack.config.devtools.js
```diff
@@ -4,20 +4,12 @@ const webpack = require("webpack");
...
 const nativeMapping = {
   "../utils/source-editor": "devtools/client/sourceeditor/editor",
   "./test-flag": "devtools/shared/flags",
   "./client/shared/shim/Services": "Services",
-
-  // React can be required a few different ways, make sure they are
-  // all mapped.
   "react": "devtools/client/shared/vendor/react",
-  "devtools/client/shared/vendor/react": "devtools/client/shared/vendor/react",
-  "react/lib/ReactDOM": "devtools/client/shared/vendor/react-dom"
```
jasonLaster: it was important that we simplified all of the `react` paths as well.

packages/devtools-launchpad/webpack.config.devtools.js
```diff
@@ -4,20 +4,12 @@ const webpack = require("webpack");
 
 const { DefinePlugin } = webpack;
 
-const ignoreRegexes = [
-  /(chrome-remote-debugging-protocol)/
-];
```
jasonLaster: we should really switch this to `chrome-remote-interface`, but in reality i'm willing to take the bundle hit for the time being and remove it later when we improve the process for dropping unused modules.

PR #178, opened 2017-02-09
standalone run

jasonLaster:
This makes it possible to run the launchpad by itself.

Along the way, it also simplifies the config considerably

* devtools-core has the shared babel plugins and react deps
* there's a new base developmentConfig


nchevobbe:
The launchpad works well, but there's an error when selecting the Settings panel : 

screenshot at fevr 10 07-38-15

bomsy:
Fixed the settings issue.

packages/devtools-launchpad/config/development.json
```diff
@@ -0,0 +1,22 @@
...
+  "favicon": "assets/launchpad-favicon.png",
+  "theme": "light",
+  "dir": "ltr",
+  "logging": {
+    "client": false,
+    "firefoxProxy": false,
+    "actions": false
+  },
+  "development": {
+    "serverPort": 8000,
```
bomsy: Do we want to use a different port, so as not to conflict with debugger running locally?

jasonLaster: heh - that sounds hard to remember. anyways, we'd only want one app running and talking to firefox

bomsy: oh yeah ..kk

PR #181, opened 2017-02-10
Flow run

jasonLaster:
This makes flow global to devtools-core and adds it to CI

### Test plan
* you can run flow at the root
* you can run flow at the client
* it catches errors


clarkbw:
👍

Issue #182, opened 2017-02-10
lerna bootstrap error in yarn install


clarkbw:
I'm getting an error with the lerna bootstrap step.  I'm looking into if I still have some alternate version of lerna installed from previous work.

```
10:27 $ yarn install
yarn install v0.19.1
warning devtools-core@0.0.1: No license field
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
$ lerna bootstrap
Lerna v2.0.0-beta.31
Independent Versioning Mode
Bootstrapping 6 packages
Preinstalling packages
Installing external dependencies
Symlinking packages and binaries
/Users/clarkbw/src/mozilla/devtools-core/node_modules/lerna/lib/commands/BootstrapCommand.js:301
                  _this6.logger.warning(dependency + " is already installed for " + filteredPackage.name + ". " + "Replacing with symlink...");
                                ^

TypeError: _this6.logger.warning is not a function
    at /Users/clarkbw/src/mozilla/devtools-core/node_modules/lerna/lib/commands/BootstrapCommand.js:301:33
    at /Users/clarkbw/src/mozilla/devtools-core/node_modules/lerna/lib/commands/BootstrapCommand.js:325:15
    at Array.forEach (native)
    at /Users/clarkbw/src/mozilla/devtools-core/node_modules/lerna/lib/commands/BootstrapCommand.js:279:12
    at Array.forEach (native)
    at BootstrapCommand.symlinkPackages (/Users/clarkbw/src/mozilla/devtools-core/node_modules/lerna/lib/commands/BootstrapCommand.js:271:29)
    at /Users/clarkbw/src/mozilla/devtools-core/node_modules/lerna/lib/commands/BootstrapCommand.js:101:23
    at /Users/clarkbw/src/mozilla/devtools-core/node_modules/async/lib/async.js:718:13
    at Immediate.iterate (/Users/clarkbw/src/mozilla/devtools-core/node_modules/async/lib/async.js:262:13)
    at runCallback (timers.js:637:20)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

juliandescottes:
For what it's worth, I had a similar issue yesterday and fixed it by doing 
`rm -rf node_modules/ packages/*/node_modules/ ; yarn install`

clarkbw:
[...]

PR #183, opened 2017-02-10
rm -r node modules

clarkbw:
Addresses #182

Destroys the `node_modules` directories to reset in case of errors
Works on Mac and Windows (with GitHub Git Shell)


package.json
```diff
@@ -7,6 +7,7 @@
     "lint": "yarn run lint-css -s; yarn run lint-js -s",
     "lint-css": "stylelint packages/devtools-launchpad/**/*.css",
     "lint-js": "eslint packages/devtools-launchpad/src packages/devtools-client-adapters/src",
+    "reset": "rm -rf node_modules/ packages/*/node_modules/ && yarn install",
```
jasonLaster: switching to `nom` to be consistent with the debugger.

PR #184, opened 2017-02-11
Sync tabs

bomsy:
Associate Issue #64

#### Changes
- Added a getTabs function
- Call getTabs on intervals of 3s
- Added a clearTabs action

sync-tabs

### DIFF
--- a/packages/devtools-launchpad/src/index.js
+++ b/packages/devtools-launchpad/src/index.js
@@ -132,6 +132,18 @@ function getTargetFromQuery() {
   return null;
 }
 
+async function getTabs(actions) {
+  actions.clearTabs();
+  const chromeTabs = await chrome.connectClient();
+  actions.newTabs(chromeTabs);
+
+  const nodeTabs = await chrome.connectNodeClient();
+  actions.newTabs(nodeTabs);
+
+  const firefoxTabs = await firefox.connectClient();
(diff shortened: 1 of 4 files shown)


packages/devtools-launchpad/src/index.js
```diff
@@ -132,6 +132,18 @@ function getTargetFromQuery() {
   return null;
 }
 
+async function getTabs(actions) {
+  actions.clearTabs();
+  const chromeTabs = await chrome.connectClient();
+  actions.newTabs(chromeTabs);
+
+  const nodeTabs = await chrome.connectNodeClient();
+  actions.newTabs(nodeTabs);
```
jasonLaster: Hmm. This shouldn't be serial. 

Perhaps just await on Firefox

packages/devtools-launchpad/src/index.js
```diff
@@ -132,6 +132,18 @@ function getTargetFromQuery() {
   return null;
 }
 
+async function getTabs(actions) {
+  actions.clearTabs();
```
jasonLaster: Why clear if we can merge when the new tabs come in?

bomsy: we also want to removeTabs when they are closed! merge does not pick that up.

PR #193, opened 2017-02-16
Add source to type Location

arthur801031:
Associated Issue: https://github.com/devtools-html/debugger.html/pull/2034

### Summary of Changes
- Location is missing `source`. Without `source`, flow check in `Breakpoints.js` would produce errors.

### DIFF
--- a/packages/devtools-client-adapters/src/types.js
+++ b/packages/devtools-client-adapters/src/types.js
@@ -43,7 +43,8 @@ export type ActorId = string;
 export type Location = {
   sourceId: SourceId,
   line: number,
-  column?: number
+  column?: number,
+  source?: any
 };
 
 /**


packages/devtools-client-adapters/src/types.js
```diff
@@ -43,7 +43,8 @@ export type ActorId = string;
 export type Location = {
   sourceId: SourceId,
   line: number,
-  column?: number
+  column?: number,
+  source?: any
```
arthur801031: Hi @jasonLaster 

`source` is actually `Map` with `ArrayMapNode`. I'm not sure if it's possible to create this type, so I use `any`. Please let me know if we want to do this differently. Thank you! :)

screen shot 2017-02-16 at 9 31 57 am

jasonLaster:
Fun fact, we have a *slightly* different Breakpoint in the component :)

https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/Breakpoints.js#L135-L138

We should re-define the type in the component to account for this

a good way to do it is with this style : 
```js
type ChromeTab = Tab & {
  webSocketDebuggerUrl: string,
  type: string
}
```
https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-client-adapters/src/chrome.js#L10-L14

Issue #199, opened 2017-02-21
[Launchpad] fails to connect to firefox on 3rd refresh


jasonLaster:
I've noticed that the launchpad fails to connect to firefox on the 3rd refresh. I think this is due to the launchpad re-pinging firefox a lot.

bomsy:
i think this fixed!

PR #200, opened 2017-02-21
Fix some issues connecting to firefox

jasonLaster:
This fixes two issues connecting to firefox:
1.  refreshing several times https://github.com/devtools-html/devtools-core/issues/199
2. re-opening firefox

We also do some cleanup
1. add react 15.3.2 to core so that the launchpad can be launched by itself
2. add a simple connected component to the launchpad to test the connection
3. refactor the dev-server so that it is in src and can be modularized and linted. We'll likely want to move some src code for launchpad into a `client` dir

### DIFF
--- a/packages/devtools-launchpad/src/index.js
+++ b/packages/devtools-launchpad/src/index.js
@@ -132,8 +132,7 @@ function getTargetFromQuery() {
   return null;
 }
 
-async function getTabs(actions) {
-  actions.clearTabs();
+async function connectClients(actions) {
   chrome.connectClient()
     .then(actions.newTabs);
 
@@ -141,6 +140,12 @@ async function getTabs(actions) {
     .then(actions.newTabs);
 
   const firefoxTabs = await firefox.connectClient();
+  actions.newTabs(firefoxTabs);
+}
+
+async function getTabs(actions) {
+  actions.clearTabs();
+  const firefoxTabs = await firefox.getTabs();
   actions.newTabs(firefoxTabs);
 }
 
@@ -159,7 +164,7 @@ async function bootstrap(React, ReactDOM, App, appActions, appStore) {
 
   const { store, actions, LaunchpadApp } = await initApp();
   renderRoot(React, ReactDOM, LaunchpadApp, store);
-  await getTabs(actions);
+  await connectClients(actions);
   setInterval(async () => await getTabs(actions), 3000);
 }
 
--- a/packages/devtools-sham-modules/sham/chrome.js
+++ b/packages/devtools-sham-modules/sham/chrome.js
@@ -19,8 +19,7 @@ var ourServices = {
 
 module.exports = {
   Cc: name => {
-    if(typeof console !== "undefined") {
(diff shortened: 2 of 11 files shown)


packages/devtools-sham-modules/sham/chrome.js
```diff
@@ -19,25 +19,23 @@ var ourServices = {
 
 module.exports = {
   Cc: name => {
-    if(typeof console !== "undefined") {
-      console.log('Cc sham for', name);
+    if (typeof console !== "undefined") {
```
juliandescottes: why the empty if ?

packages/devtools-sham-modules/sham/chrome.js
```diff
@@ -19,25 +19,23 @@ var ourServices = {
...
     }
     return {
       getService: (name) => ourServices[name],
       createInstance: (iface) => ourServices[iface],
     };
   },
   CC: (name, iface, method) => {
-    if(typeof console !== "undefined") {
-      console.log('CC sham for', name, iface, method);
+    if (typeof console !== "undefined") {
```
juliandescottes: same comment

packages/devtools-launchpad/src/index.js
```diff
@@ -159,7 +164,7 @@ async function bootstrap(React, ReactDOM, App, appActions, appStore) {
 
   const { store, actions, LaunchpadApp } = await initApp();
   renderRoot(React, ReactDOM, LaunchpadApp, store);
-  await getTabs(actions);
+  await connectClients(actions);
```
juliandescottes: If I read this correctly, every 3 seconds we empty the tab list and fetch the firefox tabs.
Isn't it a problem when debugging Chrome/Node?

packages/devtools-launchpad/src/index.js
```diff
@@ -144,6 +143,12 @@ async function getTabs(actions) {
   actions.newTabs(firefoxTabs);
 }
 
+async function getTabs(actions) {
+  actions.clearTabs();
```
juliandescottes: This won't pickup newly started firefox instances. If you do:
- yarn start in launchpad
- go to localhost:8000
- click on "Launch Firefox"
=> you now have to reload the page to see the tab from the new firefox instance
(previously it was automatically picked up)

If the connection was not established, maybe we should still try to do connectClients()?

packages/devtools-launchpad/src/index.js
jasonLaster: hmm... i'll clean that up

packages/devtools-launchpad/src/index.js
jasonLaster: We do it there too. This is here to catch new tabs.

juliandescottes: My point is that getTabs is not trying to connect to new firefox clients. So, as I said, this forces you to refresh your browser after clicking on "Launch firefox".

jasonLaster:
Hmm. I believe it works though. I'll investigate why :)

PR #209, opened 2017-02-24
Fix wrong path to development-server.js in package.json

julienw:
This prevented npm install from working in debugger.html.


julienw:
Unfortunately I can't test it before, as npm doesn't support subdirectories on git dependencies :(

julienw:
@jasonLaster r?

PR #212, opened 2017-02-24
Fix splitter position when debugger is dock…

julienw:
…ed to the right

Closes devtools-html/debugger.html#2027

### DIFF
--- a/packages/devtools-modules/client/shared/components/splitter/Draggable.js
+++ b/packages/devtools-modules/client/shared/components/splitter/Draggable.js
@@ -27,9 +27,9 @@ const Draggable = React.createClass({
 
   onMove(ev) {
     ev.preventDefault();
-    // Use screen coordinates so, moving mouse over iframes
-    // doesn't mangle (relative) coordinates.
-    this.props.onMove(ev.screenX, ev.screenY);
+    // We pass the whole event because we don't know which properties
+    // the callee needs.
+    this.props.onMove(ev);
   },
 
   onUp(ev) {
--- a/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
+++ b/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
@@ -60,8 +60,9 @@ const SplitBox = React.createClass({
   getInitialState() {
     return {
       vert: this.props.vert,
-      width: this.props.initialWidth || this.props.initialSize,
-      height: this.props.initialHeight || this.props.initialSize
+      // We use integers for these properties
+      width: parseInt(this.props.initialWidth || this.props.initialSize),
+      height: parseInt(this.props.initialHeight || this.props.initialSize)
     };
   },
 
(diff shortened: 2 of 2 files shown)


julienw:
Not ready for merge, RTL is broken.

packages/devtools-modules/client/shared/components/splitter/Draggable.js
```diff
@@ -19,24 +19,25 @@ const Draggable = React.createClass({
 
   startDragging(ev) {
     ev.preventDefault();
-    const doc = ReactDOM.findDOMNode(this).ownerDocument;
-    doc.addEventListener("mousemove", this.onMove);
-    doc.addEventListener("mouseup", this.onUp);
+    const elt = ReactDOM.findDOMNode(this);
+    elt.setCapture(true);
```
julienw: when called on mousedown, all mouse events are then redirected to this element until mouseup (no need to use releaseCapture).

"true" is to retarget events to this element -- this overcomes the iframe issue that was mentioned in the previous code.

packages/devtools-modules/client/shared/components/splitter/Draggable.js
```diff
@@ -19,24 +19,25 @@ const Draggable = React.createClass({
...
   },
 
   onMove(ev) {
     ev.preventDefault();
-    // Use screen coordinates so, moving mouse over iframes
-    // doesn't mangle (relative) coordinates.
-    this.props.onMove(ev.screenX, ev.screenY);
+    // because we use setCapture, the event's properties are correct even with
+    // iframes.
+    this.props.onMove(ev);
```
julienw: passing the whole event as there are a lot of useful properties there. We might want to create a new object instead but I'd rather not because of GC (mousemove events happen a lot)

packages/devtools-modules/client/shared/components/splitter/SplitBox.js
```diff
@@ -60,8 +60,9 @@ const SplitBox = React.createClass({
   getInitialState() {
     return {
       vert: this.props.vert,
-      width: this.props.initialWidth || this.props.initialSize,
-      height: this.props.initialHeight || this.props.initialSize
+      // We user integers for these properties
```
julienw: typo :D

packages/devtools-modules/client/shared/components/splitter/SplitBox.js
```diff
@@ -100,54 +101,30 @@ const SplitBox = React.createClass({
...
-    const win = doc.defaultView;
 
-    let size;
-    let { endPanelControl } = this.props;
+    if (this.props.endPanelControl) {
+      // For the end panel we need to increase the width/height when the
+      // movement is towards the left/top.
+      movementX = -movementX;
+      movementY = -movementY;
+    }
```
julienw: likely need the same operation for RTL.

I think we can make RTL work just magically but maybe in a separate patch.

packages/devtools-modules/client/shared/components/splitter/SplitBox.js
```diff
@@ -100,54 +101,30 @@ const SplitBox = React.createClass({
...
-      // has impact on vertical splitter only.
-      let dir = win.getComputedStyle(doc.documentElement).direction;
-      if (dir == "rtl") {
-        endPanelControl = !endPanelControl;
-      }
-
-      let innerOffset = x - this.screenX();
-      size = endPanelControl ?
-        (node.offsetLeft + node.offsetWidth) - innerOffset :
-        innerOffset - node.offsetLeft;
```
julienw: look, no more getting on the dom !

packages/devtools-modules/client/shared/components/splitter/SplitBox.js
```diff
@@ -60,8 +60,9 @@ const SplitBox = React.createClass({
   getInitialState() {
     return {
       vert: this.props.vert,
-      width: this.props.initialWidth || this.props.initialSize,
-      height: this.props.initialHeight || this.props.initialSize
+      // We user integers for these properties
+      width: parseInt(this.props.initialWidth || this.props.initialSize),
+      height: parseInt(this.props.initialHeight || this.props.initialSize)
```
clarkbw: I feel like our eslint used to require the radix with this function...

irfanhudda: Yes. But linting is limited to `devtools-launchpad` and `devtools-client-adapters` [https://github.com/devtools-html/devtools-core/blob/master/package.json#L9](https://github.com/devtools-html/devtools-core/blob/master/package.json#L9).

julienw: Why that ? parseInt always defaults to 10 these days, in all current browsers. Forcing the radix is when you need to support very old versions, I doubt this is our case.

clarkbw: I must be stuck in the past

julienw:
I think i'll remove setCapture then, and move back to adding events on the root node.

jasonLaster:
Yeah, we're supporting chrome for two reasons:

1. many contributors (including me) are using chrome for development
2. we want to launch a node debugger in electron in Q2

julienw:
@jasonLaster I think it's now clean and ready, working in Firefox and Chrome.

PR #215, opened 2017-02-27
Fix the `lint` script to fail also when the css linter fails

julienw:


package.json
```diff
@@ -4,11 +4,11 @@
   "description": "Collection of devtools packages",
   "scripts": {
     "flow": "flow",
-    "lint": "yarn run lint-css -s; yarn run lint-js -s",
+    "lint": "npm run lint-css -s; a=$?; npm run lint-js -s; b=$?; test $a -eq 0 -a $b -eq 0",
```
julienw: We could have used a simpler `npm run lint-css && npm run lint-js` but I wanted to always run both linters even if lint-css was failing, as before. But this would also work for me, so up to you.

julienw:
We discussed about this on friday, and over the week-end I realized there is no good reason we use `yarn` here. We can reasonnably assume that `npm` is always present, and it's good enough to run the scripts.

package.json
jasonLaster: Nice. 

Why npm over yarn?

julienw: This was the initial goal of the patch: use npm instead of yarn because it's always there (at least we can reasonnably assume it's globally installed).

If we use yarn, we need to either install it as devDependency or require it installed globally. I thought that to simply run scripts, npm is good enough, so no need to add an additional dependency or force the user to install it.

It's especially an issue for the "lint" script because it's run when pushing to a branch. I don't really mind for "nom" :)

julienw:
@jasonLaster I changed the PR to include only the script fix.

Issue #220, opened 2017-03-01
CSS variable --theme-comment-alt is missing in mozilla central


juliandescottes:
The variable --theme-comment-alt is used by the debugger and can be found in variables.css in this repository, but it's not defined in mozilla-central.

I guess we should simply add it to mozilla-central and then resynchronize variables.css here.
And we should try to keep them in sync in the future, with mozilla-central being the source of truth for now.

juliandescottes:
Logged https://bugzilla.mozilla.org/show_bug.cgi?id=1343478

juliandescottes:
Also missing: 
- --theme-search-overlays-semitransparent
- --theme-toolbar-background-alt

PR #221, opened 2017-03-01
move debugger-only variables to Root.css

juliandescottes:
This is a partial fix for #220 

I added two variables from devtools-core's variables.css file to the mozilla central version (cf https://bugzilla.mozilla.org/show_bug.cgi?id=1343478).

For the two others, they feel very debugger specific and I don't think we should keep them in variables.css, so I'm moving them to Root.css for now. We probably should rename them and completely move them to the debugger.html repo.

### DIFF
--- a/packages/devtools-launchpad/src/components/Root.css
+++ b/packages/devtools-launchpad/src/components/Root.css
@@ -4,7 +4,12 @@
 
 :root.theme-light,
 :root .theme-light {
-  --theme-search-overlays-semitransparent: rgba(221, 225, 228, 0.66);
+  --codemirror-gutter-background: #f4f4f4;
+}
+
+:root.theme-dark,
+:root .theme-dark {
+  --codemirror-gutter-background: #262b37;
 }
 
 * {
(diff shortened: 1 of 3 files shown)


packages/devtools-launchpad/src/components/Root.css
```diff
@@ -1,6 +1,18 @@
...
+
+:root.theme-dark,
+:root .theme-dark {
+  --theme-search-overlays-semitransparent: rgba(42, 46, 56, 0.66);
+  --theme-codemirror-gutter-background: #262b37;
+}
+
+:root.theme-firebug,
+:root .theme-firebug {
+  --theme-codemirror-gutter-background: #ebeced;
```
jasonLaster: why not move them to the debugger project? I'd like to see them in a variables.css file that's pulled in `App.js`

juliandescottes: I agree and mentioned it in the PR. I just wanted to do it in steps:
- cleanup devtools-core variables.css (this PR) 
- resynchronize variables.css from mc
- update debugger.html with a variables.css (or other name) that contains these 2 var, but renamed without --theme (which in devtools more or less means "comes from variables.css")
- remove the variables from here

I can just fold steps 1 and 4 into one and get rid of the values immediately in devtools-core. It's not like debugger.html will be impacted right away.

juliandescottes:
Meh actually --theme-codemirror-gutter-background is used in a common.css defined here, so it should probably stay in devtools core ...

juliandescottes:
Let's close this. I think --theme-codemirror-gutter-background actually deserves to be ported back to mc. I'll file a bug there and reopen a PR later here.

PR #224, opened 2017-03-03
Revert 14px font sizes

jasonLaster:
This caused some components to have 14px font sizes.

### before
screen shot 2017-03-03 at 1 23 47 pm

### after
screen shot 2017-03-03 at 1 22 57 pm


wldcordeiro:
I'd like to explore a way to override the browser default to get us a consistent size still. I'll take a look at some solutions that don't mess up other components.

PR #228, opened 2017-03-06
[WIP]Resolve path correctly

rrandom:
Associated Issue:  devtools-html/debugger.html#614

### Summary of Changes

1. Before webpack [resolve](https://github.com/devtools-html/debugger.html/blob/master/src/test/node-unit-tests.js#L48) the `__dirname` to `${root}/assets/build`. 
2. Did not use `path.resolve` because it cant handle `\` and `/` correctly

### Test plan
1. cd to debugger.html root dir.
2. run `yarn test`.
3. all tests pass

### DIFF
--- a/packages/devtools-network-request/stubNetworkRequest.js
+++ b/packages/devtools-network-request/stubNetworkRequest.js
@@ -6,7 +6,7 @@ module.exports = function(url) {
     // example.com is used at a dummy URL that points to our local
     // `/src` folder.
     url = url.replace(/http:\/\/localhost:8000/, "");
-    const requestUrl = _path.join(__dirname, "/../../src/test/mochitest/", url);
+    const requestUrl = _path.join(__dirname, "../src/test/mochitest/", url);
     const content = fs.readFileSync(requestUrl, "utf8");
 
     resolve({ content });


rrandom:
It seems also fix this issue devtools-html/debugger.html#1925 on windows if 1925 is the same thing as 614?  I'm not sure.... Maybe we should add `try...catch` and `reject` in related promises.

jasonLaster:
hmm, @rrandom this patch fails on my mac when my working directory is `debugger.html`.

```
  1) sources should download a sourcemap and create sources:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

rrandom:
It's strange on windows `path.join` returns relative path and returns absolute path when I test on ubuntu. I get confused..

jasonLaster:
Hmm, perhaps check if it's Mac or Windows. Or config library has a Windows Mac method we use in our bin scripts

rrandom:
Finally I know why. `process.platform` is `undefined` in this file, and it is using `path.join` for `posix` not windows. and get the wrong Path. Should I decide it's `undefined` or config webpack somehow so the `process.platorm` could return win32? @jasonLaster

jasonLaster:
I'll look into it Monday. Undefined is probably sufficient.

PR #233, opened 2017-03-10
Include devtools-source-map in Babel pipeline

jryans:
`devtools-source-map` is a new package using syntax that needs Babel translation, so we need to update the exclusion filter in launchpad to allow it.

### DIFF
--- a/packages/devtools-launchpad/webpack.config.js
+++ b/packages/devtools-launchpad/webpack.config.js
@@ -25,27 +25,34 @@ module.exports = (webpackConfig, envConfig) => {
   webpackConfig.module.loaders.push({
     test: /\.js$/,
     exclude: request => {
+      // Some paths are excluded from Babel
       let excludedPaths = [
         "bower_components",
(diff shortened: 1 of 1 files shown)


packages/devtools-launchpad/webpack.config.js
```diff
@@ -25,27 +25,34 @@ module.exports = (webpackConfig, envConfig) => {
...
-      // Create a regexp matching any of the paths in excludedPaths
-      let excludedRe = new RegExp(`(${ excludedPaths.join("|") })`);
-      let excluded = request.match(excludedRe);
+      let excludedRe = new RegExp(`(${excludedPaths.join("|")})`);
+      let excluded = !!request.match(excludedRe);
+
+      // Others are explicitly included for Babel
+      let includedPaths = [
+        "devtools-client-adapters",
+        path.join("devtools-launchpad", "src"),
```
jasonLaster: nice improvement here

PR #234, opened 2017-03-10
Remove extra copy of defer function and copy flow-typed version from debugger

wldcordeiro:
The defer function existed in two places and was identical, we also had a copy of it in debugger that was identical save for the flow typing that was added.


packages/devtools-launchpad/src/utils/defer.js
```diff
@@ -1,12 +1,29 @@
...
+declare var Reject: (result: any) => void;
+
+export type Defer = {
+  resolve: Resolve,
+  reject: Reject,
+  promise: Promise<any>
+};
+
+function defer(): Defer {
+  let resolve: Resolve; // eslint-disable-line no-unused-vars
```
tromey: Is this `eslint-disable` really needed?  It seems to me that `resolve` and `reject` are actually used.

jasonLaster:
Lets just nuke defer in client-adapters. It's not used.

As a general point, I don't know how i feel about exporting modules from launchpad. Rules for me are
1. launchpad bootstrap
2. shims like context menu...

not sure about what else yet

wldcordeiro:
@jasonLaster `devtools-utils`? 😄

jasonLaster:
@wldcordeiro that sounds like devtools-modules

packages/devtools-launchpad/src/utils/defer.js
jasonLaster: hmm, yeah - that can likely be removed

PR #253, opened 2017-03-16
Add bundle machinery to devtools-source-map

jryans:
This adds the Webpack config and other ancillary scripts needed to package `devtools-source-map` bundles for landing in m-c.

Much of this was copied from `devtools-reps`.


packages/devtools-source-map/assets/README.md
```diff
@@ -0,0 +1 @@
+The assets directory is mandatory to run `../bin/publish-assets.js`.
```
juliandescottes: this file was a dirty workaround to make sure the assets directory was available. Since you have firefox/moz.build here, you can remove it.

And we should simply create the assets/ folder if it doesn't exist in the publish-assets script.

packages/devtools-source-map/bin/publish-assets.js
```diff
@@ -0,0 +1,29 @@
+const { tools: { makeBundle }} = require("devtools-launchpad/index");
```
juliandescottes: publish-assets is not needed for devtools-source-map at the moment.

packages/devtools-source-map/RELEASE.md
```diff
@@ -0,0 +1,34 @@
...
+  c. create a new bug in Developer Tools [1]
+  d. commit 'Bug XXXXXXX - devtools-source-map v0.N+1.0: update devtools-source-map bundle from GitHub;r=reviewer'
+
+3. Validate & cleanup
+  a. push to try, test locally, submit for review etc ...
+  b. while try fails or some problem is detected, go back to devtools-source-map, fix the issue,
+     create a new bundle and go back to 2.a
+  c. when everything is fine and the patch is r+, land on autoland/inbound
+  d. merge the PR on github
+  e. create a tag for v0.N+1.0 on github
```
jryans: Ah yeah, I had mentioned a similar thing in a [recent][1] RFC discussion. Something like `devtools-source-map-vX.X.X`, I suppose?

[1]: https://github.com/devtools-html/rfcs/issues/16#issuecomment-285797651

packages/devtools-launchpad/webpack.config.devtools.js
```diff
@@ -22,9 +22,7 @@ module.exports = (webpackConfig, envConfig) => {
   }
 
   webpackConfig.devtool = false;
-  webpackConfig.recordsPath = path.join(
-    rootDir, "assets/module-manifest.json"
-  );
+  webpackConfig.recordsPath = path.join(os.tmpdir(), "module-manifest.json");
```
jryans: Ah, I wasn't really sure of its purpose. 😅 The main issue I was seeing is that this config from `devtools-launchpad` was always placing the file at the root of `devtools-core` (outside any packages), which seems wrong. 

The Webpack [docs][1] led me to think it was only used for hot module reloading, which is why I went down the temporary file route.

If we do in fact want it it to be checked in, we'd want the location to be based on the path to actual package in use (debugger, source map, etc.) not the launchpad. @juliandescottes, should I pass the current package path down to this launchpad config? Or override the path from the source map package? Something else...?

[1]: https://webpack.github.io/docs/configuration.html#recordspath-recordsinputpath-recordsoutputpath

packages/devtools-source-map/package.json
```diff
@@ -8,12 +8,19 @@
...
     "devtools-client-adapters": "^0.0.21",
     "devtools-config": "^0.0.12",
     "devtools-network-request": "^0.0.14",
     "md5": "^2.2.1",
     "source-map": "^0.5.6",
     "url": "^0.11.0"
+  },
+  "devDependencies": {
+    "devtools-launchpad": "^0.0.56",
+    "webpack": "1.14.0"
```
jryans: Since this package's bundling scripts use `webpack`, it seemed best to explicitly state a dev dependency, like we do in debugger, launchpad, etc.

packages/devtools-source-map/bin/publish-assets.js
jryans: Yes, fair enough, I'll remove it.

packages/devtools-source-map/assets/README.md
jryans: Okay, I'll remove it.

packages/devtools-source-map/RELEASE.md
juliandescottes: Sounds good to me!

jasonLaster: That is true - i suppose it is important to namespace commits: `[source-map] ...`
and make sure the commit range is greppable

packages/devtools-launchpad/webpack.config.devtools.js
juliandescottes: Discussed it with @jasonLaster and we surely need to improve the webpack.devtools.config.

We should be able to pass the root dir as a parameter in the env config. We should also only use this file when performing a release (logged #254 about this). 

When I tried it, the module-manifest.json was created in the sourcemap assets folder. Maybe you had problems if you are using linking?

I would say for now simply ignore the module-manifest.json created during the release, and we'll improve this in #254 ?

packages/devtools-source-map/package.json
juliandescottes: The difference is that the debugger and launchpad all use require("webpack") in a place or another. All the webpack-related things we do in devtools-source-map should be provided by launchpad.

jryans: Ah, I had thought `copy-assets` was explicitly requiring `webpack` itself, but looks like it is indeed happening via `devtools-launchpad` instead. Ok, I'll remove it here.

packages/devtools-launchpad/webpack.config.devtools.js
jryans: Okay, I'll revert this for now and we can leave it for #254. Possibly related to linking. The side effects of linking are quite frustrating since they _are_ quite convenient, but then end up breaking various things...

packages/devtools-source-map/RELEASE.md
jryans: That implies that every commit would need to apply to only a single package, doesn't it? It seems to negate some of the benefits of working in a mono-multi repo, since there will be cases where you'd want to make changes across multiple packages at once. Babel's repo, for example, seems to allow multi-package changes in a single commit. So, we should think about what we want to allow for ourselves.

If we're trying to grep changes for release notes, maybe that is better solved with tooling that filters commits based on file paths changed?

PR #255, opened 2017-03-17
[SourceMap] Add unit tests

jasonLaster:
This adds a unit test to the source map package. 

It also moves the worker code to a `source-map` module so that it is easier to separate the worker plumbing from the implementation logic. See [parser/worker](https://github.com/devtools-html/debugger.html/blob/master/src/utils/parser/worker.js).


packages/devtools-modules/client/shared/worker.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+
+export type Message = {
```
jasonLaster: I added the worker module so that it can be shared.

packages/devtools-source-map/.babelrc
```diff
@@ -0,0 +1,6 @@
+{
+  "plugins": [
+    "transform-async-to-generator",
+    "transform-flow-strip-types"
+  ]
+}
```
jasonLaster: it's helpful to ask babel to transform modules here too

packages/devtools-source-map/src/worker.js
```diff
@@ -1,234 +1,19 @@
...
-  const map = SourceMapConsumer(generator.toJSON());
-  sourceMapRequests.set(generatedId, Promise.resolve(map));
-}
-
-const publicInterface = {
   getOriginalURLs,
   getGeneratedLocation,
   getOriginalLocation,
   getOriginalSourceText,
-  enableSourceMaps,
```
jryans: Hmm, I am a bit perplexed by `enableSourceMaps`... It doesn't look like debugger is calling it even today? So... how are source maps working...? 😕 Are you removing it because it's unused?

packages/devtools-source-map/src/tests/source-map.js
```diff
@@ -0,0 +1,33 @@
...
+const { getOriginalURLs } = require("../source-map");
+
+function getMap(_path) {
+  const mapPath = path.join(__dirname, _path);
+  return fs.readFileSync(mapPath, "utf8");
+}
+
+describe("source maps", () => {
+  test("getOriginalURLs", async () => {
+    const source = {
```
jryans: Would it be possible to check in the original sources of the bundle and a script for bundling them? There could be cases where we need to fix issues with source map _generation_, so being able to regenerate test bundles in a sane way seems useful.

packages/devtools-source-map/src/worker.js
jasonLaster: It's used in restartWorker - but it's too complicated and we should be feature flagging higher up

packages/devtools-source-map/src/tests/source-map.js
jasonLaster: Hmm, that makes sense. 

I think for this test it's fine that we have a simple map file. For other tests it'd be nice to have webpack / browserify etc

packages/devtools-modules/client/shared/worker.js
jasonLaster: Hmm - I agree, this is more than just unit tests. Is it okay to do in a separate commit?

packages/devtools-source-map/src/worker.js
jryans: Ah, thanks for the pointer. Seems fine to move up to a higher level. Shouldn't we update source-map's `index.js` to stop sending the `enableSourceMap` message then?

packages/devtools-source-map/src/worker.js
```diff
@@ -1,223 +1,17 @@
...
-
-function applySourceMap(generatedId, url, code, mappings) {
-  const generator = new SourceMapGenerator({ file: url });
-  mappings.forEach(mapping => generator.addMapping(mapping));
-  generator.setSourceContent(url, code);
-
-  const map = SourceMapConsumer(generator.toJSON());
-  sourceMapRequests.set(generatedId, Promise.resolve(map));
-}
+  getOriginalURLs,
```
jryans: Please add a comment somewhere in this file to explain that the indirection to another module is for testing the worker's logic without actually launching a worker.

jasonLaster: sure

packages/devtools-source-map/.babelrc
```diff
@@ -0,0 +1,6 @@
+{
+  "plugins": [
+    "transform-async-to-generator",
```
jasonLaster: Depends on the node version :)

Also we might still want generators for perf

jryans:
I guess a good next step is to run the tests on CI as well? (I don't think we're running `jest` there for this package yet.)

jryans:
My last two review comments appear unresolved in the latest commit, so let me know when it's ready for another look. 😁

jryans:
@jasonLaster, seems like following review comments are still left:

[stop sending `enableSourceMap` message](https://github.com/devtools-html/devtools-core/pull/255#discussion_r106714836)
[add a comment about worker indirection](https://github.com/devtools-html/devtools-core/pull/255#discussion_r106715280)

jasonLaster:
Thanks @jryans I left those changes on my local branch :/

Issue #257, opened 2017-03-18
Chrome tabs do not show up


bomsy:
Chrome tabs are not showing up.
![image](https://cloud.githubusercontent.com/assets/792924/24075907/cfa44aee-0c1c-11e7-8f94-53a19c5152d1.png)

PR #258, opened 2017-03-18
Show chrome tabs

bomsy:
Associate issue: #257

### Changes
* Switch `connectClient` and `connectNodeClient` to use async await.
* Load chrome tabs and node tabs.

### DIFF
--- a/packages/devtools-launchpad/src/index.js
+++ b/packages/devtools-launchpad/src/index.js
@@ -134,20 +134,26 @@ function getTargetFromQuery() {
 }
 
 async function connectClients(actions) {
-  chrome.connectClient()
-    .then(actions.newTabs);
-
-  chrome.connectNodeClient()
-    .then(actions.newTabs);
-
   const firefoxTabs = await firefox.connectClient();
   actions.newTabs(firefoxTabs);
+
+  const chromeTabs = await chrome.connectClient();
+  actions.newTabs(chromeTabs);
+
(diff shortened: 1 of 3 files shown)


packages/devtools-launchpad/src/index.js
```diff
@@ -134,20 +134,26 @@ function getTargetFromQuery() {
...
-    .then(actions.newTabs);
-
   const firefoxTabs = await firefox.connectClient();
   actions.newTabs(firefoxTabs);
+
+  const chromeTabs = await chrome.connectClient();
+  actions.newTabs(chromeTabs);
+
+  const nodeTabs = await chrome.connectNodeClient();
+  actions.newTabs(nodeTabs);
```
jasonLaster: these should not be await because then the function waits for all three modes to finish. We should instead use `chrome.connectNodeClient().then(actions.newTabs)` to allow the function to return w/o node and chrome necessarily connecting

jasonLaster:
Nice! I'll test it soon

jasonLaster:
added the development settings so chrome would show up.

rickychien:
Many error messages throw in console after updating devtools-core. It's very annoying but I also notice that new Firefox new tab doesn't show up in launchpad.

@bomsy Hi, do you have any idea about that? I think this PR could be related to this.

screen shot 2017-03-26 at 4 07 50 pm

PR #259, opened 2017-03-20
Tree blur

roieu:
Associated Issue: #2053 (from devtools-html/debugger.html)

### Summary of Changes

* Added blur support for Tree component

### Test Plan

Tested manually using SourcesTree component from devtools-html/debugger.html

Example test plan:

- Focus in and out SourcesTree from various use-cases

Here's the Debugger's Testing doc
https://docs.google.com/document/d/1oBMRxV8A2ag2t22YsQOxTdEv0mXKzIg0tggJjRkU1S0/edit#.
Feel free to improve it!

See correspondant PR in devtools-html/debugger.html

### DIFF
--- a/packages/devtools-sham-modules/client/shared/components/tree.js
+++ b/packages/devtools-sham-modules/client/shared/components/tree.js
@@ -49,13 +49,13 @@ const TreeNode = createFactory(createClass({
   displayName: "TreeNode",
 
   componentDidMount() {
-    if (this.props.focused) {
+    if (this.props.focused && !this.props.blurred) {
       this.refs.button.focus();
     }
   },
 
   componentDidUpdate() {
-    if (this.props.focused) {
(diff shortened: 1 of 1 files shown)


packages/devtools-sham-modules/client/shared/components/tree.js
```diff
@@ -465,7 +478,22 @@ const Tree = module.exports = createClass({
    * Sets the state to have no focused item.
    */
   _onBlur() {
-    this._focus(0, undefined);
+    this.setState({ blurred: true });
```
jasonLaster: hmm - it'd be nice to avoid `blurred`... is there a way to check the blur state in `shouldComponentUpdate` and perhaps previous blur state... we do have access to the element

jasonLaster:
closing as i'd like  to avoid keep blurred state.

PR #261, opened 2017-03-21
Add properties loader

### LINKED ISSUE #18: [LaunchPad] L10N should load properties file
copy of  https://github.com/devtools-html/debugger.html/issues/1030

---

We currently load our translation strings from public/js/strings.json. This works well, but it does mean we need to maintain two sets of strings in github and m-c.

It would be better if we could load strings from debugger.properties and when we publish copy the debugger.properties file to m-c.

This [patch](https://gist.github.com/jasonLaster/416d43026dc6620d9e567cfd68426b7d). Is close to working.

jasonLaster:
Resolves #18, which will let us use `properties` files directly and bypass JSON files.

Here's how the debugger will change: 

```diff
diff --git a/src/main.js b/src/main.js
index 8d556f5..12b04a0 100644
--- a/src/main.js
+++ b/src/main.js
@@ -33,7 +33,7 @@ const actions = bindActionCreators(require("./actions"), store.dispatch);

 if (!isFirefoxPanel()) {
   window.L10N = L10N;
-  window.L10N.setBundle(require("./strings.json"));
+  window.L10N.setBundle(require("../assets/panel/debugger.properties"));
```

CC @gasolin


packages/devtools-launchpad/src/utils/L10N.js
```diff
@@ -1,9 +1,12 @@
 // @flow
 
 const { sprintf } = require("devtools-modules");
+const { parse } = require("properties-parser");
+
 let strings = {};
 
 function setBundle(bundle: { [key: string]: string }) {
+  bundle = parse(bundle);
```
jasonLaster: this runtime strategy is borrowed from [l10n.js](https://dxr.mozilla.org/mozilla-central/source/devtools/shared/l10n.js) which uses a similar runtime parser.

It would be nice to do this statically in a webpack loader, but it ultimately does not make a big difference as this is just for the web environment

PR #267, opened 2017-03-23
(WIP) Update source map worker

jasonLaster:
* stop passing using devtools-config and instead pass it in
* explicitly start the worker 
* start using worker-utils

this is blocked on adding worker utils and bumping the dependency

### DIFF
--- a/packages/devtools-source-map/src/index.js
+++ b/packages/devtools-source-map/src/index.js
@@ -1,13 +1,14 @@
 // @flow
 
-const { getValue } = require("devtools-config");
 const {
   originalToGeneratedId,
   generatedToOriginalId,
   isGeneratedId,
   isOriginalId,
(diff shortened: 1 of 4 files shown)


packages/devtools-source-map/src/index.js
```diff
@@ -1,29 +1,23 @@
...
+
 import type { Location } from "devtools-client-adapters/src/types";
 
 let sourceMapWorker;
-function restartWorker() {
-  if (sourceMapWorker) {
-    sourceMapWorker.terminate();
-  }
-  sourceMapWorker = new Worker(getValue("sourceMapWorkerURL"));
+function startWorker(url: string) {
```
jryans: Can we pair the names? Either start and stop or init and destroy. The current start and destroy seems strange to me.

PR #269, opened 2017-03-24
Simplify transform pattern

jasonLaster:
This dramatically simplifies our JS transform strategy to be: exclude all node modules except for our own.


jasonLaster:
Yeah - we really want to switch to using an NPM org or something like that so that we could make the packages publicly available under `@devtools/splitter` or `@ff-devtools/splitter` or something like that.

PR #274, opened 2017-03-26
Fix cyclic dependency in WebConsoleClient and LongStringClient

rickychien:
FF devtools loader allows you to require cyclic dependency but it is broken in node.

A workaround fix for solving cyclic dependency in WebConsoleClient and LongStringClient, which causes { LongStringClient } = require() always return empty object within node module system.


jasonLaster:
Oh interesting. Didn't realized the loader supported that.

Issue #109, opened 2017-01-20
[CI] on windows


jasonLaster:
We should run our unit tests on windows as well as unix

We need to ask https://www.appveyor.com for a free account

clarkbw:
I granted access to the org for appveyor so we can start using it.  They offer free accounts for open source projects.

clarkbw:
Here we go: https://ci.appveyor.com/project/clarkbw/devtools-core

Garbee:
@clarkbw Fantastic! Thank you so much.

clarkbw:
I didn't do any of the setup so there might be configs needed but it should run on all of our PRs now.

PR #280, opened 2017-03-26
Appveyor

clarkbw:
I made Appveyor available in #109 but without a YAML file it hasn't actually been running.  Trying this out.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,19 +1,21 @@
-build:
-  verbosity: minimal
-
 environment:
-  nodejs_version: "7.0"
-
+  matrix:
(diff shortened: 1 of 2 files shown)


clarkbw:
There was an error in the `packages/devtools-launchpad/bin/validate-deps.js` file where the path wasn't being normalized for Windows.  Added 9a8cde8 to fix that.

clarkbw:
Looks like the issue is actually with the `parse-yarn-lock` module.

clarkbw:
Closing this out, I'm not going to be able to pick this up again for a while.

PR #281, opened 2017-03-27
add search-box component into sham-modules

gasolin:
add search-box and tabs component into `devtools-sham-modules` since they have dependencies in `sham-modules`.

put requires in alphabet order.

(I'm not sure if its fine to add shortcut in index.js, with that way the unused `search-box` component will be bundled into `debugger.html`)

### DIFF
--- a/packages/devtools-sham-modules/client/shared/components/search-box.js
+++ b/packages/devtools-sham-modules/client/shared/components/search-box.js
@@ -1,0 +1,110 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global window */
+
+"use strict";
+
+const { DOM: dom, createClass, PropTypes } = require("react");
+const { KeyShortcuts } = require("../key-shortcuts");
+
+/**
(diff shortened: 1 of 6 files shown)


gasolin:
The patch works with Netmonitor.html prove-of-concept version https://github.com/gasolin/devtools-core/tree/netmonitor.html/packages/netmonitor

And we plan to use the same module dependency in gecko (via webpack alias)

wldcordeiro:
@gasolin odd question but do you think we could maybe reuse our [searchbar](https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/SearchBar.js) in the debugger.html project or [tabs](https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/Tabs.js) component? It might be a tall order to do this change but I'm open to helping. Seems weird to have an implementation of tabbed interface in the `devtools-core` and one in the debugger.

jasonLaster:
Thanks @gasolin.

@wldcordeiro we can reconcile the different search bars at a later point. Right now the netmonitor team is mostly looking to be in a position where they can easily use the launchpad.

gasolin:
@wldcordeiro thanks for aware that. At the end we should have 1 component to deal with similar things. I'll create an issue to note that, but for current phase we can keep these components as it is, focus on make things happen, then make the code base looks like what it should be.

Issue #208, opened 2017-02-23
[launchpad] tabs do not appear after firefox is launched


jasonLaster:
When firefox is launched with the button, the UI needs to know to `connect` to it and then request tabs.

PR #284, opened 2017-03-27
reconnect to firefox client again to get tab updates

bomsy:
Associate issue #208

### DIFF
--- a/packages/devtools-launchpad/src/index.js
+++ b/packages/devtools-launchpad/src/index.js
@@ -145,7 +145,7 @@ async function connectClients(actions) {
 }
 
 async function getTabs(actions) {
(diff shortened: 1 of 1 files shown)


jasonLaster:
heh - this fixes a bug where we don't see the tabs. but if we are connected, we should just be able to fetch the latest tabs...

bomsy:
@jasonLaster do you mean `debuggerClient` rather than `threadClient` ? If so, even if there is a `debuggerClient` it won't have the latest tabs,  we will always need to reconnect to the socket to get the latest list of tabs.

jasonLaster:
Ahh. I see. Okay.

PR #287, opened 2017-03-28
Move client to launchpad

jasonLaster:
This is a sibling PR to https://github.com/devtools-html/debugger.html/pull/2478


packages/devtools-launchpad/src/client/chrome.js
```diff
@@ -79,37 +75,28 @@ async function connectNodeClient() {
...
-    .then(conn => {
-      connection = conn;
-      window.addEventListener("beforeunload", () => {
-        connection.onclose = function disable() {};
-        connection.close();
-      });
-    });
-}
+async function connectNode(tab: ChromeTab) {
+  const tabConnection = CDP({ tab: tab.webSocketDebuggerUrl });
```
bomsy: Should this be `await CDP({ tab: tab.webSocketDebuggerUrl });` ?

jasonLaster: yes. honestly, I have not tested the chrome and node functionality yet

packages/devtools-network-request/package.json
```diff
@@ -1,14 +0,0 @@
-{
-  "name": "devtools-network-request",
```
jryans: `devtools-network-request` is currently also used by `devtools-source-map` for fetching mapping files. If you want to remove the package, we should also update `devtools-source-map` as well to stop using it.

jasonLaster: Yes - that's a dependency of this work as well.

packages/devtools-launchpad/src/client/chrome.js
jasonLaster: tested chrome and node. Thanks for the catch

flow-typed/debugger-html.js
```diff
@@ -0,0 +1,216 @@
+  // @flow
```
jasonLaster: I added this file to share the debugger types. Going forward, i think it would be nice to have the [lib-def](https://flow.org/en/docs/libdefs/) exported by the debugger, but that is not a blocker for this otherwise massive PR.

packages/devtools-source-map/src/index.js
```diff
@@ -11,7 +11,7 @@ const {
 const { workerUtils: { workerTask }} = require("devtools-modules");
 const { setConfig, getValue } = require("devtools-config");
 
-import type { Location } from "devtools-client-adapters/src/types";
+import type { Location } from "debugger-html";
```
jryans: `devtools-source-map` doesn't depend on the debugger, so it won't have access to this, if I am understanding correctly. Did you learn anything more about how types are shared in the Flow community?

jasonLaster:
`debugger-html` is a [lib-def](https://flow.org/en/docs/libdefs/) defined in `flow-types/debugger-html.js`. We'll want to move the def into the debugger repo and expose it later, but i don't think it's a blocker for now.

packages/devtools-source-map/src/source-map.js
```diff
@@ -5,7 +5,7 @@
  * @module utils/source-map-worker
  */
 
-const networkRequest = require("devtools-network-request");
+const { networkRequest } = require("devtools-modules");
```
jasonLaster: this replaces the `devtools-network-request` with `devtools-modules`. I'll update the webpack.config.devtools in a minute to reflect this

flow-typed/debugger-html.js
clarkbw: Agreed, we should open up an issue for that.

jasonLaster:
I had to revert the splitter css require because `devtools-source-map` now requires `devtools-modules` and the tests were failing. I can re-PR that as a follow up

PR #289, opened 2017-03-29
Set source map worker URL in Firefox

jryans:
This adds a Firefox panel config for `devtools-source-map`, which is used when this is landed as a bundle in m-c.

Since we're landing a standalone bundle in m-c, we can't rely on config values from apps like the debugger anymore, so we need to set config data (currently just the worker URL) in this package.

### DIFF
--- a/packages/devtools-source-map/README.md
+++ b/packages/devtools-source-map/README.md
@@ -14,4 +14,4 @@ This is used in multiple contexts:
 # Application Requirements
 
 This package assumes that an application using this code will make the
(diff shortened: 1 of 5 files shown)


packages/devtools-source-map/README.md
```diff
@@ -14,4 +14,4 @@ This is used in multiple contexts:
 # Application Requirements
 
 This package assumes that an application using this code will make the
-`worker.js` file available at application specified URL `sourceMapWorkerURL`.
```
jasonLaster: thanks

PR #290, opened 2017-03-29
Bundle missed devtools-modules for netmonitor

gasolin:
* PluralForm --> devtools-modules
* curl --> devtools-modules
* file-saver --> devtools-modules
* clipboard --> devtools-modules

The PluralForm now only support English-like plural only. Need better L10N support to pass different plural local once we need to deal with the L10N

### DIFF
--- a/packages/devtools-launchpad/src/index.js
+++ b/packages/devtools-launchpad/src/index.js
@@ -6,8 +6,8 @@ const { Provider } = require("react-redux");
 const { DevToolsUtils, AppConstants } = require("devtools-sham-modules");
 const { debugGlobal } = require("./utils/debug");
 const { setConfig, getValue, isDevelopment } = require("devtools-config");
-const L10N = require("./utils/L10N");
 const defer = require("./utils/defer");
+const L10N = require("./utils/L10N");
 const { showMenu, buildMenu } = require("./menu");
 
 setConfig(DebuggerConfig);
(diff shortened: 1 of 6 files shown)


gasolin:
@jasonLaster PluralForm now works more like a mockup than the fully functional lib.
The origin version have to load the `Intl.properties` to get language specified plural form. 
Do you think it still worth moved to LaunchPad?

jasonLaster:
Yea - launchpad makes sense.

packages/devtools-launchpad/src/utils/plural-form.js
```diff
@@ -0,0 +1,193 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jasonLaster: I'm sorry, I think I misunderstood the purpose of `PluralForm`. My assumption was that it would be used by L10N and not exposed directly. In this case, we should move it back to `devtools-modules` or `devtools-sham-modules`

packages/devtools-launchpad/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "devtools-launchpad",
-  "version": "0.0.60",
+  "version": "0.0.61",
```
jasonLaster: lets do the version bumps separately

packages/devtools-modules/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "devtools-modules",
-  "version": "0.0.18",
+  "version": "0.0.19",
```
jasonLaster: lets do the version bumps separately

packages/devtools-launchpad/src/utils/plural-form.js
gasolin: ok~

PR #291, opened 2017-03-29
Bundle missed devtools-sham-modules for netmonitor

gasolin:
keycodes, Charts, tooltips, view-helpers, async-utils, task, defer

Debugger.html may able to share the same `task`, `defer` as well


wldcordeiro:
The `defer` util is already in devtools-core https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/src/utils/defer.js we could probably just move it around to make it work for both debugger and inspector.

gasolin:
We don't have clear module dependency now, but I guess we can remove the defer from `devtools-launchpad` and use this version instead

packages/devtools-sham-modules/shared/task.js
```diff
@@ -0,0 +1,515 @@
...
+ *   task.  This makes it possible to call an externally provided function and
+ *   spawn a task from it, regardless of whether it is an asynchronous generator
+ *   or a synchronous function.  This comes in handy when iterating over
+ *   function lists where some items have been converted to tasks and some not.
+ */
+
+////////////////////////////////////////////////////////////////////////////////
+//// Globals
+
+const Promise = require("promise");
```
jasonLaster: do we depend on `promise` in the package.json?

gasolin: Thanks, it should be `sham/promise`

PR #295, opened 2017-03-29
Document publish step for source maps

jryans:


jryans:
Good idea, added that as well.

PR #298, opened 2017-03-30
Support LocalizationHelper

rickychien:


packages/devtools-modules/shared/l10n.js
```diff
@@ -0,0 +1,254 @@
...
+// here is to tell webpack some constraints so that it can include
+// fewer files in the pack.
+//
+// Here we introduce new require contexts for each possible locale
+// directory.  Then we use the correct context to load the property
+// file.  In the webpack case this results in just the locale property
+// files being included in the pack; and in the devtools case this is
+// a wordy no-op.
+const reqShared = require.context("raw!devtools/shared/locales/",
+                                  true, /^.*\.properties$/);
```
gasolin: "raw!devtools/shared/locales/` seems search for `devtools/modules/`, so we need change that assertion

gasolin:
We can add `const { LocalizationHelper } = require("./shared/l10n");` in index.js for easier alias

jasonLaster:
@rickychien Hey - looks like the source-map test is failing

rickychien:
```
TypeError: require.context is not a function
      at Object.<anonymous> (../node_modules/devtools-modules/shared/l10n.js:30:27)
```

require.context is a webpack function and I copy that from https://searchfox.org/mozilla-central/source/devtools/shared/l10n.js#29-35

@jasonLaster do you get any clue about that?

rickychien:
```
const reqShared = require.context("devtools/shared/locales/",
                                  true, /^.*\.properties$/);
const reqClient = require.context("devtools/client/locales/",
                                  true, /^.*\.properties$/);
const reqGlobal = require.context("toolkit/locales/",
                                  true, /^.*\.properties$/);
```
Not sure why source-map cannot evaluate `require.context`. I suspect that is a webpack magic and webpack will convert this syntax sugar something useful.

PR #301, opened 2017-03-30
[WIP] kill shams

jasonLaster:
This is a WIP PR to both signal the intent to consolidate the packages and work out all of the kinks.

The longterm goal is to have more stable package boundaries:
* fewer cross dependencies.
* packages can be required directly
* utils and components are separated


packages/devtools-launchpad/src/utils/redux/middleware/promise.js
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const defer = require("../../defer");
+const {defer} = require("devtools-modules");
```
jryans: Use spaces inside braces to match other lines.

PR #303, opened 2017-03-31
Support LocalizationHelper in devtools-modules

rickychien:


rickychien:
Create a new PR for LocalizationHelper and CircleCI is green!

PR #309, opened 2017-04-04
Add resize callback to splitter

AnshulMalik:
Related issue: https://github.com/devtools-html/debugger.html/issues/2004

Parent component can specify a callback(onStopResize) to be called
whenever splitbox is resized.

### DIFF
--- a/packages/devtools-modules/index.js
+++ b/packages/devtools-modules/index.js
@@ -9,6 +9,7 @@ const MenuItem = require("./client/framework/menu-item");
 const networkRequest = require("./client/shared/shim/networkRequest");
 const PrefsHelper = require("./client/shared/prefs").PrefsHelper;
 const SearchBox = require("./client/shared/components/search-box");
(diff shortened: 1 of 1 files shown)


packages/devtools-modules/client/shared/components/splitter/SplitBox.js
```diff
@@ -99,6 +101,12 @@ const SplitBox = React.createClass({
     doc.documentElement.style.cursor = this.state.defaultCursor;
 
     splitBox.classList.remove("dragging");
+
+    if (this.props.onStopResize) {
+      this.props.onStopResize(
+        (this.state.vert) ? this.state.width : this.state.height
+      );
+    }
```
jasonLaster: hmm, can we rename this to `onResizeEnd`

jasonLaster:
@AnshulMalik Oh yeah - we also created a new splitter package :)

PR #316, opened 2017-04-08
Add SourceEditor to devtools-modules

rickychien:
This module is basically a react component wrapper which handles source editor (code mirror). This is designed to be a shared component which is able to be shared with debugger and netmonitor...anyone else who needs to embed source editor feature.

Btw, I'm not sure this should be treated as a separate module like devtools-splitter.


jasonLaster:
Thanks. Lets make this its own package though so it can be more easily versioned.

Some guidelines I'm using for whether something should be a package:

* does it have many deps on other modules (if now, this is a good sign that it's own package)
* do other tools require it directly (if yes, that's a good sign that it is its own thing)

PR #320, opened 2017-04-08
Launchpad: add menu.css some small fixes

jasonLaster:


jasonLaster:
@rickychien let me know if this causes any issues

packages/devtools-launchpad/src/components/shared/menu.js
```diff
@@ -1,6 +1,8 @@
 const { Menu, MenuItem } = require("devtools-modules");
 const { isFirefoxPanel } = require("devtools-config");
 
+require("./menu.css");
```
nt1m: We only want to import this CSS file when isFirefoxPanel is false.

PR #322, opened 2017-04-09
Export source editor instance

rickychien:


jasonLaster:
Hmm, how's this different than requiring it directly?

The package.json already points to src/source-editor...
https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-source-editor/package.json#L10

rickychien:
Cool. I wasn't aware of main attribute in package.json. It works in my case. Thanks!

PR #323, opened 2017-04-09
Fix reqClient error

rickychien:
netmonitor ran in to this error and launchpad shows empty page after we removed these webpack hack in previous PR.

### DIFF
--- a/packages/devtools-modules/shared/l10n.js
+++ b/packages/devtools-modules/shared/l10n.js
@@ -27,12 +27,12 @@ const propertiesMap = {};
 // file.  In the webpack case this results in just the locale property
 // files being included in the pack; and in the devtools case this is
 // a wordy no-op.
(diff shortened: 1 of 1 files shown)


jasonLaster:
Thanks @rickychien. 

Lets see how the inspector used it earlier. http://searchfox.org/mozilla-central/source/devtools/client/inspector/webpack.config.js

Also, maybe @tromey, @gregtatum or @juliandescottes could explain how we can setup the context fields so that it works in the launchpad.

gregtatum:
Reading through it, I don't really know the context for what's going on and where the issue is.

rickychien:
reqClient hack is added at http://searchfox.org/mozilla-central/source/devtools/shared/l10n.js#29-34

jasonLaster:
Yep. I'm not sure why it breaks the launchpad when added. If you can
research require.context and let me know how to add it, I'm happy to help

PR #325, opened 2017-04-10
Fix to send correct args to Windows FF

LeslieMurphy:
Fix bug where "yarn firefox" would not open port 6080 for remote debug on Windows platform


jasonLaster:
Thanks, I'll merge it and we'll see if it helps windows users

PR #326, opened 2017-04-10
Atom-like style of foldgutter

AnshulMalik:
Issue: https://github.com/devtools-html/debugger.html/issues/2588


jasonLaster:
Looks great Anshul

PR #329, opened 2017-04-10
move source editor utils

jasonLaster:


packages/devtools-source-editor/index.js
```diff
@@ -0,0 +1,4 @@
+import SourceEditor from "./src/source-editor";
```
rickychien: I saw console threw a error message `SyntaxError: import declarations may only appear at top level of a module`.

jasonLaster: Hmm. Switch it to require. Sorry.

rickychien: Thanks!

PR #336, opened 2017-04-15
[launchpad] fix menus #334

### LINKED ISSUE #334: [launchpad] fix menus
Looks like the latest patch to menus broke Menus in the launchpad's settings pane.

Lets try and get a fix in along with a unit test to prevent future regressions.

#### screenshots
|before|after|
|---|--|
|||

rickychien:
Fix #334

### DIFF
--- a/packages/devtools-launchpad/src/components/shared/menu.js
+++ b/packages/devtools-launchpad/src/components/shared/menu.js
@@ -71,11 +71,13 @@ function showMenu(evt, items) {
   }
 
   let menu = new Menu();
-  items.filter((item) => item.visible).forEach((item) => {
-    let menuItem = new MenuItem(item);
-    menuItem.submenu = createSubMenu(item.submenu);
-    menu.append(menuItem);
-  });
+  items
+    .filter((item) => item.visible === undefined || item.visible === true)
+    .forEach((item) => {
+      let menuItem = new MenuItem(item);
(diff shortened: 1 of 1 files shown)


packages/devtools-launchpad/src/components/shared/menu.js
```diff
@@ -71,11 +71,13 @@ function showMenu(evt, items) {
   }
 
   let menu = new Menu();
-  items.filter((item) => item.visible).forEach((item) => {
-    let menuItem = new MenuItem(item);
-    menuItem.submenu = createSubMenu(item.submenu);
-    menu.append(menuItem);
-  });
+  items
+    .filter((item) => item.visible === undefined || item.visible === true)
```
gasolin: should we do the same check for `subMenu.items.filter` ?

gasolin: & it will be nice to provide some unit test to protect the menu not be broken again

rickychien: subMenu.items.filter fixed!

I know devtools-core is very unstable and sometime the new PR broke usability. But I think we haven't setup unit test or integration test framework in devtools-core yet. We can add a test suite for menu once testing framework is settle down.

jasonLaster:
Merging as this fixes the issue. Thanks @rickychien, I'll try and land the same jest setup we have in other places here so we can get some  test coverage

PR #340, opened 2017-04-20
[Tree] Navigate to next node on right arrow key

irfanhudda:
Navigate to next node when right arrow key is pressed and focused node is a directory. (This is consistent with how chrome devtools and VS code handle right arrow key in source tree)

### Summary of changes
* Focus next node when right arrow key is pressed and focused node is directory.

### DIFF
--- a/packages/devtools-modules/client/shared/components/tree.js
+++ b/packages/devtools-modules/client/shared/components/tree.js
@@ -519,6 +519,8 @@ const Tree = module.exports = createClass({
       case "ArrowRight":
         if (!this.props.isExpanded(this.props.focused)) {
           this._onExpand(this.props.focused);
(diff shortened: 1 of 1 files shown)


jasonLaster:
This might seem strange, but that's the desired behavior: 

https://github.com/devtools-html/devtools-core/commit/04078c0581296568ce4a43ea0b52df092a45e87c

We want to be closer to editors

irfanhudda:
@jasonLaster Interesting.

Both atom and VS code select the next subentry of current directory if it is already expanded. 

|Atom|VS code|
|------|------|
|atom|vscode|

I also seem to remember that right arrow in sublime just expands the tree like current behaviour.

https://github.com/atom/tree-view/pull/805

PR #345, opened 2017-04-24
Remove hasMappedSource

### LINKED ISSUE #311: hasMappedSource is unused
Seeing this in eslint errors.

https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-source-map/src/source-map.js#L195-L202

tromey:
hasMappedSource was not exported.  This removes it.
Fixes #311.


tromey:
I think this is wrong actually.  It should probably be exported for use in worker.js.

PR #360, opened 2017-05-05
Added license files and headers in various packages

captainbrosset:
This is a start. I haven't done all of them yet. I'm waiting on more information to finalize launchpad, source-editor and source-map.
But at least those ones should be merged.

### DIFF
--- a/packages/devtools-components/LICENSE
+++ b/packages/devtools-components/LICENSE
@@ -1,0 +1,362 @@
+Mozilla Public License, version 2.0
+
+1. Definitions
+
+1.1. "Contributor"
+
+     means each individual or legal entity that creates, contributes to the
+     creation of, or owns Covered Software.
+
+1.2. "Contributor Version"
+
+     means the combination of the Contributions of others (if any) used by a
+     Contributor and that particular Contributor's Contribution.
+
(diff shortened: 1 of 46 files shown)


jasonLaster:
@captainbrosset thanks for looking into licensing.

+ the package.json change is great
+ can we follow up with legal to see if we can avoid the license header on these files? The files will land in m-c as a bundle and this would actually be duplicated a couple hundred times :)

```
/* This Source Code Form is subject to the terms of the Mozilla Public
```

captainbrosset:
> can we follow up with legal to see if we can avoid the license header on these files? The files will land in m-c as a bundle and this would actually be duplicated a couple hundred times :)

I'm in contact with legal already about other license-related questions. I haven't asked this question specifically, but my understanding of the MPL license is that these headers are necessary (see [Q4 of this FAQ](https://www.mozilla.org/en-US/MPL/2.0/FAQ/)).

Could we maybe run a tool during the bundling process that removes the licenses, or all comments, or even uglifies the code?

This problem seems like something a tool should solve.

captainbrosset:
Just heard from legal that license headers in all files "is indeed a good practice to maintain".

PR #363, opened 2017-05-09
License check

tromey:


packages/devtools-launchpad/package.json
```diff
@@ -14,6 +14,7 @@
   "scripts": {
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
+    "license-check": "../devtools-license-check/bin/devtools-license-check",
```
jasonLaster: we should be able to replace this with. 

```
 "license-check": "devtools-license-check"
```

The only setup needed is that each component has to have `devtools-license-check` as a devDependency. Unfortunately, you need to make a `0.0.1` release to get this working the first time but after that it's actually symlinked by lerna so you can make changes to `devtools-license-check` and test it another package.

packages/devtools-license-check/bin/devtools-license-check
```diff
@@ -0,0 +1,94 @@
...
+    console.error("ERROR: " + err);
+    process.exit(1);
+  }
+
+  let exitCode = 0;
+  for (let dep in licenseInfo) {
+    if (WAIVED_DEPENDENCIES.includes(dep)) {
+      continue;
+    }
+    if (!isValidLicense(licenseInfo[dep].licenses)) {
```
tromey: I have a subsequent patch to do this - it wasn't needed by devtools-core, just debugger.  But I can roll it in.

packages/devtools-launchpad/package.json
tromey: Does that mean we'd have to do a release each time we need to change the license checker, even to get it working for devtools-core?

tromey: I guess not since you also say it's symlinked after that.  I don't understand why it would be this way.

tromey:
I'm going to a separate PR to land just the checker, then do a release of that, then do subsequent PRs to actually use it.

packages/devtools-license-check/bin/devtools-license-check
juliandescottes: Interesting, I got it just by running it on devtools-core!

PR #364, opened 2017-05-10
Add a license checker

tromey:


packages/devtools-license-check/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "devtools-license-check",
+  "version": "0.1.0",
+  "description": "Verify licenses for inclusion in Mozilla DevTools",
+  "repository": "github:devtools-html/devtools-core",
+  "author": "",
```
juliandescottes: Missed that in the other review, but maybe fill in the author field with your name?

tromey: That didn't seem right to me because in the final patch, most of the code is from Jordan.  And, we have a few other packages with an empty string here.

PR #365, opened 2017-05-10
Use the license checker

tromey:


clarkbw:
But who's checking the license checker... ? 

😨

tromey:
There didn't seem to be a need to check the license checker, because it is only ever going to be a dev dependency.  But, it's easy to add if you want me to.

clarkbw:
> But, it's easy to add if you want me to.

Nope! Just kidding 😄

tromey:
> We apparently need to add devtools-license-checkas a devDependency on each and every package if we want the script to work when being run from a package itself.

Jason warned about this in the other PR, but it didn't occur to me to test it individually; doing it via `lerna` works.  I don't know if there are downsides to adding this as a dev dep everywhere, but if not I can do it tomorrow.

juliandescottes:
The only downside I can think of is that we will be duplicating what seems to me like a redundant information. Especially the version of devtools-license-check ... having to update it in all 10 packages every time we release a new version :(

That's why I would be fine if we said that we only ever want to run it from devtools-core root.

tromey:
> That's why I would be fine if we said that we only ever want to run it from devtools-core root.

Ok, I will just merge it then.  If it proves to be annoying not to be able to do it individually, feel free to just assign a bug to that effect to me.

PR #376, opened 2017-05-22
source maps in m-c

jasonLaster:
this will add source maps when we make a bundle in m-c which will make them debuggable


packages/devtools-launchpad/webpack.config.devtools.js
```diff
@@ -20,7 +20,7 @@ module.exports = (webpackConfig, envConfig) => {
     webpackConfig.output.path = outputPath;
   }
 
-  webpackConfig.devtool = false;
+  // webpackConfig.devtool = false;
```
tromey: I find it better to just remove code rather than comment it out.  The other hunks in this file seem unrelated.

PR #378, opened 2017-05-23
Devtools source map license

captainbrosset:
The devtools-source-map package already has the right MPL-2.0 license in package.json, so we're already good.
But it doesn't have a LICENSE file or license headers in source files. MPL-2.0 requires this, and this is what this PR does.

PR #379, opened 2017-05-24
Make validate-deps use the default yarn parser

captainbrosset:
This makes the `validate-deps.js` script work for me on Windows.

It used to fail trying to get the yarn lock file parser before. Turns out the `parse-yarn-lock` module we were using didn't work for me on Windows. I think @clarkbw had already [reported it here](https://github.com/cdaringe/parse-yarn-lock/pull/1) but I'm not seeing a resolution to this. I think this was itself waiting on another [yarn fix](https://github.com/yarnpkg/yarn/pull/3397) that hasn't landed.

I've been trying my luck at different options, and ended up trying to require the yarn lockfile parser directly, which seems to work. No need to have a dependency on yarn in `package.json` it seems, we can just use the parser.

One thing though, my SublimeText editor was viewing the file with Windows line endings, which was making the parser fail because it expected all line endings to be `\n`. Once I fixed that, it worked great.

What do you think?
I need a fix, because right now that script is just failing for me and preventing me from running tests.

### DIFF
--- a/packages/devtools-launchpad/bin/validate-deps.js
+++ b/packages/devtools-launchpad/bin/validate-deps.js
@@ -1,27 +1,23 @@
-const fs   = require('fs');
+const fs = require("fs");
 const path = require("path");
+const parse = require("yarn/lib/lockfile/parse.js").default;
 
 function isUnique(list) {
-  const set = new Set(list)
+  const set = new Set(list);
   return set.size == list.length;
 }
 
-const parser = require('parse-yarn-lock')
-const lockFilePath = path.join(__dirname, '../yarn.lock');
-const lockfile = fs.readFileSync(lockFilePath).toString()
+const lockFilePath = path.normalize(path.join(__dirname, "..", "yarn.lock"));
+const lockfile = fs.readFileSync(lockFilePath).toString();
 
-parser.parse(lockfile, function (err, deps) {
-  if (err) throw err
+const parsed = parse(lockfile);
+const devtoolDeps = Object.keys(parsed).filter(dep => dep.match(/^devtools-/));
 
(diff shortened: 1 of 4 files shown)


captainbrosset:
Damn, turns out requiring the yarn lockfile parser directly doesn't work on CI:

```
Error: Cannot find module 'yarn/lib/lockfile/parse.js'
    at Function.Module._resolveFilename (module.js:472:15)
    at Function.Module._load (module.js:420:25)
    at Module.require (module.js:500:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/ubuntu/devtools-core/packages/devtools-launchpad/bin/validate-deps.js:3:15)
    at Module._compile (module.js:573:32)
    at Object.Module._extensions..js (module.js:582:10)
    at Module.load (module.js:490:32)
    at tryModuleLoad (module.js:449:12)
    at Function.Module._load (module.js:441:3)
```

captainbrosset:
Ok I fixed the issue by re-adding the dev dependency on yarn in `package.json`. The tests fail though, just like [in another PR that I've opened](https://github.com/devtools-html/devtools-core/pull/378). And I can't replicate this test failure locally yet. 
Help on this would be great, I'm not sure what's going on here.

captainbrosset:
The previous issue I was seeing last week is now gone. I just launched a new build and the devtools-launchpad tests succeeded. But now there's a new failure when running `yarn snapshot` on devtools-splitter.

captainbrosset:
And again, just restarting the build fixed the issue ... very mysterious. If anyone has an idea about what's happening here, that'd be great.
@jasonLaster do you mind taking a look at the code changes here and doing a review?

PR #380, opened 2017-05-25
Correctly resolve original locations relative to absolute source roots

tromey:
_setSourceMapRoot was not properly resolving a location relative to an
absolute source root.  This manifested in GWT Super Dev Mode (I really
can't type that enough) as a request for an invalid original source,
like:

http://127.0.0.1:9876/sourcemaps/webadmin/http:/127.0.0.1:9876/sourcemaps/webadmin/com/google/gwt/core/client/GWT.java

This is https://bugzilla.mozilla.org/show_bug.cgi?id=1355126

### DIFF
--- a/packages/devtools-source-map/src/source-map.js
+++ b/packages/devtools-source-map/src/source-map.js
@@ -7,7 +7,7 @@
 
 const { networkRequest } = require("devtools-utils");
 
-const { parse } = require("url");
+const { parse, resolve } = require("url");
 const path = require("./path");
 const { SourceMapConsumer, SourceMapGenerator } = require("source-map");
 const assert = require("./assert");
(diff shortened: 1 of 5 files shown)


packages/devtools-source-map/src/source-map.js
```diff
@@ -56,19 +56,8 @@ function _setSourceMapRoot(sourceMap, absSourceMapURL, source) {
     return;
   }
 
-  const base = path.dirname(
-    (absSourceMapURL.indexOf("data:") === 0 && source.url) ?
```
jryans: Is the `data:` case handled correctly? I can't tell...

tromey:
Patrick was also running in the launchpad test failure; anyway I think it's clearly not a problem with this patch.

jryans:
(Definitely one of the stranger URLs I've seen in a while!)

packages/devtools-source-map/src/source-map.js
tromey: Probably not, but this is a pathological situation: a source map claiming that the sources are at some `data:` URL, without providing the sources inline.  I can add the old special case back, but perhaps this should just be an error of some kind instead.

tromey: Actually I see I have that backward; I guess this kind of makes sense then.

PR #383, opened 2017-05-26
Source map url resolution

tromey:
This isn't the last time we'll have to work on source map URL resolution, but it fixes up something I introduced yesterday and lets us make a bit of progress.

### DIFF
--- a/packages/devtools-source-map/src/source-map.js
+++ b/packages/devtools-source-map/src/source-map.js
@@ -7,7 +7,6 @@
 
 const { networkRequest } = require("devtools-utils");
 
-const { parse, resolve, format } = require("url");
 const path = require("./path");
 const { SourceMapConsumer, SourceMapGenerator } = require("source-map");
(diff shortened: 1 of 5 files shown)


packages/devtools-source-map/src/source-map.js
```diff
@@ -36,7 +35,7 @@ function _resolveSourceMapURL(source: Source) {
   } else if (path.isAbsolute(sourceMapURL)) {
     // If it's an absolute path, it should be resolved relative to the
     // host of the source.
-    const { protocol = "", host = "" } = parse(url);
+    const { protocol = "", host = "" } = new URL(url);
```
jryans: This [node.js doc][1] suggests you are meant to acquire `URL` from the `url` module, but I guess it must be a global too if this change worked in the test environment...

[1]: https://nodejs.org/api/url.html#url_the_whatwg_url_api

PR #387, opened 2017-05-29
Added compatibility with Chrome on Windows for devtools-launchpad package

zaggy:
Hi,

I started to use debugger.html on Windows with Chrome. I noticed that the function start-chrome is not working on Windows, so I decided to fix that.

This is a very naive first implementation but it is working. I think that later this needs to be changed to selenium-webdriver, as it was done with firefox. 

This is my first PR to any open source project, so your feedback would be appreciated. Also, I would really love to see it merged as I need this function. 

Thanks in advance,
Sergey Lymar

### DIFF
--- a/packages/devtools-launchpad/bin/chrome-driver.js
+++ b/packages/devtools-launchpad/bin/chrome-driver.js
@@ -2,8 +2,14 @@
 
 const spawn = require('child_process').spawn;
 const minimist = require("minimist");
+const os = require("os");
 
-const chromeBinary = "/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome"
+const isWindows = /^win/.test(process.platform);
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks @zaggy, your change looks great and I'm glad that you're interested in debugger.html + chrome debugging. This is something that has been experimental for some time that I'd like to focus on in the summer: july, august.

Feel free to jump in slack and we can talk about next steps

jasonLaster:
your test failures look like something that existed on master

zaggy:
I can focus on fixing the tests, if you wish. But those need to be changed a bit so that they work with Windows too.

jasonLaster:
re-opening here: https://github.com/devtools-html/devtools-core/pull/409

PR #388, opened 2017-06-01
Adding the missing LICENSE file and headers in source-editor

captainbrosset:
The code in this package originated on mozilla-central, under the MPL2.0 license, and also mentioned MPL2.0 in the package.json file. It only lacked the LICENSE file at the root as well as license headers at the top of each source file.
This is what this PR does.

### DIFF
--- a/packages/devtools-source-editor/LICENSE
+++ b/packages/devtools-source-editor/LICENSE
@@ -1,0 +1,362 @@
+Mozilla Public License, version 2.0
+
+1. Definitions
+
+1.1. "Contributor"
+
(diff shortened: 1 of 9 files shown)


tromey:
Those license errors are nothing to be concerned with.  I've investigated them and I'll be sending a license-check PR shortly.

captainbrosset:
Thanks very much Tom! I was a bit worried by that but also surprised that my change would cause this failure. I couldn't reproduce it locally, and the licenses it complains about seem to be valid anyway. So I'll merge this for now. Thanks both.

PR #389, opened 2017-06-01
Update list of licenses and waivers

tromey:
This updates the list of licenses and waivers based on investigation of
some failures in https://github.com/devtools-html/devtools-core/pull/388


tromey:
I'll publish 0.3.0 when this lands.

PR #391, opened 2017-06-04
[WIP] Added color variables for conditional breakpoint text color

zaggy:
Trying to polish conditional breakpoint panel, as per the the issue https://github.com/devtools-html/debugger.html/issues/2918 and thought that it would be better to introduce a new variable. In light and firebug themes the color is unchanged, but dark theme color was changed to something more contrast and pleasant to the eye. Should look like this in the end:

|Current colors|Colors to be|
|----------|------|
|current|to be|

### DIFF
--- a/packages/devtools-launchpad/src/lib/themes/variables.css
+++ b/packages/devtools-launchpad/src/lib/themes/variables.css
@@ -29,6 +29,7 @@
   --theme-splitter-color: #dde1e4;
   --theme-comment: #696969;
   --theme-comment-alt: #ccd1d5;
+  --theme-conditional-breakpoint-text-color: #ccd1d5;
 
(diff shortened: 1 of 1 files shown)


zaggy:
Tests are still failing and it wasn't me)

clarkbw:
Hmm, maybe @tromey knows, it seems to fail on the license check.

jasonLaster:
@zaggy mind re-opening this in debugger.html.

We've started using debugger.html variables: 
* https://github.com/codehag/debugger.html/blob/bug/2904-breakpoints-sourcemaps-breakpoints-are-out-of-sync-with-the-server-on-reload/src/components/Editor/Editor.css#L1

PR #407, opened 2017-06-05
Move Reps into devtools-core

nchevobbe:
I think this is all good, the history should be preserved.


packages/devtools-reps/.eslintrc.js
```diff
@@ -0,0 +1,481 @@
+"use strict";
```
tromey: After landing we should compare this to devtools-core's .eslintrc, fix up the differences, and then (probably) just remove this file.

nchevobbe:
Circle is happy now (thanks @jasonLaster !)
Let's merge this.

PR #414, opened 2017-06-07
Allow to pass parameter names to Function Rep

### LINKED ISSUE #413: Merge debugger's previewFunction into function Rep
The debugger is doing some things to render function in ObjectInspector : https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/previewFunction.js .

It would be nice if we could merge back the thing that the debugger needs into reps, by allowing to pass some props (for example parameters names, which we can't get directly from the grip now).

Eventually, it would be nice to have those kind of data directly in the grip so all the tools can benefits from having this.

nchevobbe:
Fixes #413 

Added a props `parameterNames` for now since we don't have the data directly in the grip.
This way, the debugger will be able to use the Reps (since there we can retrieve parameter names from the AST)

@jbhoosreddy if you have some time I'd like you to review this PR since I think you're the one that created `previewFunction` , thanks !

---

Mochitest are passing locally

### DIFF
--- a/packages/devtools-reps/src/reps/function.js
+++ b/packages/devtools-reps/src/reps/function.js
@@ -22,6 +22,7 @@ const { span } = React.DOM;
 FunctionRep.propTypes = {
   object: React.PropTypes.object.isRequired,
   objectLink: React.PropTypes.func,
+  parameterNames: React.PropTypes.array,
 };
 
 function FunctionRep(props) {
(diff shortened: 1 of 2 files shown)


packages/devtools-reps/src/reps/function.js
```diff
@@ -51,7 +55,23 @@ function getTitle(props, grip) {
...
+  return cropString(name, 100);
+}
+
+function renderParams(props) {
+  const {
+    parameterNames = []
+  } = props;
+
+  return parameterNames
+    .filter(param => param)
```
nchevobbe: This is in `previewFunction` , I wonder if we should keep this.
From the debugger, can `parameterNames` have falsy values ?

Issue #415, opened 2017-06-07
Move Reps tests from mochitest to jest


nchevobbe:
At the moment, we have one mochitest for each Rep.
Those mochitest don't have any real value over jest since we are only doing shallow rendering with pre-generated stubs.

Having those tests converted to jest ones would allow us to:
- run the tests faster
- do not need to have a local copy of firefox to launch tests

~Here is the list of the tests we need to migrate :~ Nothing left !

DONE: 
- [x] ~test_reps_attribute.html~ 🏅 @sidthesloth92
- [x] ~test_reps_comment-node.html~ 🏅 @jakub-c
- [x] ~test_reps_date-time.html~ 🏅 @arashm
- [x] ~test_reps_document.html~ 🏅 @arashm
- [x] ~test_reps_element-node.html~ 🏅 @arashm
- [x] ~test_reps_error.html~ 🏅 @arashm 
- [x] ~test_reps_function.html~ 🏅 @arashm 
- [x] ~test_reps_infinity.html~ 🏅 @sidthesloth92
- [x] ~test_reps_failure.html~ 🏅 @sidthesloth92
- [x] ~test_reps_long-string.html~ 🏅 @jakub-c
- [x] ~test_reps_null.html~ 🏅 @sidthesloth92
- [x] ~test_reps_number.html~ 🏅 @arashm
- [x] ~test_reps_object-with-text.html~ 🏅 @amelzer
- [x] ~test_reps_promise.html~ 🏅 @arashm
- [x] ~test_reps_regexp.html~ 🏅 @arashm
- [x] ~test_reps_string.html~ 🏅 @arashm
- [x] ~test_reps_stylesheet.html~ 🏅 @arashm
- [x] ~test_reps_symbol.html~ 🏅 @arashm
- [x] ~test_reps_undefined.html~  🏅 @iammadhan
- [x] ~test_reps_window.html~ 🏅 @arashm
- [x] ~test_reps_object-with-url.html~ 🏅 @arashm
- [x] ~test_reps_object.html~ 🏅 @arashm
- [x] ~test_reps_nan.html~
- [x] ~test_reps_text-node.html~
- [x] ~test_reps_event.html~
- [x] ~test_reps_grip-map.html~
- [x] ~test_reps_grip-array.html~
- [x] ~test_reps_array.html~
- [x] ~test_reps_grip.html~

clarkbw:
[...]

PR #419, opened 2017-06-09
Move DateTime rep mochitest to jest

arashm:
Don't have much experience on React, just tried to follow the tutorial in #415. Hope it's alright.


packages/devtools-reps/src/reps/stubs/date-time.js
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+let stubs = new Map([]);
```
nchevobbe: I think we can remove the array inside, my bad for having it in the guide and in the NaN stub

nchevobbe:
Merged 🎉 Thanks again @arashm

arashm:
Cool! So I pick another test.

nchevobbe:
@arashm Maybe you could add a comment on the issue #415 to tell which test you're working on so other contributor don't take the same one :)
I'll add that in the guide too

arashm:
Good idea. Will do as you said.

PR #420, opened 2017-06-09
Fix menuitem padding

jasonLaster:
This updates the menuitem dom to accomodate windows CSS styling

https://github.com/devtools-html/debugger.html/issues/1942

### DIFF
--- a/packages/devtools-modules/src/menu/index.js
+++ b/packages/devtools-modules/src/menu/index.js
@@ -4,6 +4,10 @@
 
 const EventEmitter = require("../utils/event-emitter");
 
+function inToolbox() {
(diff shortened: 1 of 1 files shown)


packages/devtools-modules/src/menu/index.js
```diff
@@ -4,6 +4,10 @@
 
 const EventEmitter = require("../utils/event-emitter");
 
+function inToolbox() {
+  return window.parent.document.documentURI == "about:devtools-toolbox"
```
nt1m: Not sure if this check is future-proof, but if other devtoolers are ok with it, then it should be fine :)

PR #424, opened 2017-06-09
Create contextmenu package

jasonLaster:
Our contextmenu shim is currently inside the launchpad which is not quite the right place for it:

1. other tools that want the shim, but dont want the pad need to get the launchpad
2. when we make a debugger bundle we need to pull in the pad, which is silly

#### should the menu be in another package?

possibly, but it's nice to have a small package with its own tests because this is kinda an edge case in terms of tests...

#### should we move the devtools-modules libs in to the package?

I don't think so. `devtools-modules` modules are supposed to be synced with m-c, so they should stay where they are for now.


packages/devtools-contextmenu/README.md
```diff
@@ -0,0 +1,6 @@
+### DevTools ContextMenu
+
+
+The devtools contemenu is an HTML shim, which lets us create
```
bomsy: typo: contemenu => contextmenu

packages/devtools-contextmenu/README.md
```diff
@@ -0,0 +1,6 @@
...
+fake contextmenus in HTML namespaces and real contxt menus in the
```
bomsy: typo: real contxt => real context

packages/devtools-contextmenu/README.md
```diff
@@ -0,0 +1,6 @@
...
+fierfox toolbox.
```
bomsy: typo: fierfox => firefox

PR #427, opened 2017-06-09
Move StyleSheet rep mochitest to jest

arashm:
Based on #415.


nchevobbe:
ESLint and yarn test are green locally, let's merge this

PR #435, opened 2017-06-11
Move Promise rep mochitest to jest

arashm:
This was a little different from other files in #415 since it's using helper functions from other files. I tried to be as close as possible to the previous solutions and added a `test-helper` file. I don't know if what I've done is a acceptable approach or not.


packages/devtools-reps/src/reps/test-helpers.js
```diff
@@ -0,0 +1,80 @@
...
+const { REPS } = require("./rep");
+const { getGripPreviewItems } = require("./rep-utils");
+const { Rep } = REPS;
+const nodeConstants = require("../shared/dom-node-constants");
+
+class TestHelpers {
+  /**
+   * Test that a rep renders correctly across different modes.
+   */
+  static testRepRenderModes(modeTests, testName, componentUnderTest, gripStub,
```
nchevobbe: I'd prefer to completely remove this helper function.
I find it hard to evolve if needed without breaking other tests, and also, it adds a level of abstraction you have to deal with when writing your tests.
Having directly the code used to render the component directly in the test you're writing makes it simpler to know what's going on.

So, if we remove this function, we might want to architect our tests like https://github.com/devtools-html/devtools-core/pull/436/files#diff-b7f87095025dd0759eb9ce6b643dc158R40 (one `it` block by mode), or like https://github.com/devtools-html/devtools-core/pull/436/files#diff-b7f87095025dd0759eb9ce6b643dc158R40 (one `it` block for all modes).

packages/devtools-reps/src/reps/test-helpers.js
```diff
@@ -0,0 +1,80 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const { shallow } = require("enzyme");
+const { REPS } = require("./rep");
+const { getGripPreviewItems } = require("./rep-utils");
+const { Rep } = REPS;
+const nodeConstants = require("../shared/dom-node-constants");
+
+class TestHelpers {
```
nchevobbe: I think we should follow the same pattern we have in other files : have all of the helpers as plain functions and export them (so here we'd get rid of the class)

arashm:
@nchevobbe Take a look again. I fixed the mentioned issues.

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+  const stub = stubs.get("FulfilledWithNode");
+  const grips = getSelectableInInspectorGrips(stub);
+
+  it("has one node grip", () => {
+    expect(grips.length).toBe(1);
+  });
+
+  it("calls with expected argument when mouseover is fired on the node element", () => {
+    let mouseOverValue;
+    const onDOMNodeMouseOver = (object) => {
```
nchevobbe: Would you mind converting this to the "jest" way, which would look something like: 
```javascript
it("calls the expected function on mouseover", () => {
  const onDOMNodeMouseOver = jest.fn();
  const wrapper = renderRep(stub, {onDOMNodeMouseOver});

  wrapper.simulate("mouseover");

  expect(onDOMNodeMouseOver.mock.calls.length).toBe(1);
  expect(onDOMNodeMouseOver).toBeCalledWith(object);
});
```

`jest.fn()` gives us a way to control more deeply how and how many times the function was called (here's the doc for it https://facebook.github.io/jest/docs/mock-functions.html#content)

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+      mouseOverValue = object;
+    };
+
+    const renderedComponent = shallow(Rep({
```
nchevobbe: I think we can use :
```js
renderRep(stub, {onDOMNodeMouseOver})
```

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+  const stub = stubs.get("FulfilledWithNode");
+  const grips = getSelectableInInspectorGrips(stub);
+
+  it("has one node grip", () => {
+    expect(grips.length).toBe(1);
+  });
+
+  it("calls with expected argument when mouseout is fired on the node", () => {
+    let called = false;
+    const onDOMNodeMouseOut = (object) => {
```
nchevobbe: Could we use `jest.fn()` like I suggested for the `onDomMouseOver` test ?
It could be something like: 
```js
it("calls the expected function on mouseout", () => {
  const onDOMNodeMouseOut = jest.fn();
  const wrapper = renderRep(object, {onDOMNodeMouseOut});

  wrapper.simulate("mouseout");

  expect(onDOMNodeMouseOut.mock.calls.length).toBe(1);
});
```

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+      called = true;
+    };
+
+    const renderedComponent = shallow(Rep({
```
nchevobbe: I think we could use 
```js 
renderRep(object, {onDOMNodeMouseOut})
```

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+      object: stub,
+      onDOMNodeMouseOut,
+    }));
+
+    renderedComponent.find(".objectBox-node").simulate("mouseout");
+    expect(called).toBe(true);
+  });
+});
+
+describe("test on DOM inspect icon click", () => {
```
nchevobbe: this also could be merge in the previous `describer` block since we're operating on the same stub.

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+  const stub = stubs.get("FulfilledWithNode");
+  const grips = getSelectableInInspectorGrips(stub);
+  let inspectIconClickedValues = null;
+  const onInspectIconClick = (object) => {
+    inspectIconClickedValues = object;
+  };
+  const renderedComponent = shallow(Rep({
```
nchevobbe: We could use `renderRep` here

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+    object: stub,
+    onInspectIconClick,
+  }));
+
+  it("has one node grip", () => {
+    expect(grips.length).toBe(1);
+  });
+
+  it("renders no inspect icon when the node is not connected to the DOM tree", () => {
+    const renderedComponentWithoutInspectIcon = shallow(Rep({
```
nchevobbe: Could we use `renderRep` to be consistent ?

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+      object: stubs.get("FulfilledWithDisconnectedNode"),
+      onInspectIconClick,
+    }));
+    const node = renderedComponentWithoutInspectIcon.find(".open-inspector");
+
+    expect(node.exists()).toBe(false);
+  });
+
+  it("renders an inspect icon", () => {
```
nchevobbe: this block is a bit weird here, I have trouble finding to which `renderedComponent` it applies.
Could you move it into the `it` block where we declare `renderedComponent` ? Or recreate a new component so we have it in the same block.
Thanks !

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+    const icon = renderedComponent.find(".open-inspector");
+
+    expect(icon.exists()).toBe(true);
+  });
+
+  it("gets called with the expected argument when the inspect icon is clicked", () => {
+    const icon = renderedComponent.find(".open-inspector");
```
nchevobbe: Same thing here , I don't really know which `renderedComponent` we're working on.
I think we could merge this with the previous `it` block.

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+    icon.simulate("click");
+
+    expect(inspectIconClickedValues).toBe(grips[0]);
+  });
+});
+
+describe("test object link", () => {
```
nchevobbe: Could we rename this to `Promise - objectLink` ?

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -0,0 +1,208 @@
...
+  const object = stubs.get("Pending");
+  const defaultOutput = `*Promise** { *<state>: "pending"* }*`;
+  const objectLink = {
+    objectLink: (props, ...children) => React.DOM.span({},
+      "*", ...children, "*"),
+  };
+
+  it("should render as expected", () => {
+    expect(renderRep(object, Object.assign({mode: undefined}, objectLink)).text())
```
nchevobbe: I think we can get rid of `Object.assign` here, and do directly : 
```js
renderRep(object, {mode: undefined, objectLink})
```

packages/devtools-reps/src/reps/test-helpers.js
```diff
@@ -0,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
arashm: I tried that, but then Jest fails since it wants to pick it up as a test suite:

```
 FAIL  src/reps/tests/test-helpers.js
  ● Test suite failed to run

    Your test suite must contain at least one test.
      
      at onResult (node_modules/jest-cli/build/TestRunner.js:110:20)
```

Any idea how I can fix this?

nchevobbe: I think you could try to exclude the helper file by appending its path to `testPathIgnorePattern` (see https://facebook.github.io/jest/docs/configuration.html#testpathignorepatterns-array-string) in https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-reps/package.json#L49

arashm:
@nchevobbe I think I've touched everything you mentioned.

PR #436, opened 2017-06-11
Move CommentNode rep mochitest to jest

jakub-c:
Based on #415


arashm:
@nchevobbe I believe your approach is better. I think I will update #435 to follow the same pattern.

sidthesloth92:
@nchevobbe Yours look much succint. Will follow your when I take up new ones.

jakub-c:
@nchevobbe I followed your example and I wrapped all text rendering tests in one `it` block.

jasonLaster:
Thanks @jakub-c! By the way, you should stop by our #reps channel in slack!

https://devtools-html.slack.com

PR #438, opened 2017-06-12
Override console.error in Failure test.

nchevobbe:
The failure test throws an exception on purpose, so we can
catch it and render an ad-hoc component.
However, jest would still output the exception when running the
test, which can be a bit confusing since the test is still passing.
Overriding `console.error` for the time of the test allow us to
get rid of the exception message.

### DIFF
--- a/packages/devtools-reps/src/reps/tests/failure.js
+++ b/packages/devtools-reps/src/reps/tests/failure.js
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* global beforeAll, afterAll */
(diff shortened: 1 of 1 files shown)


packages/devtools-reps/src/reps/tests/failure.js
```diff
@@ -12,6 +13,14 @@ let { Rep } = REPS;
 
 const stubs = require("../stubs/failure");
 
+let originalConsoleError;
+beforeAll(() => {
+  // Let's override the console.error function so we don't get an error message
+  // in the jest output for the expected exception.
+  originalConsoleError = window.console.error.bind(window);
```
codehag: we probably don't need the bind here

nchevobbe:
@codehag CI is happy now :) Do you mind approving the review ? Thanks !

PR #440, opened 2017-06-12
Switch ArrayRep test from mochitest to jest

nchevobbe:


packages/devtools-reps/src/reps/tests/array.js
```diff
@@ -0,0 +1,117 @@
...
+  getRep,
+} = require("../rep");
+
+const { MODE } = require("../constants");
+const { ArrayRep, Rep } = REPS;
+
+const maxLength = {
+  short: 3,
+  long: 10
+};
```
janodvarko: nit: Could we use the same approach as for GripArray? I.e. exporting map of max lengths for individual modes and reusing it?

nchevobbe: sure, let me change that

PR #441, opened 2017-06-12
Update node and jest version

nchevobbe:
Try to fix CI issues we have


nchevobbe:
Circle is still failing, but on flow, and not on jest anymore.
Maybe we can merge this ?

nchevobbe:
So, re-launched 2 times, still doesn't fail on jest :)

PR #443, opened 2017-06-12
Move LongString rep mochitest to jest

jakub-c:
Based on #415


packages/devtools-reps/src/reps/tests/long-string.js
```diff
@@ -0,0 +1,89 @@
...
+
+  it("expected to omit quotes", () => {
+    const stub = stubs.get("testMultiline");
+    const renderedComponent = shallow(LongStringRep.rep({
+      object: stub,
+      cropLimit: 20,
+      useQuotes: false
+    }));
+
+    expect(renderedComponent.text()).toEqual(`a aaaaaaaaaaaaaaaaaa…`);
```
jakub-c: @nchevobbe @jasonLaster - the original assertion `"a\naaaaaaaaaaaaaaaaaa…"` failed, after some research I figured that the function actually expects `a aaaaaaaaaaaaaaaaaa…`. So the question is - what was wrong: the original assertion from `test_reps_long-string.html` (line 99) or the result of the function.

packages/devtools-reps/src/reps/tests/long-string.js
```diff
@@ -0,0 +1,89 @@
...
+  getRep,
+} = require("../rep");
+let { LongStringRep } = REPS;
+const stubs = require("../stubs/long-string");
+
+function quoteNewlines(text) {
+  return text.split("\n").join("\\n");
+}
+
+describe("test Multiline / LongStringRep", () => {
```
nchevobbe: Could this be renamed to `LongStringRep` ?

packages/devtools-reps/src/reps/tests/long-string.js
nchevobbe: Hey @jakub-c , thanks for working on this !
So the original assertion is : 
```js
is(renderedComponent.textContent, "a\naaaaaaaaaaaaaaaaaa…");
```

If you pay attention, the expected text uses double quotes (`"`), but you are using backtick quotes (`), and those handle line breaks differently. So in your case, when you have 
```js
  `a\na`
```
it does not mean there's a line break, but that the string actually contains the `\n` characters.
Switching to double quotes should resolve the issue I think.

nchevobbe: See https://github.com/devtools-html/devtools-core/pull/443#discussion_r121601406

jakub-c: @nchevobbe - I started refactoring with that original assertion `"a\naaaaaaaaaaaaaaaaaa…"` but that results with:
``` FAIL  src/reps/tests/long-string.js
  ● LongStringRep › expected to omit quotes

    expect(received).toEqual(expected)
    
    Expected value to equal:
      "a
    aaaaaaaaaaaaaaaaaa…"
    Received:
      "a aaaaaaaaaaaaaaaaaa…"
      
      at Object.it (src/reps/tests/long-string.js:87:38)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
      at process._tickCallback (internal/process/next_tick.js:109:7)
```
therefore I started thinking if the original assertion was correct. The test result indicates that the function returns one line, but the assertion states that the result should contain a line break. That's why I wasn't sure what's wrong, the function or the assertion? 
Another possibility is that I called it incorrectly - but I hope that's not the case.

nchevobbe: I think there is some issue with `.text()`.
Checking the html does give us the linebreak we expect:
```js
    expect(renderedComponent.html())
      .toEqual('<span class="objectBox objectBox-string">a\naaaaaaaaaaaaaaaaaa…</span>');
```

Let's do it that way for now.

PR #453, opened 2017-06-14
Move ElementNode rep mochitest to jest

arashm:
Based on #415.
Some notes. I have merged some of the tests that shared the same stub. Mostly related to click events.
One of the tests was failing with this error:

```
Expected value to equal:
   "<p id=\"lots-of-attributes\" a=\"\" b=\"\" c=\"\" d=\"\" e=\"\" f=\"\" g=\"\" h=\"\" i=\"\" j=\"\" k=\"\" l=\"\" m=\"\" n=\"\">"
Received:
    "<p id=\"lots-of-attributes\" n=\"\" b=\"\" c=\"\" d=\"\" e=\"\" f=\"\" a=\"\" h=\"\" i=\"\" j=\"\" k=\"\" l=\"\" m=\"\" g=\"\">"
```
Which is just because the attributes are returning by a different order. I changed the expected value and it's working now. But I'm not sure if it's ok.
Also on [line 121](https://github.com/devtools-html/devtools-core/compare/master...arashm:element-node-to-jest?expand=1#diff-9785ba37bd11cd5ac8694ea2d53a6a57R121) I'm sending an event object to the `simulate` call, since enzyme seems to not support sending actual event to a shallow rendered node.


nchevobbe:
I don't think the difference we see is normal.
In the ElementNodeRep, we order the properties to ensure we show `id` and `class` first, and then the other attributes, in the order they were given to us in the packet.
To return the original order, we use `return 0` in the sort function.

It's interesting because in this case Firefox and Chrome behaves differently
Given this test case: 
```js
['id', 'a', 'b', 'c', 'd', 'e', "f", "g", "h", "i", "k"].sort((a1, a2) => {
   return 0
})
```

Firefox will return the array in the same order, when chrome will gives us `["e", "id", "b", "c", "d", "a", "f", "g", "h", "i", "k"]`

What's even more interesting is that if you have 10 items or less in the array, chrome behaviour is not the same.
```js
['id', 'a', 'b', 'c', 'd', 'e', "f", "g", "h", "i"].sort((a1, a2) => {
   return 0
})
```
will return the original order.

MDN says 
> compareFunction(a, b) returns 0, leave a and b unchanged with respect to each other, but sorted with respect to all different elements. __Note: the ECMAscript standard does not guarantee this behaviour, and thus not all browsers respect this.__
which is kind of sad.

So for this test, I suggest that we keep the assertion as it was, and fix the way we order attributes in ElementNode rep.

For now we have : 
```js
  let attributeElements = Object.keys(attributes)
    .sort(function getIdAndClassFirst(a1, a2) {
      if ([a1, a2].includes("id")) {
        return 3 * (a1 === "id" ? -1 : 1);
      }
      if ([a1, a2].includes("class")) {
        return 2 * (a1 === "class" ? -1 : 1);
      }

      // `id` and `class` excepted,
      // we want to keep the same order that in `attributes`.
      return 0;
    })
```
and maybe we could transform that into:
```js
let attributeKeys = Object.keys(attributes);
if(attributeKeys.includes("class")) {
  attributeKeys.splice(attributeKeys.indexOf("class"), 1);
  attributeKeys.unshift("class")
}
if(attributeKeys.includes("id")) {
  attributeKeys.splice(attributeKeys.indexOf("id"), 1);
  attributeKeys.unshift("id")
}
attributeKeys.reduce(....
```
This is not really elegant, but I find it simple and efficient.
What do you think of this @arashm ? Maybe you have a better idea ?

packages/devtools-reps/src/reps/tests/element-node.js
```diff
@@ -0,0 +1,320 @@
...
+    const renderedComponent = shallow(ElementNode.rep({
+      object: stub,
+      mode: MODE.TINY
+    }));
+
+    expect(renderedComponent.text())
+      .toEqual("input#newtab-customize-button.bar.baz");
+  });
+
+  it("renders and inspect icon", () => {
```
nchevobbe: nit:
```diff
-renders and inspect icon
+render an inspect  icon
```

packages/devtools-reps/src/reps/tests/element-node.js
```diff
@@ -0,0 +1,320 @@
...
+    const onInspectIconClick = jest.fn();
+    const renderedComponent = shallow(ElementNode.rep({
+      object: stubs.get("Node"),
+      onInspectIconClick
+    }));
+
+    expect(renderedComponent.find(".open-inspector").exists()).toBeTruthy();
+  });
+
+  it("renders and inspect icon", () => {
```
nchevobbe: I think we can merge this `it` block into the previous one and do the assertion for the click

packages/devtools-reps/src/reps/tests/element-node.js
```diff
@@ -0,0 +1,320 @@
...
+
+    const node = renderedComponent.find(".open-inspector");
+    node.simulate("click", { type: "click" });
+
+    expect(onInspectIconClick.mock.calls.length).toEqual(1);
+    expect(onInspectIconClick.mock.calls[0][0]).toEqual(stub);
+    expect(onInspectIconClick.mock.calls[0][1].type).toEqual("click");
+  });
+
+  it("calls the expected argument when mouseout is fired on Rep", () => {
```
nchevobbe: nit:
```diff
-calls the expected argument 
+calls the expected function 
```

packages/devtools-reps/src/reps/tests/element-node.js
```diff
@@ -0,0 +1,320 @@
...
+      object: stub,
+      onDOMNodeMouseOut
+    }));
+
+    renderedComponent.simulate("mouseout");
+
+    expect(onDOMNodeMouseOut.mock.calls.length).toEqual(1);
+  });
+
+  it("calls the expected argument when mouseover is fired on Rep", () => {
```
nchevobbe: nit:
```diff
-calls the expected argument 
+calls the expected function 
```

packages/devtools-reps/src/reps/tests/element-node.js
```diff
@@ -0,0 +1,320 @@
...
+    const onDOMNodeMouseOver = jest.fn();
+    const renderedComponent = shallow(ElementNode.rep({
+      object: stub,
+      onDOMNodeMouseOver
+    }));
+
+    renderedComponent.simulate("mouseover");
+
+    expect(onDOMNodeMouseOver.mock.calls.length).toEqual(1);
```
nchevobbe: could we also check that the function is called with the stub as an argument ?

packages/devtools-reps/src/reps/tests/element-node.js
```diff
@@ -0,0 +1,320 @@
...
+});
+
+describe("ElementNode - Node without attributes", () => {
+  const stub = stubs.get("NodeWithoutAttributes");
+
+  it("selects ElementNode Rep", () => {
+    expect(getRep(stub)).toBe(ElementNode.rep);
+  });
+
+  it("renders with expected text content in tiny mode", () => {
```
nchevobbe: could we also add abloc to ensure we display the right thing when we don't provide a mode ?

packages/devtools-reps/src/reps/tests/element-node.js
```diff
@@ -0,0 +1,320 @@
...
+    expect(getRep(stub)).toBe(ElementNode.rep);
+  });
+
+  it("renders with expected text content", () => {
+    const renderedComponent = shallow(ElementNode.rep({
+      object: stub
+    }));
+
+    expect(renderedComponent.text()).toEqual(
+      '<p id="lots-of-attributes" n="" b="" c="" d="" e="" f="" a="" h="" ' +
```
nchevobbe: As discussed, let's revert this as it was in the mochitest, and fix the ordering in ElementNode rep

packages/devtools-reps/src/reps/tests/element-node.js
```diff
@@ -0,0 +1,320 @@
...
+    const renderedComponent = shallow(ElementNode.rep({
+      object: stub,
+      mode: MODE.TINY
+    }));
+
+    expect(renderedComponent.text()).toEqual("svg:circle.svg-element");
+  });
+});
+
+describe("ElementNode - On inspect icon clicked", () => {
```
nchevobbe: nit:
```diff
-ElementNode - On inspect icon clicked
+ElementNode - Disconnected node
```

packages/devtools-reps/src/reps/tests/element-node.js
```diff
@@ -0,0 +1,320 @@
...
+    }));
+
+    expect(renderedComponent.find(".open-inspector").exists()).toBeFalsy();
+  });
+});
+
+describe("ElementNode - Object Link", () => {
+  const stub = stubs.get("BodyNode");
+
+  it("renders with expected text content for body node", () => {
```
nchevobbe: nit:
```diff
-renders with expected text content for body node
+renders with expected text content when passed an objectLink prop

File: packages/devtools-reps/src/reps/tests/element-node.js [hunk 10]
```diff
@@ -0,0 +1,320 @@
...
+      objectLink: (props, ...children) => React.DOM.span({
+        className: "object-link"
+      }, "*", ...children, "*")
+    }));
+
+    expect(renderedComponent.text())
+      .toEqual('*<body id="body-id" class="body-class">*');
+  });
+
+  it("renders with expected text content for body node in tiny mode", () => {
```
nchevobbe: nit:
```diff
-renders with expected text content for body node in tiny mode
+renders with expected text content when passed an objectLink prop in tiny mode

arashm:
@nchevobbe Your approach is better I believe, since it removes a loop. For the sake of easier review, I pushed the changes in a new commit. If it's ok, you can later squash and merge it.

nchevobbe:
@arashm CI is failing on ESLint errors, could you fix them ? (`yarn eslint packages/devtools-reps/src/` to see them)

arashm:
@nchevobbe Oops. Sorry. I actually fixed them before, but forgot to push them.

nchevobbe:
no worries, it happens to me all the time 😀

arashm:
It still fails with `ipaddr.js` issue :/

nchevobbe:
let's rebuild

PR #456, opened 2017-06-16
Move Error rep mochitest to jest

arashm:
Based on #415


packages/devtools-reps/src/reps/tests/error.js
```diff
@@ -0,0 +1,344 @@
...
+
+    expect(renderedComponent.text()).toEqual("EvalError");
+  });
+});
+
+describe("Error - Internal error", () => {
+  // Test object = `new InternalError("InternalError message")`
+  const stub = stubs.get("InternalError");
+
+  it("correctly selects the Error Rep for EvalError object", () => {
```
nchevobbe:
```diff
- EvalError object
+InternalError object

PR #457, opened 2017-06-19
Move Event rep mochitest to jest

nchevobbe:


bgrins:
There looks like a failure on CI but appears unrelated to your change

PR #462, opened 2017-06-20
Support of the WebAssembly source maps and binary sources

yurydelendik:
Continuation of https://bugzilla.mozilla.org/show_bug.cgi?id=1364535 but for debugger.html.


yurydelendik:
(debugger.html part at https://github.com/yurydelendik/debugger.html/tree/wasm-sourcemaps)

PR #463, opened 2017-06-20
Move GripMap rep mochitest to jest

nchevobbe:


bgrins:
Looks like `yarn flow` failed but it seems unrelated

nchevobbe:
Yeah, that was fixed in another PR

PR #466, opened 2017-06-21
Move Grip rep mochitest to jest

nchevobbe:


packages/devtools-reps/src/reps/grip.js
```diff
@@ -249,10 +249,15 @@ function supportsObject(object, type) {
   return (object.preview && object.preview.ownProperties);
 }
 
+const maxLengthMap = new Map();
+maxLengthMap.set(MODE.SHORT, 3);
```
bgrins: You can use a shorthand to initialize:

```
const maxLengthMap = new Map([[MODE.SHORT, 3], [MODE.LONG, 10]]);
```

PR #478, opened 2017-06-30
Add an actor id data attribute on tree-node

nchevobbe:
This then can be used to retrieve the actor the node is representing.
For example, in the console we can use it in the context menu entry
to store an object as a global variable.

### DIFF
--- a/packages/devtools-reps/src/reps/attribute.js
+++ b/packages/devtools-reps/src/reps/attribute.js
@@ -31,7 +31,10 @@ function Attribute(props) {
   let value = object.preview.value;
 
   return (
-    safeObjectLink(props, {className: "objectLink-Attr"},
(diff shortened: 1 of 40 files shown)


janodvarko:
Ah, I forgot one thing, what about naming the attribute "data-actor-id" (just to make it shorter/simpler)

Honza

nchevobbe:
I agree that it would make it simpler. I think this could be a follow-up, because if we do that we'll have to modifiy the "store as global variable" context menu entry code in mozilla-central too.
What do you think ?

janodvarko:
> I agree that it would make it simpler. I think this could be a follow-up, because if we do that we'll have to modifiy the "store as global variable" context menu entry code in mozilla-central too.
What do you think ?

I see, ok for me.

Honza

PR #483, opened 2017-07-05
Add a prop to force object to be rendered using non-remote reps

### LINKED ISSUE #480: Do not use `type` in case of direct object access
Reps support rendering of remote objects. In this case, values passed as `object` into Reps is always a Grip object. But, reps also support direct rendering of JS objects. In this case Reps have direct access to the JS object defined on the page.

The later scenario is used by JSONViewer. It runs directly within the page and so, has direct access to JS objects defined on the page. No grips involved.

There is one catch though, `getRep` is using `object.type` to get type of the object assuming the the `object` is always a grip. That's wrong. This means that JS objects (defined on the page) using `type` field can accidentally mangle the logic by providing random data.

Also `isGrip` method accessing `object.actor` can  cause similar troubles.

Honza

nchevobbe:
Fixes #480

Modify the `getRep` and `supportsObject` functions so we can pass this `noGrip`
property to it.
Add tests in the `Obj` rep which try to render stubs representing different
remote object, with the `noGrip` prop, and makes sure it uses the `Obj` rep,
and not the "remote" one.

### DIFF
--- a/packages/devtools-reps/src/reps/grip.js
+++ b/packages/devtools-reps/src/reps/grip.js
@@ -31,6 +31,7 @@ GripRep.propTypes = {
   onDOMNodeMouseOver: React.PropTypes.func,
   onDOMNodeMouseOut: React.PropTypes.func,
   onInspectIconClick: React.PropTypes.func,
+  noGrip: React.PropTypes.bool,
 };
 
 function GripRep(props) {
@@ -242,8 +243,8 @@ function getPropValue(property) {
 }
 
 // Registration
-function supportsObject(object, type) {
-  if (!isGrip(object)) {
+function supportsObject(object, type, noGrip = false) {
+  if (noGrip === true || !isGrip(object)) {
     return false;
   }
   return (object.preview && object.preview.ownProperties);
--- a/packages/devtools-reps/src/reps/attribute.js
+++ b/packages/devtools-reps/src/reps/attribute.js
@@ -48,8 +48,8 @@ function getTitle(grip) {
 }
 
 // Registration
-function supportsObject(grip, type) {
-  if (!isGrip(grip)) {
+function supportsObject(grip, type, noGrip = false) {
+  if (noGrip === true || !isGrip(grip)) {
     return false;
   }
 
--- a/packages/devtools-reps/src/reps/comment-node.js
+++ b/packages/devtools-reps/src/reps/comment-node.js
@@ -42,8 +42,8 @@ function CommentNode(props) {
 }
 
 // Registration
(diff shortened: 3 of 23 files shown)


packages/devtools-reps/src/reps/grip.js
```diff
@@ -31,6 +31,7 @@ GripRep.propTypes = {
   onDOMNodeMouseOver: React.PropTypes.func,
   onDOMNodeMouseOut: React.PropTypes.func,
   onInspectIconClick: React.PropTypes.func,
+  noGrip: React.PropTypes.bool,
```
janodvarko: Shouldn't the new prop be defined on `Rep` component?

nchevobbe: The [Rep component](https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-reps/src/reps/rep.js#L78) isn't really one, it's just a function that calls the correct rep, and thus has no propType.

We might type it with flow later, in which case we'll take care of adding this property into what can be in the `props` object.

Loirooriol:
I think you forgot to check `noGrip` in https://github.com/devtools-html/devtools-core/blob/bede79a8f0e9d1d6bd8bb9655346578890f3d1ec/packages/devtools-reps/src/reps/rep.js#L110

Anyways, `isGrip` function is completely unreliable. I would remove the function, rename this `noGrip` parameter to `isGrip`, and use that instead.

nchevobbe:
Hello @Loirooriol , thanks for spotting that.
I will check `noGrip` in `getRep` function.

For your suggestion of removing `isGrip`, I wonder what do you see as a replacement ? Only rely on a `isGrip` (`useGrip` maybe makes more sense here) to pick Grip reps over an other ?

I think this might be a good idea, and I can't think of examples mixing RDP object and "raw" ones.
This new property might default to `true`, so we only have to make the change in the JSON Viewer.

@janodvarko , what do you think of this ?

janodvarko:
Yep, I like the idea and 'useGrip' or 'useGrips' sounds good to me.

Honza

Loirooriol:
@nchevobbe Yes, since it seems there is no reliable way to check if an object is a grip, I think Reps should only rely on this parameter.

Defaulting to `true` seems a good idea, and in that case I would check that the supposed grip really has an `actor` property. Otherwise throw an error informing that `useGrip = false` should be used instead. That is, fail fast and loud to avoid bad behaviors in edge cases like https://bugzilla.mozilla.org/show_bug.cgi?id=1377668

BTW, not only the JSON Viewer, the netmonitor also needs this.

PR #484, opened 2017-07-06
Add source-utils from m-c into devtools-core

nchevobbe:
This work is done in order to be able to import the Stacktrace (and thus Frame) component, which use those utils functions, in devtools-component from mozilla-central.
The end goal is to use the Stacktrace component in Error Rep so we can have full sourcemap support (instead of the error.stack string we use now).

The tests were also copied from mozilla-central and migrated to jest.

### DIFF
--- a/packages/devtools-reps/src/index.js
+++ b/packages/devtools-reps/src/index.js
@@ -5,6 +5,7 @@
 const { MODE } = require("./reps/constants");
 const { REPS, getRep } = require("./reps/rep");
 const ObjectInspector = require("./object-inspector/");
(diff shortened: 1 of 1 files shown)


packages/devtools-utils/src/source-utils.js
```diff
@@ -0,0 +1,365 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jasonLaster: this is a minor nit, but could you drop this module in `devtools-modules`? The rationale, is that these are the modules that are copied over and thus should be kept in sync

PR #485, opened 2017-07-07
Tweak styling in order to integrate it in the console

nchevobbe:
This patch removes CSS references to `objectLink` and adds an `objectTitle` class on all the Reps that have a title.
This also allow the tree items to be wrapped, unless a `noWrap` prop is passed to the object inspector.

### DIFF
--- a/packages/devtools-reps/bin/copy-assets.js
+++ b/packages/devtools-reps/bin/copy-assets.js
@@ -32,13 +32,6 @@ function start() {
 
   console.log("  output path is:", mcPath);
 
-  console.log("  copying reps.css...");
-  copyFile(
-    path.resolve(projectPath, "src/reps/reps.css"),
-    path.join(mcPath, mcModulePath, "reps.css"),
-    {cwd: projectPath}
-  );
-
   console.log("  creating reps.js bundle...");
   makeBundle({
     outputPath: path.join(mcPath, mcModulePath),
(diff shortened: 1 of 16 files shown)


packages/devtools-reps/bin/copy-assets.js
```diff
@@ -32,13 +32,6 @@ function start() {
 
   console.log("  output path is:", mcPath);
 
-  console.log("  copying reps.css...");
-  copyFile(
-    path.resolve(projectPath, "src/reps/reps.css"),
-    path.join(mcPath, mcModulePath, "reps.css"),
-    {cwd: projectPath}
-  );
-
```
janodvarko: Why is this removed?

nchevobbe: The makeBundle command produce a reps.js and also a {bundlename}.css file (which in our case will be reps.css, but not the same that is in `/src/reps/`).
This CSS bundle contains all the CSS files that are required  (using `require`) in the package.
In our case, we require `src/reps/reps.css` from `src/reps/rep.js`  and `src/object-inspector/index.css` from `src/object-inspector/index.js`.
This will produce a CSS bundle which is the concatenation of `src/reps/reps.css` and `src/object-inspector/index.css`.

Hence the removal of this snippet since it only targeted the `src/reps/reps.css` file.

PR #491, opened 2017-07-11
Always show quotes when rendering an empty string.

### LINKED ISSUE #490: Find a better way to represent empty strings
__From https://bugzilla.mozilla.org/show_bug.cgi?id=1375752__

---

If you try to render an empty string Rep (`""`), and have `useQuotes = false`, the Rep will render `<span class="objectBox objectBox-string"></span>`, i.e. an empty element.

This might be weird in some cases (e.g. `console.log` which pass `useQuotes = false` to render its arguments).
~It shouldn't hurt to always show the quotes if the string is empty, no matter what the value of `useQuotes` is, but maybe I'm missing something ?~

Let's think about better way to display empty strings.

nchevobbe:
Fixes #490.

### DIFF
--- a/packages/devtools-reps/src/reps/string.js
+++ b/packages/devtools-reps/src/reps/string.js
@@ -42,7 +42,7 @@ function StringRep(props) {
     config.style = style;
   }
 
-  if (useQuotes) {
+  if (useQuotes || text === "") {
     text = escapeString(text, escapeWhitespace);
   } else {
     text = sanitizeString(text);
--- a/packages/devtools-reps/src/reps/tests/string.js
+++ b/packages/devtools-reps/src/reps/tests/string.js
@@ -70,6 +70,13 @@ const testCases = [{
     escapeWhitespace: false,
   },
   result: "\"line 1\r\nline 2\n\tline 3\""
+}, {
+  name: "empty string",
+  props: {
+    object: "",
+    useQuotes: false,
+  },
+  result: `""`
 }];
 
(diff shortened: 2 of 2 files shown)


nchevobbe:
@bgrins any thoughts ?

bgrins:
Copying in some of our discussion from yesterday. The [spec seems to give room for this sort of thing](https://console.spec.whatwg.org/#printer)
> How the implementation prints args is up to the implementation, but implementations should separate the objects by a space or something similar, as that has become a developer expectation.

This would make `console.log(" ", "");` show up as ` ""`, so I think I’d prefer either:
1) Quote all string output as we do with evaluation results
2) Quote all whitespace-only strings (not just empty)

I lean towards 1 since it's most consistent (and also matches how we treat evaluation results).  Although it would be a big change in the output format.  Is there a particular reason why console.log(“foo”, “bar”) should be preferred as `foo bar` instead of `"foo" "bar"`? One argument for keeping them without quotes is that's what other consoles (including node) seem to do.

tromey:
> One argument for keeping them without quotes is that's what other consoles (including node) seem to do.

I think it's also useful to be able to display text without quoting.

nchevobbe:
what about having a pref or something in the UI that let you switch from one to another ? This way people could make it fit their needs, and we could keep doing what we do now.
The hard thing would be to make this discoverable so people can overcome the issues of empty/spaces only strings.

PR #505, opened 2017-07-17
WIP: Upgrade to webpack 3

zaggy:
I upgraded webpack to version 3 for launchpad. Also I bumped flow to version 0.50, updated lerna to a stable version, made sure lerna uses yarn client now.

This requires extensive testing, as while it is running when I do `yarn start`, there might be several regressions. But I suppose that full testing might be done only after we bump launchpad version and check it in debugger.html?

There might be issues with Chrome, as one portion of the webpack config which was used as a hack to fix Chrome performance is not working anymore.


zaggy:
@TheLarkInn I do absolutely agree with that. I think current configs are a bit complicated and would love it to be more functional in nature.

jasonLaster:
@zaggy I created a webpack-merge issue https://github.com/devtools-html/devtools-core/issues/542 :)

Would love to see us organize + test our configs

PR #508, opened 2017-07-17
Modify how we indicate there is more to see in an object.

nchevobbe:
Basically, for the most case we switched from `more…` to `…` only.
We also re-use the ellipsis in TINY mode for arrays and objects.
Now empty arrays and objects in TINY mode should render respectively
as `[]` and `{}`, and non-empty arrays and objects `[…]` and `{…}`.

This should indicate more clearly and in a more consistent way that
there are hidden properties/items in what we render, and encourage the
user to expand the object/array to check what it really contains.

Doing this work was also a great opportunity to be more consistent through
all the reps, and mostly have the same output for RDP and "plain" objects
for the same test cases.

### DIFF
--- a/packages/devtools-reps/src/reps/grip-array.js
+++ b/packages/devtools-reps/src/reps/grip-array.js
@@ -8,7 +8,6 @@ const {
   isGrip,
   wrapRender,
 } = require("./rep-utils");
-const Caption = require("./caption");
 const { MODE } = require("./constants");
 
 // Shortcuts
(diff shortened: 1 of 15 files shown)


packages/devtools-reps/src/reps/tests/array.js
```diff
@@ -90,9 +89,9 @@ describe("Array", () => {
     }];
     const renderRep = props => shallow(Rep(Object.assign({ object }, props)));
 
-    const defaultOutput = "[ Object ]";
+    const defaultOutput = "[ {…} ]";
```
jasonLaster: looks good

packages/devtools-reps/src/reps/grip-array.js
```diff
@@ -122,9 +126,9 @@ function arrayIterator(props, grip, max) {
   }
 
   const previewItems = getPreviewItems(grip);
-  if (!previewItems) {
-    return items;
-  }
+  // if (!previewItems) {
+  //   return items;
+  // }
```
janodvarko: Shouldn't we just remove this code?

nchevobbe: Good catch !

PR #513, opened 2017-07-24
Display symbols properties on objects.

nchevobbe:
This patch adds the support for the `ownSymbols` property
introduced in [Bug 881480](https://bugzilla.mozilla.org/show_bug.cgi?id=881480),
as well as some tests (and stubs) to ensure they are rendered as expected.


nchevobbe:
you're right, the symbols were only displayed in Reps preview

PR #517, opened 2017-07-25
Handle safeGetterValue in the ObjectInspector.

nchevobbe:
In getObjectProperties, we may receive a safeGetterValues property in addition to
the ownProperty one. This PR handle the data from this property so we can show
these values as children in the ObjectInspector.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/object-inspector.js
+++ b/packages/devtools-reps/src/object-inspector/tests/object-inspector.js
@@ -12,11 +12,13 @@ const { MODE } = require("../../reps/constants");
 const { Rep } = require("../../reps/rep");
 const gripStubs = require("../../reps/stubs/grip");
 const accessorStubs = require("../../reps/stubs/accessor");
+const performanceStubs = require("../stubs/performance");
 
 const {
   createNode,
   getChildren,
   getPromiseProperties,
+  getValue,
   isDefault,
   isPromise,
   makeNodesForProperties,
@@ -498,7 +500,7 @@ describe("ObjectInspector", () => {
     expect(oi.find(".node").at(1).contains(renderRep(MODE.TINY))).toBeTruthy();
   });
 
-  it.only("renders getters as expected", () => {
+  it("renders getters as expected", () => {
     const stub = accessorStubs.get("getter");
     const oi = mount(ObjectInspector({
       autoExpandDepth: 1,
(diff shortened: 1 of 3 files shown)


nchevobbe:
@jasonLaster we'd like to have that for 56, and merge day is soon. You're my only hope as Honza is in vacation and Brian unavailable for today :D

packages/devtools-reps/src/object-inspector/tests/object-inspector.js
```diff
@@ -741,4 +743,76 @@ describe("getChildren", () => {
...
+      ["navigationStart", 1500967716401],
+      ["redirectEnd", 0],
+      ["redirectStart", 0],
+      ["requestStart", 1500967716401],
+      ["responseEnd", 1500967716401],
+      ["responseStart", 1500967716401],
+      ["secureConnectionStart", 1500967716401],
+      ["unloadEventEnd", 0],
+      ["unloadEventStart", 0],
+      ["__proto__", stub.prototype]
```
jasonLaster: this is a good example where the snapshot approach could be easier to maintain. The next time you generate the stubs, you'll get different data, but it is not relevant for the test

packages/devtools-reps/src/object-inspector/utils.js
```diff
@@ -209,13 +210,22 @@ function makeNodesForOwnProps(properties, parentPath, ownProperties) {
...
+  } = objProps;
+
   const parentPath = parent.path;
   const parentValue = getValue(parent);
-  const properties = sortProperties(Object.keys(ownProperties)).filter(name =>
-    ownProperties[name].hasOwnProperty("value")
-    || ownProperties[name].hasOwnProperty("get")
-    || ownProperties[name].hasOwnProperty("set")
+
+  let allProperties = Object.assign({}, ownProperties, safeGetterValues);
```
jasonLaster: 👍

packages/devtools-reps/src/object-inspector/tests/object-inspector.js
nchevobbe: Yeah, I should learn more about snapshots.

The thing I'm afraid of is that it doesn't output in the tests how your component should be, only that it need to be like something you defined in the past. This makes me worry about regressions introduced to the output being unnoticed because the person who wrote the patch only updated the snapshot after they changed how the components is rendered.

PR #520, opened 2017-07-25
Handle exceptions in network request and worker utilities

tromey:
This changes the network request code and the worker handler code to
catch exceptions and report them to the caller via their promise
results.  It also changes the network request code to provide an error
message when the XHR's statusText is empty; this can happen when an
invalid URL is used.

### DIFF
--- a/packages/devtools-utils/src/tests/worker-utils.js
+++ b/packages/devtools-utils/src/tests/worker-utils.js
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const { WorkerDispatcher } = require("../worker-utils")
+const { WorkerDispatcher, workerHandler } = require("../worker-utils")
 
 describe("worker utils", () => {
   it("starts a worker", () => {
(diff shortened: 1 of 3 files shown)


packages/devtools-utils/src/worker-utils.js
```diff
@@ -57,17 +57,21 @@ WorkerDispatcher.prototype = {
...
+      const response = publicInterface[method].apply(undefined, args);
+      if (response instanceof Promise) {
+        response.then(val => self.postMessage({ id, response: val }),
+                      err => self.postMessage({ id, error: err }));
+      } else {
+        self.postMessage({ id, response });
+      }
+    } catch (error) {
+      self.postMessage({ id, error });
     }
```
jasonLaster: it might be nice to add unit tests where there are exceptions that are caught here.

jasonLaster: it might be nice to add unit tests where there are exceptions that are caught here.

packages/devtools-utils/src/tests/worker-utils.js
```diff
@@ -56,6 +56,22 @@ describe("worker utils", () => {
     });
 
     expect(addEventListenerMock.mock.calls.length).toEqual(1);
-  })
+  });
+
+  it("test workerHandler error case", () => {
```
jasonLaster: Looks good!

PR #522, opened 2017-07-26
Fix getValue function for legitimate falsy values.

nchevobbe:
The `getValue` was failing if the item had a falsy value (`false`, `null`, `0`, …).
This patch fixes this issue and adds some tests to ensure `getValue` works as expected.


packages/devtools-reps/src/object-inspector/tests/object-inspector.js
```diff
@@ -816,3 +816,89 @@ describe("getChildren", () => {
...
+
+    item = {
+      contents: {
+        get: "get",
+        set: "set"
+      }
+    };
+    expect(getValue(item)).toEqual({get: "get", set: "set"});
+  });
+});
```
jasonLaster: You might want to start breaking up the object-inspector tests file along describe sections.

packages/devtools-reps/src/object-inspector/utils.js
```diff
@@ -12,8 +12,11 @@ if (typeof window === "object") {
 }
 
 function getValue(item) {
-  let value = get(item, "contents.value", undefined)
-    || get(item, "contents.getterValue", undefined);
+  let value = get(
+    item,
+    "contents.value",
+    get(item, "contents.getterValue", undefined)
+  );
```
jasonLaster: using the default value works, but it is a little unintuitive.

Perhaps 

```js
const value = get(item, "contents.value")
const getterValue = get(item, "contents.value")

return `insert some better logic here`
```

packages/devtools-reps/src/object-inspector/utils.js
```diff
@@ -12,10 +13,13 @@ if (typeof window === "object") {
...
+  let value;
 
-  if (!value && nodeHasAccessors(item)) {
+  if (has(item, "contents.value")) {
+    value = get(item, "contents.value");
+  } else if (has(item, "contents.getterValue")) {
+    value = get(item, "contents.getterValue", undefined);
+  } else if (nodeHasAccessors(item)) {
     value = item.contents;
   }
```
jasonLaster: I like this pattern. I personally find early returns nicer than conditional assignments


```js
if (x) {
  return y.x
}
  
if (y) {
  return y.y
}
```

packages/devtools-reps/src/object-inspector/tests/utils/make-node-for-properties.js
```diff
@@ -0,0 +1,258 @@
...
+
+  it("includes getters and setters", () => {
+    const nodes = makeNodesForProperties(
+      {
+        ownProperties: {
+          foo: { value: "foo" },
+          bar: {
+            "configurable": true,
+            "enumerable": true,
+            "get": {
```
jasonLaster: this is minor, but do we need all these fields for the test to be valid?

I like small fixture data because it's easier to see what the data dependencies are.

packages/devtools-reps/src/object-inspector/tests/utils/promises.js
```diff
@@ -0,0 +1,82 @@
...
+describe("promises utils function", () => {
+  it("is promise", () => {
+    const promise = {
+      contents: {
+        enumerable: true,
+        configurable: false,
+        value: {
+          frozen: false,
+          ownPropertyLength: 0,
+          preview: {
```
jasonLaster: do we need preview here? and other fields that are not related to promises...

packages/devtools-reps/src/object-inspector/utils.js
nchevobbe: right, I'll make the change

packages/devtools-reps/src/object-inspector/tests/utils/make-node-for-properties.js
nchevobbe: I guess there are not needed, we just want to make sure get/set are picked up

packages/devtools-reps/src/object-inspector/tests/utils/promises.js
nchevobbe: removed

Issue #23, opened 2016-12-21
ReactDOM 15.3 does not work with single module plugin


jasonLaster:
react-dom has an index.js file with just a module.exports = require('react/lib/reactDOM), which our single-module-plugin turns into module.exports = require() and fails to load react dom.

juliandescottes:
Does this mean SMP confuses the index.js from react-dom with another similar file ?
I see two options here: 
- ignore small files (kind of random but should work here)
- improve the plugin to run in several steps:
  - a: mark all the purely duplicated files (without touching require statements)
  - b: if a require() points to a duplicate marked in step a, normalize it to always point to the same ID
  - c: repeat until no more duplicates are marked

jasonLaster:
```js
'use strict';

module.exports = require('react/lib/ReactDOM');
```

clarkbw:
With SMP gone, is this still an issue?

juliandescottes:
No, moreover we're already using react & react-dom 15.3 :)

PR #525, opened 2017-07-26
[WIP] Trying out react@next

clarkbw:
Trying out [React 16 Beta](https://github.com/facebook/react/issues/10294)

It works, but I don't think this is merge ready yet.

Mostly we've fallen behind the react train.  I've updated PropTypes and CreateClass to their shims but if this moved to ES6 Modules and JSX that would help quite a bit further.


packages/devtools-launchpad/.babelrc
```diff
@@ -10,7 +10,7 @@
 
     ["module-resolver", {
       "alias": {
-        "react-dom": "react-dom/dist/react-dom",
+        "react-dom": "react-dom/umd/react-dom.development",
```
gaearon: Curious: what is the necessity for these aliases?

bvaughn: Looks like it's forcing UMD instead of default (CJS) target?

gaearon: Right. I’m curious why. In the past people used to do this to speed up DEV rebuilds (since React had too many modules). But now CJS is also a single bundle so maybe it‘s unnecessary.

bvaughn: Yeah, sorry. I wasn't really trying to answer your question so much as confirming that I understood what it was doing.

Your assertion seems legit.

clarkbw: @jasonLaster is the right person to answer that question, my understanding is that this has to do with how we package react within the browser toolbox.

jasonLaster: Here was the warning we saw:
https://github.com/devtools-html/debugger.html/issues/1640

And the _fix_:
https://github.com/devtools-html/debugger.html/issues/1731

We've had several other issues creep in over time
https://github.com/devtools-html/devtools-core/issues/179
https://github.com/devtools-html/devtools-core/issues/23

gaearon: So the issue is due to a custom module system (per #23)? I'm not sure what “single module plugin” is.

I guess if it works around a bug for you, that’s cool, I just wanted to verify. Please not `react.development` is *not* suitable for production.

clarkbw:
Closing this out for now, its a good start but will take some more work to make it happen. 🙇

PR #527, opened 2017-07-27
adds streaming worker

jbhoosreddy:
This initial PR adds the streaming worker.

I'm still working on some kinks here. But this is relatively how it should look.

Issue: On the browser, it works better but in the jest test environment using something like
````js
function delay(milliseconds) {
  return function(result) {
    return new Promise(function(resolve, reject) {
      setTimeout(function() {
        resolve(result);
      }, milliseconds);
    });
  };
}
````
would never resolve for `await` ing callers. Which prevents the tests from passing. Returning promises which resolve without setTimeout still work in the tests.
@codehag @jasonLaster @bomsy 

I want to tidy this up a bit. But suggestions welcome here. 😉

### DIFF
--- a/packages/devtools-utils/src/tests/worker-utils.js
+++ b/packages/devtools-utils/src/tests/worker-utils.js
@@ -2,8 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const { WorkerDispatcher, workerHandler, streamingWorkerHandler } = require("../worker-utils")
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 10000;
+const { WorkerDispatcher, workerHandler, streamingWorkerHandler } = require("../worker-utils");
 
 describe("worker utils", () => {
   it("starts a worker", () => {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/devtools-utils/src/worker-utils.js [hunk 1]
```diff
@@ -76,7 +78,51 @@ function workerHandler(publicInterface: Object) {
   };
 }
 
+function streamingWorkerHandler(publicInterface: Object, { timeout = 100 } = {}, worker) {
```
jasonLaster: lets move this to its own module

codehag:
hmm .. if there are issues with testing setTimeout my first gut instinct is to use jasmine.clock...

https://jasmine.github.io/2.0/introduction.html#section-Mocking_the_JavaScript_Timeout_Functions
i will take a look also, but this might be something to play with to see if it works!

jbhoosreddy:
Unfortunately, jasmine.clock didn't work out. But jest.useRealTimers did. Also, jasmine.clock is no longer available in jest.

Either way, this works very well, and is good to go!

First round review! @codehag @jasonLaster @bomsy

packages/devtools-utils/src/tests/worker-utils.js
```diff
@@ -74,4 +72,55 @@ describe("worker utils", () => {
...
+  expect(postMessageMock.mock.calls[2][0]).toEqual({
+    id,
+    status: "pending",
+    data: [1]
+  });
+  expect(postMessageMock.mock.calls[3][0]).toEqual({
+    id,
+    status: "done"
+  });
+});
```
jasonLaster: Haha - can you explain what this is testing?

I expected to see call reaults stream in over time.

jbhoosreddy: So in this example, the test sets the worker with 25ms timeout. But each task takes 50ms. So in this example the worker should be able to only send each task result separately.
So the expected result, is the worker sending a `start`, result [2], result [1], and `done`.

It does stream in the results as expected. I added a `done` message because I thought there maybe a usecase for that. @jasonLaster

packages/devtools-utils/src/tests/worker-utils.js
```diff
@@ -74,4 +72,55 @@ describe("worker utils", () => {
...
+  });
+  expect(postMessageMock.mock.calls[1][0]).toEqual({
+    id,
+    status: "pending",
+    data: [2]
+  });
+  expect(postMessageMock.mock.calls[2][0]).toEqual({
+    id,
+    status: "pending",
+    data: [1]
```
jasonLaster: why does the first pending call have two results and the second have one?

packages/devtools-utils/src/tests/worker-utils.js
jasonLaster: Oh wow, that makes a lot of sense! Thanks

packages/devtools-utils/src/tests/worker-utils.js
jbhoosreddy: oh yeah.. that confused me also. I realized it was because I did `Array.prototype.pop` to remove tasks from the list of tasks. So the second one came out first. I could change it in the test to make more sense? But I think its a minor thing.

jbhoosreddy:
@jasonLaster fixed behavior as per your recommendation. @codehag @bomsy

PR #529, opened 2017-07-27
Display map and set entries in the ObjectInspector

nchevobbe:
Add a way to load and display entries.

This adds a loadObjectEntries as well as a getObjectEntries function as props, which will be used to fetch the entries of an object.
The loadObjectEntries is called only when a custom <entries> node is expanded by the user (if we don't already have all the entries in the preview object).

This adds multiple tests to assert the util functions do what is expected, as well as a more complete case on the whole ObjectInspector component.

Since we needed Maps and Sets for those tests, this patch also adds some Sets and WeakSets stubs along with some basic tests.

This provides the getObjectEntries and loadObjectEntries to the ObjectInspector used in the Result component.
A new "maps & sets" quick link is added as well to be able to quickly see maps and states of different size.
We take this as a opportunity to add redux logger to it so we can see what actions are fired when.


packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -196,12 +209,29 @@ class ObjectInspector extends Component {
...
         loadObjectProperties(value);
+        return;
+      }
+
+      if (
+        nodeIsEntries(item)
+        && !hasAllEntriesInPreview(parent)
+        && !getObjectEntries(parentActor)
+      ) {
+        loadObjectEntries(parentValue);
```
jasonLaster: why differentiate between entries and properties?

also, do we want to return after loading?

packages/devtools-reps/src/object-inspector/tests/utils/__snapshots__/promises.js.snap
```diff
@@ -16,6 +32,22 @@ Array [
       },
     },
     "name": "<reason>",
+    "parent": Object {
```
jasonLaster: glad we have a parent :)

packages/devtools-reps/src/object-inspector/utils.js
```diff
@@ -67,40 +78,120 @@ function nodeHasProperties(item) {
...
+  if (!nodeValue || !nodeValue.preview) {
+    return [];
+  }
+
+  const {key, value} = nodeValue.preview;
+  const path = item.path;
+
+  return [
+    createNode(item, "<key>", `${path}/##key`, {value: key}),
+    createNode(item, "<value>", `${path}/##value`, {value}),
```
jasonLaster: this is one of our least favorite hacks :)

packages/devtools-reps/src/object-inspector/utils.js
```diff
@@ -283,21 +386,30 @@ function createNode(name, path, contents) {
...
+  return { parent, name, path, contents };
 }
 
-function getChildren({ getObjectProperties, actors, item }) {
+function getChildren({
+  getObjectEntries,
+  getObjectProperties,
+  actors,
+  item
+}) {
```
jasonLaster: this is a minor point, but it would be really nice to add flow types to these utils

packages/devtools-reps/src/object-inspector/index.js
nchevobbe: I did it that way because it makes it clear what the consumer needs to load.
If we expect to load entries from `loadObjectProperties`, that means the consumer has to check what kind of data it needs to load.
Now that I think about it, we might be able to do this with a couple of changes.

We should change `getObjectProperties(actor:string)` to `getObjectProperties(path:string)`, which means the consumer needs to keep track of the paths it has already loaded.

In order to do that, we need to change `loadObjectProperties(value:Grip)` to `loadObjectProperties(value:Grip, path:string, nodeTypes: Array<nodeType>)`.
`nodeTypes` would be the inversed tree of the nodeTypes for the path we want to load properties.It would help us when/if we decide to lazy load bucketed items (e.g. _"Let's load the 100-199 slice of the `<entries>` node"_ could be done because `nodeTypes` would be something like `[bucket, entries, grip]`).
The `nodeType` would be pulled of a `type` property on the node (which can be one of bucket, entries, get, set, promiseState, promiseValue, promiseReason. …), which will be nice to have too since it will allow us to rely on it instead of the path to check if a node is a bucket, or an `entries` node for example.
We should then expose those types so the consumer can test against them in order to fetch the correct data.

What do you think @jasonLaster ?

bgrins: Can we have both entries and properties?  i.e. in a scenario like:

```
var f = new Map([[1, 2]]);
f.foo = 'foo';
f;
```

nchevobbe: We can, but we won't load properties and entries at the same time.
When we expand a map, we load its properties (like any objects/arrays). So the `foo` property will show up there.
This gives us : 
```
▼ Map {1 → 2} 
     foo: "foo"
     size: 1
     ▶︎ <entries>
```

Then we load the entries if the user click on the `<entries>` node.

nchevobbe:
This is how a version without a separate function to load entries could look like : https://github.com/devtools-html/devtools-core/commit/d192916a98e3ae74f5dca41a1dde6b38ae384d79 (not fully working, but you can get the idea).

How does it feel @jasonLaster @bgrins ?

nchevobbe:
And here is a first version of it working in the console https://reviewboard.mozilla.org/r/163020/diff/3#index_header (I went with the separate `loadEntries` since it was the easiest, but I don't think switching to the other approach would require too much time if we finally decide to do it).

bgrins:
I'm fine with separate functions for loading entries and properties. Either way the consumer is going to need to provide a way to load both through the client

nchevobbe:
@jasonLaster @bgrins , I think that this is ready to review.
As discussed with Brian, we chose to go with the separate loadEntries callback (the consumer of the OI don't have to do the "branching logic" by itself and this roughly map what we already have in ObjectClient. We can quickly switch to a unified loadProperties callback in the future if we need it).

packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -51,7 +59,9 @@ type Props = {
   mode: Mode,
   roots: Array<Node>,
   disableWrap: boolean,
-  getObjectProperties: (actor:string) => any,
+  getObjectEntries: (actor:string) => ?LoadedEntries,
+  getObjectProperties: (actor:string) => ?LoadedProperties,
+  loadObjectEntries: (value:RdpGrip) => any,
```
jasonLaster: we generally follow this style: `(RdpGrip) => void`

packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -169,13 +183,30 @@ class ObjectInspector extends Component {
...
+      const parentActor = parentValue
+        ? parentValue.actor
+        : null;
+
       if (nodeHasProperties(item) && value && !getObjectProperties(value.actor)) {
         loadObjectProperties(value);
       }
+
+      if (
+        nodeIsEntries(item)
```
jasonLaster: perhaps: `nodeHasEntries`

I could also see moving this to a util: `shouldLoadEntries`

packages/devtools-reps/src/launchpad/index.js
```diff
@@ -35,6 +35,20 @@ function onConnect(connection) {
       const objClient = connection.tabConnection.threadClient.pauseGrip(grip);
       const resp = await objClient.getPrototypeAndProperties();
       return resp;
+    },
+    getEntries: async function (grip) {
+      const objClient = connection.tabConnection.threadClient.pauseGrip(grip);
+      console.log("getEntries", grip);
```
bgrins: Extra logging

PR #532, opened 2017-07-28
Add a MapEntry Rep.

nchevobbe:
When we fetch entries from the server, we get a grip with a "mapEntry" type that we weren't handling previously.
This patch adds a Rep to deal with those kind of grips and harmonize the way we render map entries (even in the map preview), to match what we had previously in the VariableView (e.g. `Map {a → 1}` instead of `Map {a: 1}`, this works better in an OI context where you can have `0: a → 1`, instead of an awkward `0: a: 1`).

~This relies on https://github.com/devtools-html/devtools-core/pull/531 (Some tests are failing without this patch).~  https://github.com/devtools-html/devtools-core/pull/531 Is now merged, CI should be happier.

### DIFF
--- a/packages/devtools-reps/src/reps/tests/grip-map-entry.js
+++ b/packages/devtools-reps/src/reps/tests/grip-map-entry.js
@@ -1,0 +1,143 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ /* global jest */
+const { shallow } = require("enzyme");
+
+const {
+  REPS,
+  getRep,
+} = require("../rep");
+
+const { GripMapEntry } = REPS;
+const { createGripMapEntry } = GripMapEntry;
+const { MODE } = require("../constants");
+
+const stubs = require("../stubs/grip-map-entry");
+const nodeStubs = require("../stubs/element-node");
+const gripArrayStubs = require("../stubs/grip-array");
+
+const renderRep = (object, mode, props) => {
+  return shallow(GripMapEntry.rep(Object.assign({
+      object,
+      mode,
+  }, props)));
+};
+
+describe("GripMapEntry - simple", () => {
+  const stub = stubs.get("A → 0");
+
+  it("Rep correctly selects GripMapEntry Rep", () => {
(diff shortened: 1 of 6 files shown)


jasonLaster:
@nchevobbe could you share some screenshots. Also, have you updated the launchpad examples so it's easy to see them?

nchevobbe:
You can see it here
https://screenshots.firefoxusercontent.com/images/c7ccc9e1-43b8-46f5-9518-5529004f1554.png

Le 28 juil. 2017 7:24 PM, "Jason Laster" <notifications@github.com> a
écrit :

bgrins:
I like the distinction of `→` vs `:` for Maps

packages/devtools-reps/src/reps/tests/grip-map-entry.js
```diff
@@ -0,0 +1,143 @@
...
+    let renderedComponent = renderRep(entry, MODE.TINY, {
+      onInspectIconClick,
+      onDOMNodeMouseOut,
+      onDOMNodeMouseOver,
+    });
+    expect(renderRep(entry, MODE.TINY).text())
+      .toEqual("A → input#newtab-customize-button.bar.baz");
+
+    let node = renderedComponent.find(".objectBox-node");
+    let icon = node.find(".open-inspector");
```
bgrins: For inspect icon tests it'd be nice if we didn't need to retest the whole interaction for every rep that can have a  PropRep as a child. Either some kind of shared testing utility that does this event simulation / mocked calls dance, or if we could assume that the inspect icon works fine as long as the `.open-inspector` DOM Node is there (as I'm sure it's tested on its own)

nchevobbe:
You mean in the quick links ?

Le 28 juil. 2017 8:23 PM, "Brian Grinstead" <notifications@github.com> a
écrit :

bgrins:
> You mean in the quick links ?

Yes

nchevobbe:
Sure, I'll add some Set/Map examples in the launchpad in the entries PR since I'm modifying the test app there.

packages/devtools-reps/src/reps/tests/grip-map-entry.js
nchevobbe: Not sure about not testing the actual behaviour of the inspect icon, I'm afraid some cases could break it without us detecting it.
I filed https://github.com/devtools-html/devtools-core/issues/533 to add something to not be redundant for those icon tests.

PR #536, opened 2017-07-31
Change default properties node styling

nchevobbe:
First, we add the missing rule for the default-property class.
This patch changes how we style default properties node in
objects. Previously, we were looking for properties with the
same name as a window ownProperty. This could be dangerous
since you can have properties with the same name as one in
Window, but not necessarely under the window object.
Also, this can be weird to dim all the window default properties
when you expanded the node. If you did so, you are probably
interested in seeing what's inside.
Taking this into account, this patch applies the default-property class
to the default properties node and __proto__ ones, but only when they are
collapsed.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/index.css
+++ b/packages/devtools-reps/src/object-inspector/index.css
@@ -55,3 +55,8 @@ html[dir="rtl"] .arrow svg,
 .object-label {
   color: var(--theme-highlight-blue);
 }
+
+.lessen {
+  opacity: 0.6;
+}
+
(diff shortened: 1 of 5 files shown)


nchevobbe:
Here's how it looks : 

screen shot 2017-07-31 at 17 21 33

nchevobbe:
> Looks great. Please add some screenshots for marketing and posterity 📷

Sure, I added one just before your message came in :)

jasonLaster:
Guess i'm too fast for you :)

PR #537, opened 2017-07-31
Do not put a rejected promise into sourceMapRequests cache

tromey:
My previous patch in this area was mistaken - it correctly allowed
errors from _fetchSourceMap to be propagated to the caller, but it
also cached the rejected promise in sourceMapRequests, resulting in
the same error being reported via other APIs.  This patch fixes the
oversight by caching a non-rejecting promise; this results in the
other APIs keeping their previous behavior.


tromey:
> Is there a test that demonstrates the fix?

There's one in M-C already.  I'll see if I can write a local one though.

PR #543, opened 2017-08-02
Remove unused string.js file

nchevobbe:


nchevobbe:
r=me since it's a really small and harmless thing

PR #545, opened 2017-08-02
Fix Reps devDependencies

nchevobbe:


packages/devtools-reps/package.json
```diff
@@ -44,7 +44,9 @@
     "jest-cli": "^20.0.4",
     "jest-flow-transform": "^1.0.1",
     "react-addons-test-utils": "=15.3.2",
-    "redux-logger": "=3.0.6"
+    "react-redux": "4.4.5",
+    "redux-logger": "=3.0.6",
+    "redux": "3.5.2"
```
jasonLaster: Shouldn't redux come from the root package?

packages/devtools-reps/package.json
```diff
@@ -34,7 +34,7 @@
   },
   "devDependencies": {
     "devtools-config": "^0.0.15",
-    "devtools-launchpad": "^0.0.87",
+    "devtools-launchpad": "^0.0.89",
```
jasonLaster: Nice catch.

packages/devtools-reps/package.json
nchevobbe: You mean I should add the redux entries directly in `/package.json` ?

jasonLaster:
I think it might be there. Will be at my computer soon

PR #555, opened 2017-08-04
Add a prop to dim top level window node.

nchevobbe:
This is done to fill the need of the debugger Scopes panel,
since the global Window object is always shown.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/object-inspector.js
+++ b/packages/devtools-reps/src/object-inspector/tests/object-inspector.js
@@ -14,8 +14,10 @@ const {
 } = require("../utils");
 const { MODE } = require("../../reps/constants");
 const { Rep } = require("../../reps/rep");
+
 const gripRepStubs = require("../../reps/stubs/grip");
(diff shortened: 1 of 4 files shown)


packages/devtools-reps/src/object-inspector/tests/object-inspector.js
```diff
@@ -404,6 +406,66 @@ describe("ObjectInspector", () => {
...
+      dimTopLevelWindow: true,
+      getObjectProperties: () => {},
+      loadObjectProperties: () => {},
+    }));
+    const nodes = oi.find(".node");
+    const win = nodes.at(1);
+
+    // Make sure we target the window object.
+    expect(win.find(".objectBox-Window").exists()).toBeTruthy();
+    expect(win.hasClass("lessen")).toBeFalsy();
```
jasonLaster: I generally have many its here, but I think it's okay :)

Also, might be time to split up these tests too :)

nchevobbe: I put multiple `it`, it's cleaner.
And I'll split the test into multiple files in my PR for functions :)

PR #562, opened 2017-08-07
Remove type parameter in supportsObject.

### LINKED ISSUE #554: Remove type parameter in supportsObject
We currently pass a `type` argument to `supportsObject` functions in Reps.
This type is computed in `getRep` and only needs the grip.
I think it would be nice to be able to use those `supportsObject` function in the object inspector so we don't reimplement those checks there (in object-inspector/utils, we have nodeIsPromise, nodeIsFunction, …).
But using supportsObject at the moment feels weird because of the type argument.

So my proposal would be to get rid of this argument, have a getType function in rep utils, and call it from supportsObject when/if it's needed.

nchevobbe:
Fixes #554.
This removes the type parameter from supportsObject so those functions
can be used by another consumer than getRep.
The type computation is moved to its own function in utils.

### DIFF
--- a/packages/devtools-reps/src/reps/promise.js
+++ b/packages/devtools-reps/src/reps/promise.js
@@ -6,6 +6,7 @@
 const React = require("react");
 // Dependencies
 const {
+  getGripType,
   isGrip,
   wrapRender,
 } = require("./rep-utils");
@@ -100,11 +101,11 @@ function getProps(props, promiseState) {
 }
 
 // Registration
-function supportsObject(object, type, noGrip = false) {
+function supportsObject(object, noGrip = false) {
   if (noGrip === true || !isGrip(object)) {
     return false;
   }
-  return type === "Promise";
+  return getGripType(object, noGrip) === "Promise";
 }
 
 // Exports from this module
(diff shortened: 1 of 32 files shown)


packages/devtools-reps/src/reps/rep.js
```diff
@@ -105,24 +104,13 @@ const Rep = function (props) {
...
-  }
-
   for (let i = 0; i < reps.length; i++) {
     let rep = reps[i];
     try {
       // supportsObject could return weight (not only true/false
       // but a number), which would allow to priorities templates and
       // support better extensibility.
-      if (rep.supportsObject(object, type, noGrip)) {
+      if (rep.supportsObject(object, noGrip)) {
```
jasonLaster: :thumbsup:

packages/devtools-reps/src/reps/promise.js
```diff
@@ -100,11 +101,11 @@ function getProps(props, promiseState) {
 }
 
 // Registration
-function supportsObject(object, type, noGrip = false) {
+function supportsObject(object, noGrip = false) {
   if (noGrip === true || !isGrip(object)) {
     return false;
   }
-  return type === "Promise";
+  return getGripType(object, noGrip) === "Promise";
```
jasonLaster: not a big deal, but you use `==` in other places

PR #564, opened 2017-08-07
Make a little space between a label and submenu arrow (bug 1384892)

janodvarko:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1384892

Honza


janodvarko:
menu

PR #571, opened 2017-08-08
Use fetch in privileged-network-request and inhibit caching

### LINKED ISSUE #551: change privileged network request not to use cache
The code in `packages/devtools-utils/src/privileged-network-request.js` has a FIXME for not using the cache.  It would be good to fix this for source map fetching.  Currently one can wind up using a stale source map due to the cache.

One possible approach might be to change the code to use `fetch`.  Though, if doing this, we might as well just remove this code and implement it directly in `network-request.js`.

tromey:
This change privileged-network-request to use fetch rather than
XMLHttpRequest; and also changes it to respect the "loadFromCache"
option.  Fixes #551.


tromey:
I don't know how to test this in place; but I landed a cross-domain source-map test in M-C in preparation for this patch.

PR #572, opened 2017-08-08
Fix uglify by using a beta version in launchpad

zaggy:
This is connected to the #567 

Had to use beta version of `uglifyjs-webpack-plugin` because stable version uses uglify-js version which does not support es2015.

Also cleaned up some unused code.

### DIFF
--- a/packages/devtools-launchpad/webpack.config.js
+++ b/packages/devtools-launchpad/webpack.config.js
@@ -41,7 +41,6 @@ module.exports = (webpackConfig, envConfig) => {
     loader: `babel-loader?${defaultBabelPlugins.map(
       p => `plugins[]=${p}`
     )}&ignore=src/lib`
-    // isJavaScriptLoader: true
   });
   webpackConfig.module.rules.push({
     test: /\.svg$/,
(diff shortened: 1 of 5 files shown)


packages/devtools-launchpad/webpack.config.js
```diff
@@ -41,7 +41,6 @@ module.exports = (webpackConfig, envConfig) => {
     loader: `babel-loader?${defaultBabelPlugins.map(
       p => `plugins[]=${p}`
     )}&ignore=src/lib`
-    // isJavaScriptLoader: true
```
zaggy: Just removed unnecessary comment

packages/devtools-launchpad/webpack.config.js
```diff
@@ -145,16 +144,5 @@ module.exports = (webpackConfig, envConfig) => {
...
-  // destructuring params https://github.com/devtools-html/debugger.html/issues/67
-  if (getValue("transformParameters")) {
-    webpackConfig.module.rules.forEach(spec => {
-      if (spec.isJavaScriptLoader) {
-        const idx = spec.rules.findIndex(loader => loader.includes("babel-loader"));
-        spec.rules[idx] += "&plugins[]=transform-es2015-parameters";
-      }
-    });
-  }
-
```
zaggy: This is working anyway, because `spec.isJavascriptLoader` will always be `undefined` due to changes in webpack

nchevobbe:
is this still valid ?

jasonLaster:
Yes, the benefit is dead code elimination. 

@zaggy what's the status here?

zaggy:
I think this should be closed. With Webpack 4 we will not need this at all

jasonLaster:
yep, makes sense

Issue #580, opened 2017-08-09
Prefs of type Json should use type "String" instead of "Char"


juliandescottes:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1388825 for the initial debugger bug report.

Summary:
If you try to save a string containing unicode characters such as "⌘" as a Char pref, the character will be mangled. When trying to retrieve the value, the helper crashes when calling JSON.parse. Switching the pref to type String fixes the issue.

The code responsible for this is: https://github.com/devtools-html/devtools-core/blob/9881990dd1e37c0ede527e2e88979a3c1c110037/packages/devtools-modules/src/prefs.js#L107-L113

I don't know if we can simply change the type of existing preferences? I tested locally and it seemed to work but it would be nice if someone else could confirm? cc @jryans @bgrins maybe?

Typically those prefs are stored as follows in a prefs.js file:
> user_pref("devtools.debugger.tabs", "[\"http://www.piskelapp.com/static/js/piskel-animated-preview.js?20170502\"]");

Nothing here about the preference type so I guess the migration should be smooth.

juliandescottes:
For the record, in the debugger we have:
- "debugger.tabs"
- "debugger.pending-selected-location"
- "debugger.pending-breakpoints"
- "debugger.expressions"

that are using the Json type. Not sure if other projects are using the prefs helper.

juliandescottes:
Note that this can only tested properly in panel with a mochitest unless we "nerf" our Services shim in order to replicate the behavior of the actual Services.prefs for Char preferences.

PR #581, opened 2017-08-09
Issue #580 : switch to type String for Json prefs

juliandescottes:
I tested locally this change by modifying the corresponding code in the debugger bundle, and my breakpoints, tabs, expressions were still the ones from the previous session (which means that changing the type from Char to String didn't seem to create any issue).

Note that this won't fix existing bad preferences, such as the one that triggered https://bugzilla.mozilla.org/show_bug.cgi?id=1388825 , since the issue occurs on save.

### DIFF
--- a/packages/devtools-modules/src/prefs.js
+++ b/packages/devtools-modules/src/prefs.js
@@ -29,7 +29,7 @@ const EventEmitter = require("./utils/event-emitter");
  * @param string prefsRoot
  *        The root path to the required preferences branch.
  * @param object prefsBlueprint
- *        An object containing { accessorName: [prefType, prefName] } keys.
+ *        An object containing { accessorName: [prefType, prefName, prefDefault] } keys.
  */
 function PrefsHelper(prefsRoot = "", prefsBlueprint = {}) {
   EventEmitter.decorate(this);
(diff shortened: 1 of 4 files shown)


juliandescottes:
It would be nice if we could fallback to a sensible default value in case JSON.parse fails. However this is challenging given the current usage the "Json" pref in the debugger. Two prefs are objects, and two prefs are arrays. 

Maybe the prefsBlueprint could also accept an optional default value.

packages/devtools-modules/src/prefs.js
```diff
@@ -60,6 +60,7 @@ function get(cache, prefType, prefsRoot, prefName) {
   if (cachedPref !== undefined) {
     return cachedPref;
   }
+  console.log(prefType, "@@@@@@@@@");
```
jasonLaster: :)

juliandescottes:
Added two additional changesets to support String prefs in the Services shim, and support a default/fallback value in the PrefsHelper.

If you are ok with the approach the only thing left to do (after bumping and releasing a new devtools modules) will be to update https://github.com/devtools-html/debugger.html/blob/0ab605e797178aac92e9fd5480b63844ca4be15b/src/utils/prefs.js to add default values.

packages/devtools-modules/src/prefs.js
juliandescottes: oops

juliandescottes:
Thanks for the review! Merging.

PR #584, opened 2017-08-10
[WIP] Fix Tree component and add storybook

nchevobbe:


packages/devtools-components/webpack.config.js
```diff
@@ -0,0 +1,160 @@
...
+      webpackConfig.resolve.alias.react = "react/lib/ReactUMDEntry";
+    }
+  }
+
+  webpackConfig.node = { fs: "empty" };
+
+  webpackConfig.plugins = webpackConfig.plugins || [];
+  webpackConfig.plugins.push(
+    new webpack.DefinePlugin({
+      "process.env": {
```
jasonLaster: do we need the define plugin?

packages/devtools-components/webpack.config.js
```diff
@@ -0,0 +1,160 @@
...
+    })
+  );
+
+  if (isDevelopment()) {
+    webpackConfig.module.rules.push({
+      test: /\.css$/,
+      use: [
+        { loader: "style-loader" },
+        { loader: "css-loader", options: { importLoaders: 1 } },
+        // {
```
jasonLaster: what's going on here?

packages/devtools-components/stories/tree.js
```diff
@@ -0,0 +1,86 @@
...
+        focused ? " (focused)" : null
+      ),
+      getRoots: () => ["A"],
+      getKey: x => "key-" + x,
+      itemHeight: 1,
+      onExpand: x => TEST_TREE.expanded.add(x),
+      onCollapse: x => TEST_TREE.expanded.delete(x),
+      isExpanded: x => TEST_TREE.expanded.has(x),
+    });
+  });
```
jasonLaster: these tests are not run on CI,

but i'd love to get them running on CI so we can make sure they don't regress: https://github.com/mthuret/storybook-addon-specifications

packages/devtools-components/src/tree.js
```diff
@@ -94,22 +131,6 @@ const TreeNode = createFactory(createClass({
       // unless there is an input/button child.
       dom.button(this._buttonAttrs)
```
jasonLaster: does this still exist?

packages/devtools-components/webpack.config.js
nchevobbe: I copied the webpack config from the launchpad, which is not the right thing to do :)

packages/devtools-components/stories/tree.js
nchevobbe: Yeah, I wanted to have story book so we can have at least a way to see how it looks.
This component should be properly tested as well (render, keyboard navigation, …)

packages/devtools-components/src/tree.js
nchevobbe: Didn't check, I will !

PR #585, opened 2017-08-10
chore: add assets/ to .gitignore

juliandescottes:
When running yarn install on the root of devtools-core, `lerna bootstrap` is creating an assets folder which contains a module-manifest.json file.

This should either be ignored or added to source control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ npm-debug.log
 .idea/
 .vscode/
 storybook-static/
+assets/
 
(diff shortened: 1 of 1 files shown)


nchevobbe:
I think https://github.com/devtools-html/devtools-core/pull/579 is going to fix that

juliandescottes:
@nchevobbe no it doesn't. #579  has just been merged but I still get the issue.

PR #588, opened 2017-08-10
Add a prop to have ObjectInspector inline

nchevobbe:
This adds an `inline` class to the tree which display the tree as `inline-block`, which might be useful for some consumer (e.g. the console where you can have multiple ObjectInspectors side by side; even if for now we are using flexbox to lay out them).
This changes needed devtools-components 0.0.2 for the Tree component to handle the classname override.

Also, this was a good opportunity to add tests for the class names in the object inspector.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/component/classnames.js
+++ b/packages/devtools-reps/src/object-inspector/tests/component/classnames.js
@@ -23,32 +23,29 @@ function generateDefaults(overrides) {
 describe("ObjectInspector - classnames", () => {
   it("has the expected class", () => {
     const props = generateDefaults();
-    const oi = ObjectInspector(props);
-    const wrapper = mount(oi);
+    const wrapper = mount(ObjectInspector(props));
 
     expect(wrapper.hasClass("tree")).toBeTruthy();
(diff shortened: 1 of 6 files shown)


packages/devtools-reps/src/object-inspector/tests/component/classnames.js
```diff
@@ -0,0 +1,54 @@
...
+describe("ObjectInspector - classnames", () => {
+  it("has the expected class", () => {
+    const props = generateDefaults();
+    const oi = ObjectInspector(props);
+    const wrapper = mount(oi);
+
+    expect(wrapper.hasClass("tree")).toBeTruthy();
+    expect(wrapper.hasClass("inline")).toBeFalsy();
+    expect(wrapper.hasClass("nowrap")).toBeFalsy();
+    expect(oi).toMatchSnapshot();
```
jasonLaster: you might want to try `expect(oi.debug()).toMatchSnapshot();`

this will give you the html, which will be more releated

nchevobbe: interesting, thanks !

PR #592, opened 2017-08-11
Handle Proxy target and handler in ObjectInspector.

nchevobbe:
Previously, when the user expanded a Proxy object, we loaded the properties
of the target, which is only half the information the user might want to see.
Now we build 2 ndoes, target and handler, that can be expanded as well so
the user can inspect the Proxy object properly.
Some tests were added to make sure we handle proxies as expected, and the
setExpanded function was refactored a bit to be more readable.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/component/proxy.js
+++ b/packages/devtools-reps/src/object-inspector/tests/component/proxy.js
@@ -1,0 +1,82 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ /* global jest */
+
+const { mount } = require("enzyme");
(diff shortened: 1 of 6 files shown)


packages/devtools-reps/src/object-inspector/tests/component/proxy.js
```diff
@@ -0,0 +1,82 @@
...
+    });
+    const oi = mount(ObjectInspector(props));
+    expect(oi.debug()).toMatchSnapshot();
+
+    const nodes = oi.find(".node");
+    /*
+     * The OI should look like:
+     * ▶︎ Proxy
+     *   ▶︎ <target> : Object { … }
+     *   ▶︎ <handler> : Array [ … ]
```
janodvarko: heh, nice structured comment :-)

nchevobbe:
Thanks for the review Honza.
I added a Proxy example to the test app.
For the changes on the Proxy rep, let's file an issue to discuss it.

PR #603, opened 2017-08-17
Bump Launchpad

jasonLaster:
bumps the launchpad and introduces a small fix for making bundles


packages/devtools-launchpad/src/tools/mc/make-bundle.js
```diff
@@ -13,7 +13,7 @@ function makeBundle({ outputPath, projectPath, watch = false, updateAssets = fal
   const webpackConfig = require(path.resolve(projectPath, "webpack.config.js"));
 
   return new Promise((resolve, reject) => {
-    if (updateAssets) {
+    if (!updateAssets) {
```
jasonLaster: got the negation wrong. If we're not updating the assets we should delete the records path...

packages/devtools-launchpad/src/tools/mc/make-bundle.js
```diff
@@ -11,12 +11,11 @@ function makeBundle({ outputPath, projectPath, watch = false, updateAssets = fal
   process.env.OUTPUT_PATH = outputPath;
 
   const webpackConfig = require(path.resolve(projectPath, "webpack.config.js"));
+  if (!updateAssets) {
+    delete webpackConfig.recordsPath;
+  }
```
jasonLaster: got the negation wrong. If we're not updating the assets we should delete the records path...

PR #605, opened 2017-08-19
Update lanuchpad config

jasonLaster:
* add an extra `options` option for passing exclude mappings so the debugger can decide what to exclude
* drop postcss from production builds. we don't want to add prefixes or drop `inline-start` props

### DIFF
--- a/packages/devtools-source-editor/examples/index.js
+++ b/packages/devtools-source-editor/examples/index.js
@@ -5,7 +5,7 @@
 const SourceEditor = require("../src/source-editor");
 
 require("devtools-launchpad/src/lib/themes/light-theme.css");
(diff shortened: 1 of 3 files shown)


packages/devtools-source-editor/examples/index.js
```diff
@@ -5,7 +5,7 @@
 const SourceEditor = require("../src/source-editor");
 
 require("devtools-launchpad/src/lib/themes/light-theme.css");
-require("devtools-launchpad/src/lib/themes/dark-theme.css");
+require("devtools-modules/src/themes/dark-theme.css");
```
nchevobbe: why only change the dark theme path ?

jasonLaster: haha - oops

PR #611, opened 2017-08-24
Handle tasks completing when the worker is shutdown

jasonLaster:
Fixes an annoying issue where the worker would crash with a typeError where `removeEventListener` was not defined on undefined.

### DIFF
--- a/packages/devtools-utils/src/tests/worker-utils.js
+++ b/packages/devtools-utils/src/tests/worker-utils.js
@@ -20,11 +20,9 @@ describe("worker utils", () => {
     const dispatcher = new WorkerDispatcher();
     const terminateMock = jest.fn();
 
-    global.Worker = jest.fn(() => {
-      return {
-        terminate: terminateMock
-      };
-    });
+    global.Worker = jest.fn(() => ({
+      terminate: terminateMock
+    }));
 
     dispatcher.start();
     dispatcher.stop();
@@ -75,7 +73,7 @@ describe("worker utils", () => {
 
     expect(postMessageMock.mock.calls[0][0]).toEqual({
       id: 53,
-      error: "Error: failed",
+      error: "Error: failed"
     });
   });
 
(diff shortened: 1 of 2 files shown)


packages/devtools-utils/src/worker-utils.js
```diff
@@ -45,6 +45,10 @@ WorkerDispatcher.prototype = {
             return;
           }
 
+          if (!this.worker) {
+            reject("Oops, The worker has shutdown!");
```
codehag: :+1:

packages/devtools-utils/src/tests/worker-utils.js
```diff
@@ -61,17 +65,43 @@ describe("worker utils", () => {
...
+    handler({ data: { id: 53, method: "doSomething", args: [] } });
 
     expect(postMessageMock.mock.calls[0][0]).toEqual({
       id: 53,
-      error: new Error("failed"),
+      error: new Error("failed")
     });
   });
+
+  it("test a task completing when the worker has shutdown", () => {
```
codehag: \o/ yey for tests!

packages/devtools-utils/src/tests/worker-utils.js
```diff
@@ -61,17 +65,43 @@ describe("worker utils", () => {
...
+    const dispatcher = new WorkerDispatcher();
+    const postMessageMock = jest.fn();
+    const addEventListenerMock = jest.fn();
+    const terminateMock = jest.fn();
+
+    global.Worker = jest.fn(() => {
```
codehag: nit: we can probably use a normal function here rather than a jest function

packages/devtools-utils/src/tests/worker-utils.js
```diff
@@ -61,17 +65,43 @@ describe("worker utils", () => {
...
+    });
+
+    dispatcher.start();
+    const task = dispatcher.task("foo");
+    const resp = task("bar");
+    resp.catch(e => expect(e).toEqual("Oops, The worker has shutdown!"));
+
+    const listener = addEventListenerMock.mock.calls[0][1];
+    dispatcher.stop();
+    listener({ data: { id: 1 } });
```
codehag: what is happening here?

Issue #604, opened 2017-08-17
Improve the launchpad sidebar


jasonLaster:
It would be great if the launchpad sidebar got some love

* the title included launchpad and the rocket
* settings was visually separated from the other tabs so that it is more obvious

I started the work here and got stuck

https://github.com/devtools-html/devtools-core/compare/master...jasonLaster:launchpad-sidebar?expand=1

<img src='https://shipusercontent.com/47adab20f9cc6ee6485ba2e657d070b7/Screen%20Shot%202017-08-17%20at%2012.36.45%20PM.png' title='Screen Shot 2017-08-17 at 12.36.45 PM.png' width=586>

Ronsho:
On it.

connorholyday:
It looks like this has been worked on, is there anything further that needs doing?

<img width="1419" alt="image" src="https://user-images.githubusercontent.com/4559119/43223164-1cdc7414-904a-11e8-8008-c3586e875579.png">

PR #613, opened 2017-08-26
Feature/boost sidebar

Ronsho:
Associated Issue: #604

### Summary of Changes

* Added launchpad icon. 
* Added a border between the settings and rest of the sidebar items

screen shot 2017-08-26 at 6 22 07 pm
screen shot 2017-08-26 at 6 22 18 pm


packages/devtools-launchpad/src/components/Sidebar.js
```diff
@@ -19,6 +21,35 @@ const Sidebar = React.createClass({
     onSideBarItemClick: React.PropTypes.func.isRequired
   },
 
+  renderTitle(title) {
+
+    return dom.div({className: "title-wrapper"},
+                    dom.h1({}, "Debugger"),
```
jasonLaster: lets use `title` here...

That'll mean that we get Launchpad/Launchpad... but we could do something cute in that case too... like change the development.json to be "Launchy"

packages/devtools-launchpad/src/components/Sidebar.js
```diff
@@ -19,6 +21,35 @@ const Sidebar = React.createClass({
...
+                    dom.div({className: "launchpad-container"},
+                            dom.div({className: "launchpad-container-icon", dangerouslySetInnerHTML: {__html: rocketSvg }}),
```
jasonLaster: nice - i wouldn't have thought of this. I don't think we have a better way :)

This might be an alternative, but i'm not sure:
https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/Breakpoint.js#L10-L22

either way... this is cool for our usecase

jasonLaster: we might want to vertically align Launchpad a bit and add some more spacing

packages/devtools-launchpad/package.json
```diff
@@ -102,6 +100,7 @@
     "@kadira/storybook": "^2.35.3",
     "ipaddr": "^0.0.9",
     "jest": "^19.0.2",
+    "svg-inline-loader": "^0.8.0",
```
jasonLaster: if we have svg loader, should we use that? If not we can remove this:

this is how we use it
https://github.com/devtools-html/debugger.html/blob/master/assets/images/Svg.js#L1

packages/devtools-launchpad/src/components/Sidebar.js
Ronsho: Yep, also think it's good for now :)

packages/devtools-launchpad/src/components/Sidebar.js
Ronsho: How about setting the title property in the development.json to be "Debugger" and adding a new property called "subtitle" with "Launchpad" value?

packages/devtools-launchpad/package.json
Ronsho: I use it in webpack.config to load svg

packages/devtools-launchpad/src/components/Sidebar.js
Ronsho: Do you mean adding more space between the icon and the "Launchpad"?

packages/devtools-launchpad/src/components/Sidebar.js
jasonLaster: Hmm, I don't anticipate anything other than launchpad at the moment. We might want something in the future, but we can always do that then :)

packages/devtools-launchpad/src/components/Sidebar.js
Ronsho: screen shot 2017-08-28 at 3 22 06 pm

Done

packages/devtools-launchpad/src/components/Sidebar.js
```diff
@@ -19,6 +21,35 @@ const Sidebar = React.createClass({
     onSideBarItemClick: React.PropTypes.func.isRequired
   },
 
+  renderTitle(title) {
+
+    return dom.div({className: "title-wrapper"},
+                    dom.h1({}, title),
+                    dom.div({className: "launchpad-container"},
+                            dom.div({className: "launchpad-container-icon", dangerouslySetInnerHTML: {__html: rocketSvg }}),
+                            dom.h2({className: "launchpad-container-title"}, title)));
```
jasonLaster: This should be "Launchpad"

PR #615, opened 2017-08-28
Refactor/sidebar sass

Ronsho:
* Added Sass support to the project (linting + webpack)
* refactored sidebar.css to sidebar.scss


packages/devtools-launchpad/src/components/sidebar.scss
```diff
@@ -0,0 +1,84 @@
+$sidebar-width: 200px;
```
jasonLaster: can we remove `sidebar.css`? Can we rename `sidebar.scss` to `sidebar.css`? It'd be nice not to differentiate between the two

Ronsho: Removed the sidebar.css. Don't think it's common to name give a .css suffix to a sass file...

wldcordeiro:
Hadn't we started using PostCSS in this or the Debugger? I think it'd be nice if we could continue using it over Sass or maybe a CSS-in-JS solution.

nyrosmith:
I think a css-in-js solution would be really cool because it would give us the best of both worlds

jasonLaster:
I agree, it'd be neat to try css-modules and glamorous. Lets make separate branches for each to see what it would look like

digitarald:
@jasonLaster can this be closed for now or and file any follow up work?

Issue #448, opened 2017-06-13
 Make ErrorRep look more like real error


nchevobbe:
_moved from [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1317325)_

The Rep was created with the output of the old console in mind, i.e. a simple string which describes the error and the stack, but not really actionable (except clicking on the message to open the variable view).

We could do a better job here : 
- have more error-like styling
- have collapsible stacktrace
- allow the "location" (file + line) of the trace clickable to open in the debugger

We could use the frame component we already use in the console to have proper stack trace, and have some mechanism to expand/collapse the stacktrace (by taking inspiration on how LongStringRep works maybe ? ).

nchevobbe:
@arashm I wonder if this is something you'd be interested in ?

nchevobbe:
That would be super nice.
Of you want to know what it is about, in Nightly you can enter this in the console (which use reps to output the data) : 
```js
let foo = () => new Error("boo")
foo()
```

Which would be rendered as : 
```
 Error: boo
Stack trace:
foo@debugger eval code:1:17
@debugger eval code:2:1
```
All in "plain text" without being able to link to the proper location (and without sourcemap support too).

Why is that ? If you look at the [grip we get when we log an error](https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-reps/src/reps/stubs/error.js#L31), you can see that the `stack` property is only one big string. We just happen to put this into the Rep as is.

The idea here would be to parse this string to extract a proper stacktrace array, which would contain a structure that holds all that we need (function name, location. line and column numbers , ...) so we can output them in a nicer way.

The next step would then to be able to copy the `Stacktrace` [component we have in mozilla-central](http://searchfox.org/mozilla-central/source/devtools/client/shared/components/stack-trace.js) that would take care of the layout when we pass it the proper props. You can even see that a sourcemapService can be passed, which would then take care of the sourcemap if there are some.
I might do this (copy from m-c and put it in packages/devtools-components)


---
In order to help you see what the Rep look like when you do changes in it, you can run the test app (with `yarn start` in devtools-reps)
Hope this helps, don't hesitate to ask questions here or in Slack :)

gregtatum:
[...]

PR #621, opened 2017-08-29
Add formatting for stack trace

arashm:
Not sure if this is what you expected for #448, but here I give it a try.

Associated Issue: #448 

### Summary of Changes

* Import `fmt` from m-c
* Implement `parseStack` using `match` and regex
* Import `formatTrace` from m-c with some minor changes

### Test Plan

Updated the test file `error.js` based on the output changes

### DIFF
--- a/packages/devtools-reps/src/reps/error.js
+++ b/packages/devtools-reps/src/reps/error.js
@@ -36,18 +36,116 @@ function ErrorRep(props) {
     : `${name}: ${preview.message}`;
 
   if (preview.stack && props.mode !== MODE.TINY) {
+    let formattedStack = formatTrace(parseStack(preview.stack));
     /*
       * Since Reps are used in the JSON Viewer, we can't localize
       * the "Stack trace" label (defined in debugger.properties as
       * "variablesViewErrorStacktrace" property), until Bug 1317038 lands.
       */
-    content = `${content}\nStack trace:\n${preview.stack}`;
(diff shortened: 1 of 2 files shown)


nchevobbe:
Hello @arashm , thanks for working on this !
I'd say this is going into the right direction but I think we can do even better :)

The main thing, parsing the stack trace, is really nice, that's what we'll rely on.
What we could do is display the stack trace as a true html list. Which means building a ul/li list with the parsed stack:  
```js
const stack = parseStack(preview.stack);
content = dom.div({},
  `${content}\nStack trace:`,
  dom.ul(
    {}, 
    stack.map(frame => dom.li(
      {}, 
      // build the frame text here
    )))
)`;
```
This would be a good start. An even better thing would be to re-use the [Stacktrace and Frame components from mozilla-central](http://searchfox.org/mozilla-central/source/devtools/client/shared/components/stack-trace.js), which handle clicks on the source (to navigate to the debugger for example), and most important: sourcemaps.

I have a stale branch where I imported those files in devtools-core so we can use it in the console, I may revive and finish that.

In the meantime, feel free to implement the stacktrace list, I do think there is value in having this as a first step (I'd say that we don't need fmt though, let's display this as is). And then I'll ping you when/if I manage to get the Stacktrace component in devtools-core so we can use it.

nchevobbe:
Closing in favor of https://github.com/devtools-html/devtools-core/pull/942

PR #624, opened 2017-08-29
AST scopes/binding to names mapping.

yurydelendik:
Associated Issue: https://github.com/devtools-html/debugger.html/issues/3427

### Summary of Changes

* merges parsed scopes in generated source with names information from source maps

### Test Plan

See https://github.com/devtools-html/debugger.html/pull/3817

### DIFF
--- a/packages/devtools-map-bindings/src/tests/scopes.js
+++ b/packages/devtools-map-bindings/src/tests/scopes.js
@@ -1,0 +1,159 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const { getLocationScopes } = require("../scopes");
+const { SourceMapConsumer, SourceMapGenerator } = require("source-map");
+
+jest.mock("devtools-utils/src/network-request");
+
+test("getLocationScopes", async () => {
+  const source = {
+    id: "source-mapped.js",
+    sourceMapURL: "source-mapped.js.map",
+    url: "http://example.com/source-mapped.js"
+  };
+
+  // Testing something like:
+  // `function () { var zero; { let one = 1; one++; } zero = 0; }`
+  var map = new SourceMapGenerator({
+    file: "source-mapped.js"
+  });
+
+  map.addMapping({
+    generated: {
+      line: 1,
+      column: 3
+    },
+    source: "foo.js",
+    original: {
+      line: 1,
+      column: 1
+    },
+    name: "zero"
+  });
+
+  map.addMapping({
+    generated: {
+      line: 11,
+      column: 15
+    },
+    source: "foo.js",
+    original: {
+      line: 33,
+      column: 5
+    },
+    name: "one"
+  });
+
+  map.addMapping({
+    generated: {
+      line: 11,
+      column: 35
+    },
+    source: "foo.js",
+    original: {
+      line: 33,
+      column: 9
+    },
+    name: "one"
+  });
+
+  map.addMapping({
+    generated: {
+      line: 20,
+      column: 5
+    },
+    source: "foo.js",
+    original: {
+      line: 50,
+      column: 1
+    },
+    name: "zero"
+  });
+
+  map = SourceMapConsumer.fromSourceMap(map);
+
+  // The generated source predends to have two scopes and the `zero` and `one`
+  // variables were changes to `z` and `o` respectively.
+
+  // Testing two scopes. The location of interest located somewhere at
+  // `... { /* here */ let one = 1 ...`.
+  const mapped1 = await getLocationScopes(
+    map,
+    [
+      {
+        type: "block",
+        start: { sourceId: source.id, line: 30, column: 1 },
+        end: { sourceId: source.id, line: 39, column: 1 },
+        bindings: {
+          o: [
+            { sourceId: source.id, line: 11, column: 15 },
+            { sourceId: source.id, line: 11, column: 33 }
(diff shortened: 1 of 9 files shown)


packages/devtools-source-map/src/source-map.js
```diff
@@ -182,6 +184,108 @@ async function getOriginalLocation(location: Location) : Promise<Location> {
...
+    return null;
+  }
+
+  return generatedSourceScopes.map(scope => {
+    // The scope has bindings that contains positions of reference to the
+    // generated name (that may has different original name). Going over all
+    // these positions, check what original name is listed there. To make things
+    // more interesting the same generated name can potentially be used for
+    // different original names.
+    let trackedBindings: { [string]:
```
jasonLaster: This type should be defined above:

`let trackedBindings: TrackedBindings`

packages/devtools-source-map/src/source-map.js
```diff
@@ -182,6 +184,108 @@ async function getOriginalLocation(location: Location) : Promise<Location> {
...
+      {
+        name: string,
+        bestLocation: Location
+      }
+    } = Object.create(null);
+    Object.keys(scope.bindings).forEach(name => {
+      const positions = scope.bindings[name];
+      positions.forEach(position => {
```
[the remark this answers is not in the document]
jasonLaster: `positions.reduce` might need the initial value `positions.reduce(position => findBestLocations(position), {}`

packages/devtools-source-map/src/source-map.js
yurydelendik: done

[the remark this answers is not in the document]
yurydelendik: done

packages/devtools-source-map/src/source-map.js
```diff
@@ -182,6 +184,108 @@ async function getOriginalLocation(location: Location) : Promise<Location> {
...
+          ) {
+            trackedBinding.name = name;
+            trackedBinding.bestLocation = position;
+          }
+        }
+      });
+    });
+
+    let bindings: { [string]: string } = Object.create(null);
+    Object.keys(trackedBindings).forEach(name => {
```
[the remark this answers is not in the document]
yurydelendik: fixed here and above.

packages/devtools-source-map/src/source-map.js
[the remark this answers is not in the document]
yurydelendik: find best location shall be performed across all generated names -- I reviewed summary comment before algorithm and provided example why things are done the way they are.

packages/devtools-source-map/src/source-map.js
```diff
@@ -182,6 +189,140 @@ async function getOriginalLocation(location: Location) : Promise<Location> {
...
+    !originalPositionTest ||
+    originalPositionTest.column !== originalPosition.column ||
+    originalPositionTest.line !== originalPosition.line
+  ) {
+    return null;
+  }
+  return originalPosition.name;
+}
+
+async function getLocationScopes(
```
jasonLaster: I think `getLocationScope` is sufficiently complex to justify its own module with some helper methods.

```js

function appendTrackedBinding() {}

function appendBinding() {}

function getScopeBindings() {}

function getLocationScope() {

  generatedSourceScopes.map(scope => {
    // ...
    const trackedBindings = bindings.reduce(appendTrackedBinding, {})
    const bindings = trackedBindings.reduce(appendBinding, {});
    return {
      type: scope.type,
      bindings,
    };
  })
}

module.exports = { getLocationScope}
```

yurydelendik: appendTrackedBinding and appendBinding extraction is not possible due to dependency on location/currentPosition.

yurydelendik: getLocationScope also depends on _getSourceMap

jasonLaster: we might need to do a bit more splitting to share `_getSourceMap` between modules.

we also might need to pass in the closure state:

```js
bindings.reduce(binding => appendTrackedBinding(binding, location, ...), {})
```

yurydelendik: Did not extracted _getSourceMap -- just exposed it from ./source-map

Moved scopes into separate file and refactored to have appendTrackedBinding.

jasonLaster:
I think it would be nice to switch to using a fixtures:

1. we can visualize the mappings 
2. easily regenerate them
3. be sure we didn't hardcode a bug in the mapping

Here is a [commit] where i setup uglify to make to do the minifying.

[commit]: https://github.com/devtools-html/devtools-core/commit/0e6b1da757999ff9054b40e0199f1dba794141f0

packages/devtools-source-map/src/scopes.js
```diff
@@ -0,0 +1,193 @@
...
+        generatedName,
+        nameLocation,
+        map,
+        currentPosition
+      ),
+    trackedBindings
+  );
+}
+
+async function getLocationScopes(
```
tromey: I'd like jsdoc for all new exported functions.  Lack of documentation is one of the bigger problems in this entire module already.

packages/devtools-map-bindings/src/tests/scopes.js
```diff
@@ -0,0 +1,159 @@
...
+      line: 20,
+      column: 5
+    },
+    source: "foo.js",
+    original: {
+      line: 50,
+      column: 1
+    },
+    name: "zero"
+  });
```
jasonLaster: we can clean this up after we add the parser code and can improve the test in one pass.

PR #627, opened 2017-08-30
Add source-utils from m-c into devtools-modules

nchevobbe:
This takes the code from https://hg.mozilla.org/mozilla-central/file/018b3829d0a7/devtools/client/shared/source-utils.js and https://hg.mozilla.org/mozilla-central/file/018b3829d0a7/devtools/client/shared/test/unit/test_source-utils.js for the test. The test was converted to jest.

This is needed so we can import Stacktrace and Frames component from mozilla-central as well


nchevobbe:
Thanks @tromey

PR #628, opened 2017-08-30
Components storybook

nchevobbe:
**Add storybook for devtools-component**

This will allow us to have a better way to hack on those components. (I plan to add Stacktrace and Frames soon)

**Fix Tree component**

- Fix ESLint warnings
- Simplify the Arrow component
- Do not render the Arrow component if we don't need to
- Use an SVG for the arrow

The reasoning behind this is that in the object inspector (and I think it might be the same in the source panel in the debugger), we recreate the arrow, which mean we do 2 times the same work, and might end up with different arrow in different places.
The arrow size and color can still be modified with CSS, as well as the indent width.

This is a ground work, I want to have tests for this component, as well as a solid keyboard navigation.


nchevobbe:
CI is complaining about different react versions on devtools-reps ... I'll see what can be done

packages/devtools-components/package.json
```diff
@@ -5,10 +5,18 @@
...
   "author": "Jason Laster",
   "license": "MPL-2.0",
   "dependencies": {
+    "react-dom": "=15.3.2",
+    "react": "=15.3.2",
+    "svg-inline-loader": "^0.8.0",
+    "svg-inline-react": "^1.0.2"
+  },
+  "devDependencies": {
+    "@kadira/storybook": "^2.35.3"
```
jasonLaster: they might have a 3.0 you could steal from. Check the debugger config

packages/devtools-components/src/tree.js
```diff
@@ -2,7 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const { DOM: dom, createClass, createFactory, PropTypes } = require("react");
+const { DOM: dom, createClass, createFactory, createElement, PropTypes } = require("react");
+const InlineSVG = require("svg-inline-react");
+const svgArrow = require("./images/arrow.svg");
+require("./tree.css");
```
jasonLaster: We made some changes to tree so that it would work in chrome, as well as support drag events that left the window. 

It'd be nice to look at the commits to this file and make sure that we address them.

packages/devtools-components/stories/tree.js
```diff
@@ -0,0 +1,93 @@
...
+          itemHeight: 1,
+          onExpand: x => this.setState({expanded: TEST_TREE.expanded.add(x)}),
+          onCollapse: x => {
+            const expanded = new Set(this.state.expanded);
+            expanded.delete(x);
+            this.setState({expanded});
+          },
+          isExpanded: x => this.state && this.state.expanded.has(x),
+        });
+      }
```
jasonLaster: would be nice to exercise autoExpandDepth: 1 and 2...

jasonLaster: would be nice to exercise autoExpandDepth: 1 and 2...

nchevobbe:
Updated to storybook 3+, like the debugger
Added stories for different autoExpandDepth and multiple roots

nchevobbe:
Tests are still failing because of snapshot changes.
Why does devtools-reps pick changes from devtools-component ? It should work with the npm package, which does not have those changes 🤔

jasonLaster:
Lerea moves the React install to the top dir to save time. Unless the
packages are different

nchevobbe:
@jasonLaster to you think I can merge this (even if CI fails), and then I'll bump up devtools-component to `0.0.3`, and then use the new version in devtools-reps (which means updating the snapshots, hence fixing the failures)

PR #630, opened 2017-08-31
Fix grip map entry test

nchevobbe:
Remove the `.only` on test


jasonLaster:
we do this too. we should lint against it

PR #634, opened 2017-08-31
Lazy load buckets, properties, entries, symbols and prototypes.

nchevobbe:
This patch turns the object inspector into a more smart component, that can decide when and what to load from the server.
To achieve that, a new `createObjectClient` prop is created, which will allow the object inspector to create an `ObjectClient` and calls the server in order to fetch properties, entries, symbols and prototype.
This allow us to only load the data when we need it, which will prevent the server to hang on very large arrays (and objects soon).

The ObjectInspector keep track of its expanded nodes (like it used to), as well as the loaded properties and actors for a node (via its path).

Another new prop, `releaseActor`, will allow the object inspector to release all the actor when it unmounts so we don't consume too much memory.

The utils.js file has bee moved to a util folder and renamed node.js, as another new util `client.js` file is created to deal with the ObjectClient calls.

This PR does not contains tests and reps-test-app changes so it isn't too long (so it will fail on CI).
I'm going to publish those as independent PR to make review easier.


packages/devtools-reps/src/object-inspector/utils/node.js
```diff
@@ -619,101 +608,253 @@ function setNodeChildren(
...
+
+  // Because we are dynamically creating the tree as the user
+  // expands it (not precalculated tree structure), we cache child
+  // arrays. This not only helps performance, but is necessary
+  // because the expanded state depends on instances of nodes
+  // being the same across renders. If we didn't do this, each
+  // node would be a new instance every render.
+  // If the node needs properties, we only add children to
+  // the cache if the properties are loaded.
+  const addToCache = (children: Array<Node>) => {
```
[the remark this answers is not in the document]
nchevobbe: Oh, I did not think of this.
Maybe some mechanism to invalidate the cache at some point ?
Do you still have the same component instance over stepping though ?

nchevobbe:
> This is hard to review without tests to help walk through the change. Are the tests already there or are they coming?

The test PR is here https://github.com/devtools-html/devtools-core/pull/635

packages/devtools-reps/src/object-inspector/utils/node.js
jasonLaster: I think we'll unmount the component when we resume, and if the cache is in the component we're safe...

packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -66,10 +79,9 @@ type Props = {
   roots: Array<Node>,
   disableWrap: boolean,
   dimTopLevelWindow: boolean,
-  getObjectEntries: (actor:string) => ?LoadedEntries,
-  getObjectProperties: (actor:string) => ?LoadedProperties,
-  loadObjectEntries: (value:RdpGrip) => void,
-  loadObjectProperties: (value:RdpGrip) => void,
+  releaseActor: string => void,
```
bgrins: There are two types of `releaseActor` here.  Is this intentional, and does flow let you collapse this into one line? Or if not, can you move them to adjacent lines?

nchevobbe: not wanted, it should be only the one taking a string as an argument

packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -148,21 +167,43 @@ class ObjectInspector extends Component {
   }
 
   state: State;
+
+  shouldComponentUpdate(nextProps : Props, nextState: State) {
```
janodvarko: nit: do we have style-guide for @flow syntax? I am seeing that you have a space between the `nextProps` and `: ` but not after `nextState`

Similar inconsistency is when you define a type for function return value.

packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -175,64 +216,107 @@ class ObjectInspector extends Component {
     return item.path;
   }
 
-  setExpanded(item: Node, expand: boolean) {
```
janodvarko: The `expand` method looks like a key to understand the on-demand fetching logic. An explanatory comment (in-front of the method) would help to understand the code.

packages/devtools-reps/src/object-inspector/index.js
nchevobbe: There is no style guides, I try to keep it consistent with the code (and missed this one)
It would be nice if we could enforce this with eslint

packages/devtools-reps/src/object-inspector/index.js
nchevobbe: Done

nchevobbe:
> Could we use OI for the DOM panel?
I think we could at some point. I guess there are some differences though, like how we have a `[defaultProperties]` node on the window (but we might want it in Dom panel), or the soon to be buckets for non-indexed properties (which would impact big objects).
In the end it might be okay for the DOM panel, and I already though about it :) 
The only thing is I think the DOM panel uses a Tree component smarter than the one we have use in ObjectInspector (uses some virtualization, and we don't in OI).

> How hard it would be to consume & display JS objects directly (not clients & grips). I am thinking about JSON Viewer that could be built on top of OI as well.

Yes, I thought about that too. At the moment, we only say that an object is expandable if it is a grip. So we'd have to remove that. Then we have the notion of actors in different places that wouldn't make sense for the JSON Viewer. IF we fix those, then we could pass it an ObjectClient that would directly query the object. I think this would be a bit weird and risky (we'll have to think of the JSON Viewer each time we do things in the ObjectInspector), and would not bring many benefits for the JSON Viewer (we don't really care of "loading" properties, entries, prototype, …), except for bucketing I guess. But maybe you have other ideas for the JSON Viewer that would benefit having the OI as the ground component

nchevobbe:
CI is failing because lerna makes the OI use the devtools-components Tree that is in master instead of the 0.0.2 version of the npm package.
I will publish a 0.0.3 version of the package soon and use it in devtools-reps so we don't have those errors anymore.

PR #636, opened 2017-08-31
Adapt Reps test app to ObjectInspector changes

nchevobbe:
This patch is based on #634,
so you can only look at https://github.com/devtools-html/devtools-core/commit/2370562c843ab60de228cbbcec51724284b33b95


nchevobbe:
Merged as part of https://github.com/devtools-html/devtools-core/pull/634

Issue #641, opened 2017-09-01
[themes] Remove CSS namespaces


jasonLaster:
We currently namespace our CSS themes, this is making it hard to keep the themes in sync.

We should remove the prefixes and so something else:

1. write a small theme loader (we can't use MCs because it's priviledged)
2. do some post-processing with post-css

PR #642, opened 2017-09-01
Remove theme namespaces

jasonLaster:
Associated Issue: #641 

### Summary of Changes

We would like to remove the theme namespacing because it is making it harder to sync. Eventually it would be nice to pull these styles from searchfox, but we don't need to do that now. 

Next steps:

1. get this working
2. see what regressions exist between themes and rectify that. I'm happy to add a script that does the diff so it's easy to see what's different and fix it.

### DIFF
--- a/packages/devtools-launchpad/package.json
+++ b/packages/devtools-launchpad/package.json
@@ -77,6 +77,7 @@
     "node-static": "^0.7.7",
     "postcss": "^6.0.6",
     "postcss-bidirection": "^2.4.0",
+    "postcss-class-namespace": "^0.1.0",
     "postcss-loader": "^2.0.6",
     "properties-parser": "^0.3.1",
     "ps-node": "^0.1.4",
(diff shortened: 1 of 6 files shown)


packages/devtools-launchpad/package.json
```diff
@@ -77,7 +77,9 @@
     "node-static": "^0.7.7",
     "postcss": "^6.0.6",
     "postcss-bidirection": "^2.4.0",
+    "postcss-class-namespace": "^0.1.0",
     "postcss-loader": "^2.0.6",
+    "postcss-namespace": "^1.1.1",
```
jasonLaster: i will drop this soon

packages/devtools-launchpad/package.json
```diff
@@ -77,7 +77,9 @@
     "node-static": "^0.7.7",
     "postcss": "^6.0.6",
     "postcss-bidirection": "^2.4.0",
+    "postcss-class-namespace": "^0.1.0",
```
jasonLaster: https://github.com/jasonLaster/postcss-class-namespace

might need to support some extra rules: like root and moz-selection... but it learned pretty quickly

PR #643, opened 2017-09-02
Extract source map utils

jasonLaster:
### Summary of Changes

One of the things I find confusing when thinking about the source map worker is understanding the different tasks it tries to perform. In this PR I tried separating out some of the tasks a bit and moving some of the logic into utils:

1. handling the source map requests cache
2. fetching a source map
3. calculating original and generated locations

The new directory structure has a `utils` directory for fetching source maps, remapping wasm, and other utilities. We separated the parser tasks in a similar way a couple months ago and found it easier to work on the different APIs at that point.

I'm curious to see what others think of having smaller modules. If we go down this route, it would be nice to have @yurydelendik's scope locations work in its own module as well as it is fairly complicated and not that related to the other location methods.


packages/devtools-source-map/src/utils/fetchSourceMap.js
```diff
@@ -0,0 +1,102 @@
...
+
+  // Fire off the request, set it in the cache, and return it.
+  const req = _resolveAndFetch(generatedSource);
+  // Make sure the cached promise does not reject, because we only
+  // want to report the error once.
+  setSourceMap(generatedSource.id, req.catch(() => null));
+  return req;
+}
+
+module.exports = { fetchSourceMap }
```
jasonLaster: I like that in the case of fetch, it's clear that we only expose one function. Previously, we were relying on `_` convention, but it wasn't clear if other APIs were inappropriately reaching in and making calls to the other fetching utils

packages/devtools-source-map/src/utils/sourceMapRequests.js
```diff
@@ -0,0 +1,22 @@
...
+
+function setSourceMap(generatedId, request) {
+  sourceMapRequests.set(generatedId, request);
+}
+
+module.exports = {
+  clearSourceMaps,
+  getSourceMap,
+  setSourceMap
+};
```
jasonLaster: I like having the state wrapped in a module with a clear API. It gives me some more confidence that we're not manipulating the map in unintended ways. There were some other call-sites that i switched to `set` and `get` with this move.

packages/devtools-source-map/src/utils/fetchSourceMap.js
```diff
@@ -0,0 +1,102 @@
+// @flow
```
nchevobbe: Shouldn't it have the license header ?

packages/devtools-source-map/src/utils/fetchSourceMap.js
```diff
@@ -0,0 +1,102 @@
...
+  Location, Source, SourceScope, MappedScopeBindings
+} from "debugger-html";
+
+
+/**
+ * Sets the source map's sourceRoot to be relative to the source map url.
+ * @memberof utils/source-map-worker
+ * @static
+ */
+function _setSourceMapRoot(sourceMap, absSourceMapURL, source) {
```
nchevobbe: no flow typing ?

packages/devtools-source-map/src/utils/fetchSourceMap.js
```diff
@@ -0,0 +1,102 @@
...
+  return sourceMap.sourceRoot;
+}
+
+function _resolveSourceMapURL(source: Source) {
+  const { url = "", sourceMapURL = "" } = source;
+  if (path.isURL(sourceMapURL) || url == "") {
+    // If it's already a full URL or the source doesn't have a URL,
+    // don't resolve anything.
+    return sourceMapURL;
+  } else if (path.isAbsolute(sourceMapURL)) {
```
nchevobbe: nit: Turn it on a separate if block to match devtools style ?

packages/devtools-source-map/src/utils/fetchSourceMap.js
```diff
@@ -0,0 +1,102 @@
...
+  if (existingRequest) {
+    // If it has already been requested, return the request. Make sure
+    // to do this even if sourcemapping is turned off, because
+    // pretty-printing uses sourcemaps.
+    //
+    // An important behavior here is that if it's in the middle of
+    // requesting it, all subsequent calls will block on the initial
+    // request.
+    return existingRequest;
+  } else if (!generatedSource.sourceMapURL) {
```
nchevobbe: nit: have this as a separate if block

tromey:
> One of the things I find confusing when thinking about the source map worker is understanding the different tasks it tries to perform.

FWIW the main problem I've had is not understanding the hidden requirements of the API.  I'll file a bug about writing jsdoc for this module.

PR #653, opened 2017-09-07
whitelist ini-parser

jasonLaster:
bump license checker to include a dev dependency


jasonLaster:
landing as it seems straight forward. If there are any objections we can roll it back

PR #654, opened 2017-09-07
Fix sidebar svg usage

jasonLaster:
This fixes our use of SVG to be consistent as well. Note that this is the purpose of `svg-inline-react`
`
https://github.com/sairion/svg-inline-react


packages/devtools-launchpad/src/components/Sidebar.js
```diff
@@ -22,14 +23,16 @@ const Sidebar = React.createClass({
...
   renderTitle(title) {
-    return dom.div({className: "title-wrapper"},
+    return dom.div(
+      { className: "title-wrapper" },
       dom.h1({}, title),
-      dom.div({className: "launchpad-container"},
-        dom.div({
+      dom.div(
+        { className: "launchpad-container" },
+        createElement(InlineSVG, {
```
nchevobbe: 👍

PR #656, opened 2017-09-08
Implements getScope parsing in the devtools-map-bindings

yurydelendik:
Associated Issue: #624

### Summary of Changes

* Extracts code from https://github.com/devtools-html/debugger.html/pull/3852
* Adds babel-traverse for testing

### DIFF
--- a/packages/devtools-map-bindings/src/parser.js
+++ b/packages/devtools-map-bindings/src/parser.js
@@ -1,0 +1,301 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// @flow
+
+import type { Source, SourceId, Location, SourceScope } from "debugger-html";
+import type { NodePath, Node, Location as BabelLocation } from "babel-traverse";
+
+export type ParsedScopeNamesReferences = {
+  [name: string]: BabelLocation[]
+};
+
+export type ParsedScope = {
+  start: Location,
+  end: Location,
+  type: string,
+  bindings: ParsedScopeNamesReferences,
+  children: ?(ParsedScope[])
+};
+
+export type ParseJSScopeVisitor = {
+  traverseVisitor: any,
+  toParsedScopes: () => ParsedScope[],
+};
+
+type TempScopeNameReferences = {
+  type: string,
+  refs: BabelLocation[]
+};
+
+type TempScopeNamesReferences = {
+  [name: string]: TempScopeNameReferences
+};
+
+type TempScope = {
+  type: string,
+  parent: TempScope | null,
+  children: Array<TempScope>,
+  loc: BabelLocation,
+  names: TempScopeNamesReferences
+};
+
+function createTempScope(
+  type: string,
+  parent: TempScope | null,
+  loc: BabelLocation
+): TempScope {
+  const result = {
+    type,
+    parent,
+    children: [],
+    loc: loc,
+    names: Object.create(null)
+  };
+  if (parent) {
+    parent.children.push(result);
+  }
+  return result;
+}
+
+function isNode(node?: Node, type: string): boolean {
+  return node ? node.type === type : false;
+}
+
+function getFunctionScope(scope: TempScope): TempScope {
+  let s = scope;
+  while (s.type !== "Function" && s.type !== "Script") {
+    if (!s.parent) {
+      return s;
+    }
+    s = s.parent;
+  }
+  return s;
+}
+
+function fromBabelLocation(
+  location: BabelLocation,
+  sourceId: SourceId
+): Location {
+  return {
+    sourceId,
+    line: location.line,
+    column: location.column
+  };
+}
+
+function parseDeclarator(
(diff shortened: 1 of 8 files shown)


packages/devtools-map-bindings/src/tests/parser.js
```diff
@@ -0,0 +1,40 @@
...
+const { createParseJSScopeVisitor, findScopes } = require("../parser");
+const { getSource, traverseAst } = require("./helpers");
+
+import type { Source, Location, SourceScope } from "debugger-html";
+
+function parseScopes(location, source) {
+  const { sourceId } = location;
+  const visitor = createParseJSScopeVisitor(sourceId);
+  traverseAst(source, visitor.traverseVisitor);
+  const parsedScopes = visitor.toParsedScopes();
```
jasonLaster: I like how you pull in the test `traverseAst` helper

packages/devtools-map-bindings/src/parser.js
```diff
@@ -0,0 +1,297 @@
...
+        savedParents.set(path, parent);
+        const scope = createTempScope("Function", parent, location);
+        if (isNode(tree.id, "Identifier")) {
+          const functionName = { type: "fn", refs: [] };
+          getFunctionScope(parent).names[tree.id.name] = functionName;
+          scope.names[tree.id.name] = functionName;
+        }
+        tree.params.forEach(param => {
+          parseDeclarator(param, scope, "param");
+        });
```
jasonLaster: nit:
`tree.params.forEach(param => parseDeclarator(param, scope, "param")))`

packages/devtools-map-bindings/src/parser.js
```diff
@@ -0,0 +1,297 @@
...
+    declaratorId.properties.forEach(prop => {
+      parseDeclarator(prop.key, targetScope, type);
+    });
+  } else if (isNode(declaratorId, "ArrayPattern")) {
+    declaratorId.elements.forEach(item => {
+      parseDeclarator(item, targetScope, type);
+    });
+  } else if (isNode(declaratorId, "RestElement")) {
+    parseDeclarator(declaratorId.argument, targetScope, type);
+  }
```
jasonLaster: we'll want some tests for these cases

packages/devtools-map-bindings/src/tests/fixtures/es6.js
```diff
@@ -0,0 +1,6 @@
+let zero = 0;
+function fn(a, b) {
+  for (let i of [1, 2]) {
+    const i = a + i + zero;
```
jasonLaster: it'd be nice if `i` were `index` so that we can have another minifier check

packages/devtools-map-bindings/src/parser.js
```diff
@@ -0,0 +1,297 @@
...
+
+function recordIdentifierPosition(
+  scope: TempScope,
+  name: string,
+  loc: BabelLocation
+) {
+  // Find nearest outer scope with the specifed name and add reference.
+  for (let s = scope; s; s = s.parent) {
+    if (name in s.names) {
+      s.names[name].refs.push(loc.start);
```
jasonLaster: it would be nice if this function avoided mutating `scope` and instead returned the found location:

packages/devtools-map-bindings/src/parser.js
```diff
@@ -0,0 +1,297 @@
...
+          return node.kind === "let" || node.kind === "const";
+        });
+        if (hasLet) {
+          savedParents.set(path, parent);
+          parent = createTempScope("Block", parent, location);
+        }
+        return;
+      }
+      if (path.isVariableDeclaration()) {
+        const hoistAt =
```
jasonLaster: neat. I imagine this might be one of the more interesting considerations because babel is free to add other hoisting rules in the future:

perhaps we want to try looking for the ref:

```js
path.findParent((path) => path.scope.bindings.include(?));
```

NOTE: just guessing based on reading the code, i could be mis-reading this

packages/devtools-map-bindings/src/parser.js
```diff
@@ -0,0 +1,297 @@
...
+        const hasLet = tree.body.some(node => {
+          if (!isNode(node, "VariableDeclaration")) {
+            return false;
+          }
+          return node.kind === "let" || node.kind === "const";
+        });
+        if (hasLet) {
+          savedParents.set(path, parent);
+          parent = createTempScope("Block", parent, location);
+        }
```
jasonLaster: whats this doing?

packages/devtools-map-bindings/src/parser.js
```diff
@@ -0,0 +1,297 @@
...
+        parent = scope;
+        return;
+      }
+      if (path.isBlockStatement()) {
+        const hasLet = tree.body.some(node => {
+          if (!isNode(node, "VariableDeclaration")) {
+            return false;
+          }
+          return node.kind === "let" || node.kind === "const";
+        });
```
jasonLaster: nit: can we move this function up top?

```js
      function hasLet(path) {
         return path.node.body.some(node => {
           if (!isNode(node, "VariableDeclaration")) {
             return false;
           }
           return node.kind === "let" || node.kind === "const";
         });
       }
```

PR #662, opened 2017-09-12
Use arrow and styles from the Tree component

nchevobbe:
This allow us to remove our custom arrow and custom styles.
We add an extra `object-inspector` classname on the tree so we can override CSS properties safely


nchevobbe:
Aaah, gone are the devtools-reps errors in CI :)

PR #663, opened 2017-09-12
Better support for noGrip property

janodvarko:
### Summary of Changes
This is a follow up for issue #480

* Null rep can be used  for both: grips as well as plain JS objects
* Undefined rep can be used  for both: grips as well as plain JS objects

Honza


janodvarko:
Yep, I'll do it, filled https://github.com/devtools-html/devtools-core/issues/664
Honza

PR #666, opened 2017-09-12
Support symmetric locations

jasonLaster:
fixes: https://github.com/devtools-html/debugger.html/issues/3399

### Summary of Changes

We currently have a problem where the user adds a breakpoint on line 5 of the example below at the top of the IF statement and the breakpoint slides to line 8 at the bottom of the IF statement definition.

This happens because when convert the location from original, to generated, and back to original, the location moves from line 5 to line 8. We should support symmetry in mapping where the location does not change if when we map it` toOriginal(toGeneral(oLoc))`. The analogy is 2 inches is 5.08cm and 5.08cm is 2inches. Changing units should not change the values.

```js
function componentWillReceiveProps(nextProps) {
    console.log('start');
    const { selectedSource } = nextProps;

    if (
      nextProps.startPanelSize !== this.props.startPanelSize ||
      nextProps.endPanelSize !== this.props.endPanelSize
    ) {
      this.state.editor.codeMirror.setSize();
    }
	console.log('done');
}
```

#### updates tests 
  * clears cache
  * groups test and attempts to give them names 
  * adds a test helpers file
     * adds formatLocation for nicer logging

#### logic changes

* start search for a original lines at the beginning of the line
* dont convert columns to undefined on the way out. We used to try and help the debugger by converting 0 columns to undefined, turns out this creates some problems w/ symmetry.

### Test Plan

Test going from original -> generated -> original w/ both 0 and undefined columns.

### DIFF
--- a/packages/devtools-source-map/src/source-map.js
+++ b/packages/devtools-source-map/src/source-map.js
@@ -68,9 +68,8 @@ async function getGeneratedLocation(
 
   return {
     sourceId: generatedSourceId,
-    line: line,
-    // Treat 0 as no column so that line breakpoints work correctly.
-    column: column === 0 ? undefined : column
+    line,
+    column
   };
 }
 
@@ -84,19 +83,19 @@ async function getOriginalLocation(location: Location): Promise<Location> {
     return location;
   }
 
-  const { source: url, line, column } = map.originalPositionFor({
+  const { source: sourceUrl, line, column } = map.originalPositionFor({
     line: location.line,
-    column: location.column == null ? Infinity : location.column
+    column: location.column == null ? 0 : location.column
   });
 
-  if (url == null) {
+  if (sourceUrl == null) {
     // No url means the location didn't map.
     return location;
   }
 
   return {
-    sourceId: generatedToOriginalId(location.sourceId, url),
-    sourceUrl: url,
+    sourceId: generatedToOriginalId(location.sourceId, sourceUrl),
+    sourceUrl,
     line,
     column
   };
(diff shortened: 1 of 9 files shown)


packages/devtools-source-map/src/source-map.js
```diff
@@ -68,9 +68,8 @@ async function getGeneratedLocation(
 
   return {
     sourceId: generatedSourceId,
-    line: line,
-    // Treat 0 as no column so that line breakpoints work correctly.
-    column: column === 0 ? undefined : column
+    line,
+    column
```
tromey: IIUC, this change and the corresponding change in `getOriginalLocation` are the critical bits of this patch, and the rest is updating tests, etc.

If that's correct then this patch seems good to me.  It's hard to know offhand if this will have any negative fallout, but I think less weirdness at this layer is a better way generally.

I don't know if it matters here but it's worth noting that the mapping isn't necessarily 1:1.  For example, if a sourcemap generator did inlining, then you could see multiple generated locations with the same source location.  I think sourcemaps can't represent the inverse, but conceptually you could have, say, identical code folding and then a single generated location would have multiple original locations.  So anyway the most that can be said is that when mapping from generated to original to generated, you'll be guaranteed that your initial value will appear in the result set.

jasonLaster: > then you could see multiple generated locations with the same source location

Yeah, i was thinking that too. I feel like this reduces the number of divergences, but it there could always be an onto relationship where many many G maps to one O location.

PR #667, opened 2017-09-13
Sync CSS themes file from mozilla-central

nchevobbe:
This PR syncs CSS themes file from mozilla-central, which now includes photon variables and styles.

I'm not sure of what I'm doing here, warn me if something is wrong :)


nchevobbe:
closing this since better work is happening in https://github.com/devtools-html/devtools-core/pull/719

PR #669, opened 2017-09-13
Render Reps as TINY mode when the node is expanded.

nchevobbe:
Doing this makes the tree less cluttered.

Before: 

sep-13-2017 18-16-08

After: 

sep-13-2017 15-47-07


jasonLaster:
@nchevobbe can you share the before version too :)

nchevobbe:
> @nchevobbe can you share the before version too :)

Yes, I added it in the first comment

> Snapshots look awesome!

yes, you convinced me on that :)

PR #672, opened 2017-09-13
bump source map worker

jasonLaster:
this bumps the source map worker. 

Would it be alright if i updated mc when I update the debugger? the two should be in sync

### DIFF
--- a/packages/devtools-source-map/package.json
+++ b/packages/devtools-source-map/package.json
@@ -1,6 +1,6 @@
 {
   "name": "devtools-source-map",
-  "version": "0.12.0",
+  "version": "0.13.0",
   "description": "DevTools utilities for working with source maps",
   "license": "MPL-2.0",
   "author": "J. Ryan Stinnett <jryans@gmail.com>",


tromey:
> Would it be alright if i updated mc when I update the debugger? the two should be in sync

It's fine with me.  I don't think there's a particular need to land every release into M-C; just ones that we need or ones where we think there are issues to solve first, where the original author is better situated to solve them.

jasonLaster:
Agreed on both accounts.

juliandescottes:
>> Would it be alright if i updated mc when I update the debugger? the two should be in sync

> It's fine with me. I don't think there's a particular need to land every release into M-C; just ones that we need or ones where we think there are issues to solve first, where the original author is better situated to solve them.

It's a tough call. It puts more pressure on releasing debugger bundles. 

The goal of landing a release in m-c is to actually validate the release: check that it doesn't break any mochitest etc... 

Let's say we skip several sourcemaps releases without checking in m-c. Then we try to update the debugger bundle in m-c, we also need to update the sourcemaps bundle. What if this breaks stuff in m-c? Then you need to go back and fix sourcemaps, update the debugger again etc... 

Our goal right now should be to make the bundle updating process easier (and automate it!). I think allowing sourcemaps (or reps for that matter) releases to be published on npm without having been validated on m-c would go in the opposite direction.

And as it just happens, the latest update of the debugger forgot to include the new sourcemaps bundle :) 
https://bugzilla.mozilla.org/show_bug.cgi?id=1399673

PR #673, opened 2017-09-14
Update snapshot because markup changed in another PR

nchevobbe:


juliandescottes:
Thanks for fixing the snapshot!

PR #678, opened 2017-09-15
Wrap provided objectClient function calls in try/catch blocks.

nchevobbe:
Since we don't know what the ObjectClient we get looks like, it might happens that it doesn't have
the method we try to call (e.g. connecting to an older server, you will miss the enumSymbols function).
Wrapping everything in try/catch blocks mitigate the risk of crashing the ObjectInspector (in the beforementioned case, everything will be fine except we wouldn't get symbol properties, which seems acceptable).

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/component/create-object-client.js
+++ b/packages/devtools-reps/src/object-inspector/tests/component/create-object-client.js
@@ -81,4 +81,16 @@ describe("createObjectClient", () => {
     }));
     expect(createObjectClient.mock.calls[0][0]).toBe(grip);
   });
+
+  it("does not fail if the ObjectClient does not have the expected functions", () => {
+    const stub = gripRepStubs.get("testMoreThanMaxProps");
+    const root = createNode(null, "root", "/", {value: stub});
+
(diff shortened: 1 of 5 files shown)


nchevobbe:
I'm not sure about the `console.warn` thing though. I ended up adding them because the consumer need to know if something is wrong with the `createObjectClient` prop it provides. But it does spam the terminal a bit when running `yarn test` (at least for the test where we ensure we don't throw when the functions do not exist on the ObjectClient)

nchevobbe:
Originally I was testing that if the function existed on the ObjectClient, and I was early returning if not, the warning was more to indicate something was missing.
But now since I'm in a try/catch (did that because the server can throws), it makes more sense to use `console.error`.
I'll make the change before merging

tromey:
> connecting to an older server, you will miss the enumSymbols function

In the past, we would add traits to the actor's form so that clients could safely be aware of new functionality.  I'm not sure if that applies in this case but I thought I'd mention it.

jasonLaster:
yeah, i think a traits property or checking if the feature exists is a better pattern for feature detection.

PR #689, opened 2017-09-18
Unvirtualize the Tree component.

nchevobbe:
We were having an `onScroll` event that was making the console lag
if you had a large number of ObjectInspector in it.
None of the consumer of this Tree component take advantage of this
so let's remove it.
We might want to reintroduce some virtualization later, but maybe using
the IntersectionObserver instead, since we can have variable height items.
Also this PR breaks keyboard navigation, as if you navigate to an
off-canvas item you won't be scrolled. I think this was already broken for
variable height items though. This should be fixed in another PR
with keeping in mind that this tree is not meant to be used as the single
component of a container. At the moment, every consumer disable the
keyboard navigation, so we don't put anyone at risk.

### DIFF
--- a/packages/devtools-components/src/tree.js
+++ b/packages/devtools-components/src/tree.js
@@ -333,10 +333,6 @@ const Tree = createClass({
     //     isExpanded: item => item.expanded,
     isExpanded: PropTypes.func.isRequired,
 
-    // The height of an item in the tree including margin and padding, in
-    // pixels.
-    itemHeight: PropTypes.number.isRequired,
-
     // Optional props
 
     // The currently focused item, if any such item exists.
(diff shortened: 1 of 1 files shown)


packages/devtools-components/src/tree.js
```diff
@@ -554,22 +532,23 @@ const Tree = createClass({
...
-      // visible. Instead, give priority to the start of the item by checking its
-      // position first, and then using an "else if", rather than a separate "if",
-      // for the end position.
-      if (this.state.scroll > itemStartPosition) {
-        this.refs.tree.scrollTop = itemStartPosition;
-      } else if ((this.state.scroll + this.state.height) < itemEndPosition) {
-        this.refs.tree.scrollTop = itemEndPosition - this.state.height;
-      }
-    }
+    // TODO: Revisit how we should handle focus without having fixed item height.
```
jasonLaster: hmm, so _focus is important in two cases: 

1. revealing a node and scrolling to it
2. when the user arrows down repeatedly and keeping the focus in view

will these two use cases be supported?

nchevobbe: With this PR, in this state, no.
Do you think we should handle this here or as part of https://github.com/devtools-html/devtools-core/issues/690 ?
This PR fixes a pretty important perf regression on the console, and I'd like to get it pretty soon.

jasonLaster: Lets land this and then fix #690 before the debugger pulls in the patch.

PR #693, opened 2017-09-18
Guard nodeHasAllEntriesInPreview

### LINKED ISSUE #681: Expanding a cyclic referenced map throws
Steps to reproduce:
1. Start the reps test app (or open the console in nightly)
2. Evaluate 
```js
var map1 = new Map([]); 
map1.set(map1, 1); map1;
```
3. Expand the object, then the `entries` node, then the `key` one

Expected results:
It shows child nodes of the `key` node (which should be `Map → 1`)

Actual results:
Nothing happens.
If you look at the logs, there is an error: 
```
TypeError: items is undefined
nodeHasAllEntriesInPreview utils/node.js:222:2
```
which points to https://github.com/devtools-html/devtools-core/blob/b9616604b90565680292f4ff3bfe2565ad415b17/packages/devtools-reps/src/object-inspector/utils/node.js#L222 .
When the exception throws, here is what we have in `preview`:
```js
{
  "kind": "MapLike",
  "size": 1
}
```
So no `entries` neither `items`. In that case, we should directly return false.

nchevobbe:
Fixes #681.

Some Map grip can have a preview object which contains only a size property.
In such case, the ObjectInspector was throwing in `nodeHasAllEntriesInPreview`,
since it expected to have at least an entries or items property.
We now return early on the function if the preview object does not have an
entries nor an item property.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/utils/node-has-all-entries-in-preview.js
+++ b/packages/devtools-reps/src/object-inspector/tests/utils/node-has-all-entries-in-preview.js
@@ -28,13 +28,13 @@ describe("nodeHasEntries", () => {
     ).toBe(true);
   });
 
-  it("returns false for a Map with nore than 10 items", () => {
+  it("returns false for a Map with more than 10 items", () => {
     expect(nodeHasAllEntriesInPreview(
       createRootNode(gripMapStubs.get("testMoreThanMaxEntries"))
     )).toBe(false);
   });
 
-  it("returns false for a WeakSet with nore than 10 items", () => {
+  it("returns false for a WeakSet with more than 10 items", () => {
     expect(nodeHasAllEntriesInPreview(
       createRootNode(
         gripArrayStubs.get("new WeakSet(document.querySelectorAll('div, button'))")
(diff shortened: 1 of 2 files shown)


packages/devtools-reps/src/object-inspector/tests/utils/node-has-all-entries-in-preview.js
```diff
@@ -12,29 +12,53 @@ const {
...
-
+  });
+  it("returns true for a Set with every entries in the preview", () => {
     expect(nodeHasAllEntriesInPreview(
       createRootNode(gripArrayStubs.get("new Set([1,2,3,4])")))
     ).toBe(true);
   });
 
-  it("returns false when expected", () => {
+  it("returns false for a Map with nore than 10 items", () => {
```
bgrins: 'nore'->'more'

packages/devtools-reps/src/object-inspector/tests/utils/node-has-all-entries-in-preview.js
```diff
@@ -12,29 +12,53 @@ const {
...
     expect(nodeHasAllEntriesInPreview(
       createRootNode(gripMapStubs.get("testMoreThanMaxEntries"))
     )).toBe(false);
+  });
 
+  it("returns false for a WeakSet with nore than 10 items", () => {
```
bgrins: 'nore'->'more'

PR #694, opened 2017-09-18
Don't escape the Unicode replacement character

### LINKED ISSUE #644: Stop escaping � as \ufffd
Enter this to the new console:
```js
new TextDecoder("utf-16").decode(new Uint8Array([0x30]))
```

Result: `"\ufffd"`
My reaction: Huh, what happened? What was this symbol again? I have to search it.

---

Enter this to the old console:
```js
new TextDecoder("utf-16").decode(new Uint8Array([0x30]))
```

Result: `"�"`
My reaction: Oh, I messed with the encoding.

---

Please make the life of JS developers easier and don't escape �. It's a well-know symbol because it usually appears when web pages use the wrong encoding. But its unicode number is much less recognizable.

tromey:
It makes sense to pass through the Unicode replacement character,
because users understand it.  However, the other specials (including
unassigned ones) should be quoted.
Fixes #644

### DIFF
--- a/packages/devtools-reps/src/reps/rep-utils.js
+++ b/packages/devtools-reps/src/reps/rep-utils.js
@@ -49,8 +49,8 @@ const escapeRegexp = new RegExp(
   "\x7f-\x9f" +
   // BOM
   "\ufeff" +
-  // Replacement characters and non-characters.
-  "\ufffc-\uffff" +
+  // Specials, except for the replacement character.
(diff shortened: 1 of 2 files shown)


jasonLaster:
Will this cover this case?

tromey:
I don't think so, because the error rep doesn't use the escaping code that the string rep uses.  I guess that's a second bug.

The string rep should show that "001b" character as `\u001b`:

image

PR #696, opened 2017-09-19
Add an omitLinkHref prop to the String rep.

nchevobbe:
This boolean prop prevents the creation of the href attribute on
generated links in the String rep. It is true by default so
consumer can use the String rep safely. We still want to be
able to generate the href attribute (for the JSON panel for example),
this is why we offer the choice to still create it.

Tests are added to make sure we don't regress.


nchevobbe:
CI errors not related

PR #697, opened 2017-09-19
Use aria-activedescendant properly.

nchevobbe:
This patch adds a proper id attribute to the tree nodes
so that the `aria-activedescendant` attribute links to an
actual element.

It also fixes the use of `wrapper.everyWhere` which except to be returned an assertion.

### DIFF
--- a/packages/devtools-components/src/tests/tree.js
+++ b/packages/devtools-components/src/tests/tree.js
@@ -66,7 +66,7 @@ describe("Tree", () => {
     expect(mountTree()).toBeTruthy();
   });
 
-  it.only("is accessible", () => {
+  it("is accessible", () => {
     const wrapper = mountTree({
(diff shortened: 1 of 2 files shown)


yzen:
This looks good. I would only recommend to actually verifying in tests where aria-activedescendant is checked that it matches the ID of the currently focused tree node.

nchevobbe:
> I would only recommend to actually verifying in tests where aria-activedescendant is checked that it matches the ID of the currently focused tree node.

Good idea, I did it in https://github.com/devtools-html/devtools-core/pull/697/commits/dd933aaf8c63b6b2cf5b2216c6da71c8bb85b378

nchevobbe:
Merging this as yzen approved it

yzen:
👍

PR #698, opened 2017-09-19
Compute sourceRoot when only relative URLs are seen

### LINKED ISSUE #356: [SourceMap] CNN generates a garbled source map
1. go to CNN.com
2. observe the error in the console

the source map worker fails to fetch the map locally, the content [here] looks like [this], and should look like [this]

This is the URL networkRequest fails to fetch: http://cdn.livefyre.com/libs/Livefyre/v1.1.1/builds/325/Livefyre.min.js.map

[this]:http://cdn.livefyre.com/libs/Livefyre/v1.1.1/builds/325/Livefyre.min.js.map

[this]:https://gist.github.com/f8deaa389b1310c96561f1e95e05acbb
[here]:https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-source-map/src/source-map.js#L87

tromey:
Currently _setSourceMapRoot bails if the source map contains all the
sources.  However, this is not valid according to the source map spec;
and the debugger doesn't work properly with the returned URLs.  On the
other hand, some source maps have non-URLs like "webpack:/bootstrap",
which we do want to treat as absolute.  So, this patch works around
the problem by examining the URLs in a mildly lax way.
Fixes #356

Associated Issue: #<issue number>

### Summary of Changes

* change 1
* change 2

### Test Plan

Example test plan:

- [x] Command-P opens the panel
- [x] Clicking “+” opens the panel
- [x] Clicking a source navigates to the source
- [x] Clicking "x" closes the panel

### DIFF
--- a/packages/devtools-source-map/src/tests/source-map.js
+++ b/packages/devtools-source-map/src/tests/source-map.js
@@ -50,6 +50,11 @@ describe("source maps", () => {
       const urls = await setupBundleFixture("noroot");
       expect(urls).toEqual(["http://example.com/heart.js"]);
     });
+
+    test("Non-existing sourceRoot resolution with relative URLs", async () => {
+      const urls = await setupBundleFixture("noroot2");
+      expect(urls).toEqual(["http://example.com/heart.js"]);
+    });
   });
 
   describe("hasMappedSource", async () => {
--- a/packages/devtools-source-map/src/utils/fetchSourceMap.js
+++ b/packages/devtools-source-map/src/utils/fetchSourceMap.js
@@ -11,6 +11,8 @@ const { SourceMapConsumer } = require("source-map");
 
 const path = require("./path");
 
+const URL_ISH = new RegExp("^[a-z]+:/");
+
 import type {
   Location,
   Source,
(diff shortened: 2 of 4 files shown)


packages/devtools-source-map/src/utils/fetchSourceMap.js
```diff
@@ -28,9 +30,12 @@ function _setSourceMapRoot(
...
 ) {
-  // No need to do this fiddling if we won't be fetching any sources over the
-  // wire.
-  if (sourceMap.hasContentsOfAllSources()) {
+  // No need to do this fiddling if we won't be fetching any sources
+  // over the wire.  However, we do still want to if any of the source
+  // URLs are relative.  What's difficult is that we want to pretend
+  // that some non-URLs, like "webpack:/whatever", are actually URLs.
+  if (sourceMap.hasContentsOfAllSources() &&
+      sourceMap.sources.every(sourceName => URL_ISH.test(sourceName))) {
```
jasonLaster: this is minor, but can we move this to a temporary variable so that it is a bit more self documenting

```js
const allURLsAreAbsolute = sourceMap.sources.every(sourceName => URL_ISH.test(sourceName)
```

packages/devtools-source-map/src/tests/source-map.js
```diff
@@ -50,6 +50,11 @@ describe("source maps", () => {
       const urls = await setupBundleFixture("noroot");
       expect(urls).toEqual(["http://example.com/heart.js"]);
     });
+
+    test("Non-existing sourceRoot resolution with relative URLs", async () => {
+      const urls = await setupBundleFixture("noroot2");
+      expect(urls).toEqual(["http://example.com/heart.js"]);
+    });
```
jasonLaster: i'm not really sure what this test is documenting. It might be nice to see the webpack example in a test as well.

tromey: The other tests cover webpack -- I initially wrote my patch a different way, which is how I knew that I needed to do this regexp thing.

Do you want some other change here as well?

jasonLaster: i see. that's okay.

PR #702, opened 2017-09-21
Bump up devtools-components to 0.0.5

nchevobbe:
package version update


nchevobbe:
Failing on license-check, we can safely merge

PR #709, opened 2017-09-24
Add a script for checking for changes

jasonLaster:
We should be keeping some devtools modules in sync w/ mc. This script checks to see if components are in sync.
 
 
 ---
 

---


nchevobbe:
what's the status of this @jasonLaster ?

jasonLaster:
lets close it :)

PR #718, opened 2017-09-26
Bump React

jasonLaster:
We're bumping to 15.6.2 so we can get the React.Perf goodness in MC. Lets do the same thing here too


jasonLaster:
@nchevobbe oh yeah, nice catch

PR #720, opened 2017-09-26
Load indexed and non-indexed properties separately only for arrays.

### LINKED ISSUE #707: ObjectInspector fails to retrieve object properties in object with `length` property
Steps to reproduce: 
1. Open the console or the Reps repl
2. Evaluate `({a: 1, length: 2})`
3. Expand the result object

Expected results:
The object renders as : 
```
▼ {…}
    a: 1
    length: 2
   ▶︎ __proto__: Object { … }
```

Actual results:
The object renders as : 
```
▼ {…}
   ▶︎ __proto__: Object { … }
```
i.e. without any properties

---

This is because we call the `enumProperties` function on the server with the `ignoreIndexedProperties` options, which ends up here: http://searchfox.org/mozilla-central/source/devtools/server/actors/object.js#902-903,920-921
Here the server strips every properties whose index is < `length`. In that case, since we only have 2 properties, they are both stripped from the array, so we get nothing.

The variable view did not pass the `ignoreIndexedProperties` if the object [was not an array](http://searchfox.org/mozilla-central/source/devtools/client/shared/widgets/VariablesViewController.jsm#258-259,276,292).

nchevobbe:
Fixes #707

This patch makes the ObjectClient use enumProperties options for ArrayLike objects since it may cause a bug in object with a length property.

Hopefully a better fix will happen in  https://bugzilla.mozilla.org/show_bug.cgi?id=1403065  so we'll be able to revert this change.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/index.js
+++ b/packages/devtools-reps/src/object-inspector/index.js
@@ -43,6 +43,7 @@ const {
   nodeIsPrototype,
   nodeIsSetter,
   nodeIsWindow,
+  shouldLoadAllItemProperties,
   shouldLoadItemEntries,
   shouldLoadItemIndexedProperties,
   shouldLoadItemNonIndexedProperties,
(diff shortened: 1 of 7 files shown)


nchevobbe:
closing for now since a fix was made server-side instead

PR #723, opened 2017-09-27
Update to source-map 0.6.0

tromey:
Update to source-map 0.6.0.  This allows removing some code, by passing
the source map URL directly to SourceMapConsumer; this also fixes some
corner cases in source URL resolution.


packages/devtools-source-map/src/utils/fetchSourceMap.js
```diff
@@ -9,95 +9,41 @@ const { getSourceMap, setSourceMap } = require("./sourceMapRequests");
...
+    // If the source doesn't have a URL, don't resolve anything.
+    return { sourceMapURL, baseURL: sourceMapURL };
   }
 
-  if (path.isAbsolute(sourceMapURL)) {
-    // If it's an absolute path, it should be resolved relative to the
-    // host of the source.
-    const { protocol = "", host = "" } = new URL(url);
-    return `${protocol}//${host}${sourceMapURL}`;
+  const resolvedURL = new URL(sourceMapURL, url);
```
nchevobbe: Should this be wrapped in a try/catch block ? `new URL` will throw if `url` is not a url (but maybe this can't happen here ?)

packages/devtools-source-map/src/tests/source-map.js
```diff
@@ -33,7 +33,7 @@ describe("source maps", () => {
     test("source with a url", async () => {
       const urls = await setupBundleFixture("bundle");
       expect(urls).toEqual([
-        "webpack:/webpack/bootstrap 4ef8c7ec7c1df790781e",
+        "webpack:///webpack/bootstrap 4ef8c7ec7c1df790781e",
```
nchevobbe: what explains this change ?

tromey: The previous URL here was a bug in source-map, which is now fixed.  See https://github.com/mozilla/source-map/commit/d21b0e1b6e4d6efe7c39cf1607966c417d916be7

packages/devtools-source-map/src/utils/fetchSourceMap.js
tromey: I see now this can be simplified a little.

However, if this doesn't resolve to a URL one way or another, then I think we want to fail -- because it means the caller passed in something wrong.

Issue #726, opened 2017-09-28
Launchpad: Pass custom properties to renderRoot


janodvarko:
There is currently no way to pass custom props into my App() component that represents the root of the tool rendered on top of the Launchpad.

The renderProps could look like as follows, check out the additional `props` argument.

`function renderRoot(_React, _ReactDOM, component, _store, props)`

Consequently we can pass the props into `createElement` as follows:

```
_ReactDOM.render(
  createElement(Provider, { store: _store }, createElement(component, props)),
  root
);
```
Honza

PR #727, opened 2017-09-28
Pass custom properties to root App component

janodvarko:
Associated Issue: #726

### DIFF
--- a/packages/devtools-launchpad/src/index.js
+++ b/packages/devtools-launchpad/src/index.js
@@ -87,7 +87,7 @@ async function initApp() {
   return { store, actions, LaunchpadApp };
 }
 
-function renderRoot(_React, _ReactDOM, component, _store) {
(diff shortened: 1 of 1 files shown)


packages/devtools-launchpad/src/index.js
```diff
@@ -87,7 +87,7 @@ async function initApp() {
   return { store, actions, LaunchpadApp };
 }
 
-function renderRoot(_React, _ReactDOM, component, _store) {
+function renderRoot(_React, _ReactDOM, component, _store, props) {
```
nchevobbe: maybe default to an empty object ? not sure if it is risky to pass undefined to createElement though.

janodvarko: @nchevobbe we have been passing undefined to `createElement` till now, no? Btw. I tested that and works for me.

Honza

nchevobbe: true  !

PR #729, opened 2017-09-28
Bump source maps

jasonLaster:
bumping the source maps lib to include an up to date version of mappings

### DIFF
--- a/packages/devtools-source-map/package.json
+++ b/packages/devtools-source-map/package.json
@@ -1,6 +1,6 @@
 {
   "name": "devtools-source-map",
-  "version": "0.14.0",
+  "version": "0.14.1",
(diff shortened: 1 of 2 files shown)


tromey:
It's possible you'll run into the problem that caused https://github.com/mozilla/source-map/pull/289

jasonLaster:
i don't follow...

tromey:
Sorry - just that devtools-source-map is slightly broken because source-map is slightly broken; and until that PR lands you may not want to pull a new devtools-source-map into anything else.

tromey:
Once that PR lands I plan to rev source-map, then devtools-source-map; then land it all in M-C.  Hopefully tomorrow.

PR #736, opened 2017-10-03
Add a 30,000 items tree story

nchevobbe:
Huge tree so we can sort out performances


nchevobbe:
Merging without review since it's only a story

PR #745, opened 2017-10-05
Migrate ObjectInspector state to Redux

nchevobbe:
~~This patch allow us to use an object inspectors handler to persist the state of `ObjectInspector`s which might be unmounted and re-mounted (e.g. when the debugger is paused, and resizing it shifts the layout from horizontal to vertical, or the other way around)~~
~~Note that this is blocking the switch to React 16 in devtools because the debugger uses an old version of Reps.~~

This does not handle state persistance anymore since we'd better handle things in the debugger in a different way. But this switches the ObjectInspector to Redux.

Which means tests should be migrated as well


nchevobbe:
@codehag since you offer some help on that, could you try to use this in the debugger and tell me if it works as intended (e.g. for stepping)

nchevobbe:
@codehag this is what I was suggesting on Friday.

The consumer code would look like this : 
https://github.com/devtools-html/devtools-core/pull/745/files#diff-729e66e66ba50db07bd83fef1fdafad3R103

Basically, creating an `ObjectInspectorsHandler.create(id, props)` instead of directly using `<ObjectInspector ...props>`. (Note that calling `<ObjectInspector>` would still work, but the state wouldn't persist).

Then, when you don't need to have the state of the ObjectInspector anymore, you can call `ObjectInspectorHandlers.clear()` (see https://github.com/devtools-html/devtools-core/pull/745/files#diff-f626d9e22cd2cd18927c8fa95d606060R32), or `ObjectInspectorHandlers.delete(id)` if you want fined grained control over the cache.

The naming might be terrible here, `ObjectInspectorsHandler` could be `ObjectInspectorPersister` or something better :)

Do tell me how it would feel to use this in debugger-html


cc @rpl

jasonLaster:
@nchevobbe it feels weird to have many stores in the persistor cache. 

Is it possible to have one store and add it to the application store at bootstrap? e.g.

```diff
diff --git a/src/utils/bootstrap.js b/src/utils/bootstrap.js
index 281b078..b6082bd 100644
--- a/src/utils/bootstrap.js
+++ b/src/utils/bootstrap.js
@@ -35,7 +35,9 @@ export function bootstrapStore(client, { services, toolboxActions }) {
     }
   });

-  const store = createStore(combineReducers(reducers));
+  const store = createStore(
+    combineReducers([...reducers, objectInspectorStore])
+  );
   store.subscribe(() => updatePrefs(store.getState()));

   const actions = bindActionCreators(
```

nchevobbe:
yeah, i was thinking about it this morning, let me see what can be done.
In your case, how would we create an object inspector ? dispatching an action ? Or like previously, except we would pass to the object inspector the persisted (or initial state) ? But i don't see how this could be done

nchevobbe:
@jasonLaster would you mind giving this a look ?
This needs test, but I want to see how it looks to you first.
Also, let's out aside discussions about how to expose actions to consumers for now, and just focus on reduxifying the OI.
And last, do not rely on previous comments on this OR, the store persistence was dropped, we'll figure out if we need it in the future.

Thanks !

nchevobbe:
Sorry for the nudge, but I'd like to end this soon-ish so I can tackle other ObjectInspector related issues

jasonLaster:
@nchevobbe happy to look! sorry didn't realize this was ready

packages/devtools-reps/src/object-inspector/reducer.js
```diff
@@ -0,0 +1,70 @@
...
+        } else if (typeof accumulator[k] === "object") {
+          accumulator[k] = Object.assign({}, accumulator[k], v);
+        }
+      } else {
+        accumulator[k] = v;
+      }
+    });
+    return accumulator;
+  }, {});
+}
```
jasonLaster: This is a bit complex...

- Is "from" being overridden a problem?
- are there ways to simplify this by focusing on some types of keys ownSymbols, ownProperties, etc...

nchevobbe:
Tests added (and couple of fixes in the component to make them pass)

packages/devtools-reps/src/object-inspector/reducer.js
nchevobbe: fixed

nchevobbe:
@jasonLaster do you mind giving a final look ?

jasonLaster:
no problem!

nchevobbe:
Thanks for the review Jason, merged !

PR #751, opened 2017-10-11
Support SVGs when creating a bundle

jasonLaster:


jasonLaster:
merging because this only affect builds... We can tweak it for `reps` soon

PR #757, opened 2017-10-13
Update ObjectInspector when root property changes

### LINKED ISSUE #756: ObjectInspector does not update if the root changes
`shouldComponentUpdate` in the ObjectInspector only looks for changes in loaded properties and expanded paths.
But it can happen that the parent wants to change the root itself. So we need to update in this case too.

nchevobbe:
Fix #756
~Need a test~  test added


nchevobbe:
CI failed because yarn died. locally it was fine, let's push

PR #759, opened 2017-10-15
Initial Style Guide

jasonLaster:
### Summary of Changes

* Creates a Style Guide that pulls in the themes
* It would be nice if we didn't need to use the launchpad. The rationale for that is that the launchpad can pull in the mc assets and do the webpack/postcss mappings... but we could move that logic to `devtools-mc-assets` too

Next Steps:

* Add a couple components: buttons, Tree, Accordion, ...
* Add theme switching
* Refactor so it uses CRA in is lighter weight... e.g remove launchpad and the create-store utils...

screen shot 2017-10-15 at 1 04 00 pm


jasonLaster:
screen shot 2017-10-15 at 10 36 39 pm

jasonLaster:
Yep, I still feel like this initial version is pulling in a lot of cruft that we won't need soon, but I'd prefer to land it and iterate on the content for a bit

jasonLaster:
merging now as this only touches the styleguide and we're going to iterate quickly.

PR #762, opened 2017-10-16
Do not use CSS custom property in Tree.

nchevobbe:
Because it takes a lot of time to compute on large tree.
It wasn't only the massive "background as a border", but the indent as well.
Without those 2 scrolling is quite fast.

Using multiple elements for the indent and border works quite well.
Here is 2 profiles, with and without the patch, scrolling through a tree of a thousand node with one expanded root.

In order to get consistent results, I triggered the scrolling with this code : 
```js
var index = 0;
function scrollNext() {
  index = index + 40;
  nodes[index].scrollIntoView();
  if (index < nodes.length - 1) {
    requestAnimationFrame(scrollNext);
  }
}
scrollNext();
```

where `nodes` contains all the nodes of the tree.

Profile without patch:  https://perf-html.io/public/24499d1b8bcd2e67f5071a867b0c3823f2e6c289/calltree/?hiddenThreads=&range=1.4863_2.8093&thread=5&threadOrder=0-2-3-4-5-1&v=2

Profile with patch :  https://perf-html.io/public/6a238fd4b31c250e730970dfe4d488edeef87a31/calltree/?hiddenThreads=&range=1.0600_1.8985&thread=5&threadOrder=0-2-3-4-5-1&v=2

Not perfect, but definitely better.

### DIFF
--- a/packages/devtools-components/src/tests/tree.js
+++ b/packages/devtools-components/src/tests/tree.js
@@ -368,7 +368,7 @@ describe("Tree", () => {
     expect(formatTree(wrapper)).toMatchSnapshot();
 
     getTreeNodes(wrapper).forEach(n => {
-      if ("ABECDMN".split("").includes(n.text())) {
+      if ("ABECDMN".split("").includes(getSanitizedNodeText(n))) {
         expect(n.find(".arrow.expanded").exists()).toBe(true);
       } else {
         expect(n.find(".arrow").exists()).toBe(false);
@@ -446,12 +446,16 @@ function formatTree(wrapper) {
       let arrowStr = "  ";
       if (arrow.exists()) {
         arrowStr = arrow.hasClass("expanded") ? "▼ " : "▶︎ ";
-      } else {
-        arrowStr = "  ";
       }
-      return `${indentStr}${arrowStr}${node.text()}`;
+
+      return `${indentStr}${arrowStr}${getSanitizedNodeText(node)}`;
     })
     .join("\n");
+}
+
+function getSanitizedNodeText(node) {
+  // Stripping off the invisible space used in the indent.
+  return node.text().replace(/^\u200B+/, "");
 }
 
 // Encoding of the following tree/forest:
--- a/packages/devtools-components/src/tree.css
+++ b/packages/devtools-components/src/tree.css
@@ -3,16 +3,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 .tree {
-  --arrow-width: 10px;
-  --arrow-single-margin: 5px;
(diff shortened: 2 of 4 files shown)


nchevobbe:
ping @digitarald so he can check against https://bugzilla.mozilla.org/show_bug.cgi?id=1405052 :)

nchevobbe:
CI is unhappy but only because lerna makes the objectInspector uses the devtools-component from this PR (which change), and not the one from npm.

packages/devtools-components/src/tests/tree.js
```diff
@@ -446,14 +446,18 @@ function formatTree(wrapper) {
...
-      return `${indentStr}${arrowStr}${node.text()}`;
+
+      return `${indentStr}${arrowStr}${getSanitizedNodeText(node)}`;
     })
     .join("\n");
 }
 
+function getSanitizedNodeText(node) {
+  // Stripping off the invisible space used in the indent.
+  return node.text().replace(/\u200B/g, "");
```
digitarald: `/^\u200B+/` would make it more intentional to only strip leading spaces.

nchevobbe: 👍, DONE ✅

packages/devtools-components/src/tree.css
```diff
@@ -36,25 +26,35 @@
...
+  margin-inline-start: 5px;
+  border-inline-start: 1px solid #A2D1FF;
+}
+
 .tree .tree-node[data-expandable="true"] {
   cursor: default;
 }
 
 .tree .tree-node:not(.focused):hover {
-  background-color: var(--tree-node-hover-background-color);
```
janodvarko: Why not using a variable?

packages/devtools-components/src/tree.css
```diff
@@ -36,25 +26,35 @@
...
+  background-color: #F0F9FE;
 }
 
 .tree .tree-node.focused {
-  color: var(--tree-node-focus-color);
-  background-color: var(--tree-node-focus-background-color);
-  --arrow-fill-color: currentColor;
+  color: white;
+  background-color: var(--theme-selection-background, #0a84ff);
```
janodvarko: Why `--theme-selection-background` should be invalid? (so we need the fallback value)

nchevobbe: It is not known when running storybook for devtools-component for example (or when used in the Reps test app)

packages/devtools-components/src/tree.css
nchevobbe: for consistency sake. I nuked all the Component specific css variables, since the style can style be overidden by adding more specific selectors when needed.

Issue #741, opened 2017-10-03
[Mapping] we should map identifiers in an expression


jasonLaster:
We currently map minified variables when the debugger pauses. We should try and map all of the identifies. Here's an example:

#### Minified

```js
function sum(n,u) {
  return n + u
}
```

#### Original source

```js
function sum(first, second) {
  return first + second;
}
```

We can currently map `n` to `first`, but if the user types in the console `first * 2`, we can't map it to `n * 2`. We can solve this by parsing the expression and looking for identifiers to map.

jasonLaster:
here is a [test][t] 

[t]:https://github.com/devtools-html/devtools-core/compare/master...jasonLaster:mapping?expand=1

aleksdao:
@jasonLaster We worked on this as part of the React NYC / FF Devtools Contribution night last week. Unfortunately, I was unable to submit a PR as I was having some issues with the pre-commit hook complaining about `yarn run lint`.

I went ahead and pushed up the code to my fork with `--no-verify`. Hope that's alright and won't cause any issues.

Should I create a PR for feedback? https://github.com/devtools-html/devtools-core/compare/master...aleksdao:feature/ast-mapping-identifiers?diff=unified&expand=1&name=feature%2Fast-mapping-identifiers

jasonLaster:
Thanks @aleksdao! This looks great. Feel free to make a PR. If you're busy, someone else can as well!

aleksdao:
@jasonLaster Great - thanks! I'll make a PR this evening... just didn't want to make a PR and waste everyone's time if it wasn't on the right track :)

jasonLaster:
yep

PR #765, opened 2017-10-18
map identifiers to names

aleksdao:
Associated Issue: #741 

### Summary of Changes

* `mapIdentifiers` parses an expression via AST and maps identifiers to their minified values

### Test Plan

* wrote some assertions to test behavior of the `mapIdentifiers` function

### DIFF
--- a/packages/devtools-map-bindings/src/utils.js
+++ b/packages/devtools-map-bindings/src/utils.js
@@ -12,20 +12,22 @@ function replaceOriginalVariableName(
 ): string {
   const ast = babylon.parse(expression, {
     sourceType: "module",
-    plugins: ["jsx", "flow", "objectRestSpread"],
+    plugins: ["jsx", "flow", "objectRestSpread"]
   });
 
   traverse(ast, {
     Identifier(path) {
       const { node: { name } } = path;
-      const foundScope = generatedScopes.find(({ bindings }) => name in bindings);
+      const foundScope = generatedScopes.find(
+        ({ bindings }) => name in bindings
+      );
       if (foundScope) {
         path.node.name = foundScope.bindings[name];
       }
-    },
+    }
   });
 
-  return generate(ast, { concise: true, compact: true }).code.replace(";", "");
+  return generate(ast, { concise: true, compact: true }).code.replace(/;$/, "");
 }
 
 module.exports = { replaceOriginalVariableName };
--- a/packages/devtools-map-bindings/src/utils.js
+++ b/packages/devtools-map-bindings/src/utils.js
@@ -1,35 +1,31 @@
 // @flow
 
 import type { MappedScopeBindings } from "debugger-html";
+const { default: generate } = require("babel-generator");
+
+const babylon = require("babylon");
+const { default: traverse } = require("babel-traverse");
 
 function replaceOriginalVariableName(
   expression: string,
   generatedScopes: MappedScopeBindings[]
 ): string {
-  // JavaScript indetifier is a complex thing: it can contain '$', '\', ZWJ, any
-  // unicode letter or number. Simplifing that to latin character set for now.
-  // For simplicity, search only first characters of the expression to find an
-  // identifier/variable name.
-  const possibleVarNameRegex = /^([$\w]+)/;
-  const matchedVarName = possibleVarNameRegex.exec(expression);
-  if (!matchedVarName) {
-    return expression;
-  }
+  const ast = babylon.parse(expression, {
+    sourceType: "module",
+    plugins: ["jsx", "flow", "objectRestSpread"],
+  });
 
-  const originalName = matchedVarName[1];
-  // The generatedScopes sorted in inner-to-outer scope order, finding first.
-  const foundScope = generatedScopes.find(
-    ({ bindings }) => originalName in bindings
-  );
-  if (!foundScope) {
-    return expression;
-  }
+  traverse(ast, {
+    Identifier(path) {
+      const { node: { name } } = path;
+      const foundScope = generatedScopes.find(({ bindings }) => name in bindings);
+      if (foundScope) {
+        path.node.name = foundScope.bindings[name];
+      }
+    },
+  });
 
-  // Replace original name (which will be at the beginning of the expression)
-  // with found generated name.
-  const generatedName = foundScope.bindings[originalName];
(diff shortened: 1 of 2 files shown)


jasonLaster:
This is the first time we're using `babel-generator` I think it's a great usecase

packages/devtools-map-bindings/src/utils.js
```diff
@@ -1,35 +1,31 @@
...
+      }
+    },
+  });
 
-  // Replace original name (which will be at the beginning of the expression)
-  // with found generated name.
-  const generatedName = foundScope.bindings[originalName];
-  const expressionWoOriginalName = expression.substring(originalName.length);
-  return `${generatedName}${expressionWoOriginalName}`;
+  return generate(ast, { concise: true, compact: true }).code.replace(";", "");
```
jasonLaster: what does this do: `.replace(";", "");`

packages/devtools-map-bindings/src/utils.js
```diff
@@ -1,35 +1,31 @@
 // @flow
 
 import type { MappedScopeBindings } from "debugger-html";
+const { default: generate } = require("babel-generator");
```
jasonLaster: we usually do `const generate = require("babel-generator").default`

packages/devtools-map-bindings/src/utils.js
aleksdao: `babel-generator` appears append a `;` on the generated output i.e. `"first*2"` becomes `"n*2;"`. To match the test assertions you provided, I removed the trailing `;`; thinking about this some more, there's probably a more standardized approach :)

yurydelendik: there may be a case when user typed ';' as part of the expression, e.g. `str.replace(";", "")`. it's better to use regexp there: `.replace(/;$/, "")`.

yurydelendik:
Can you also add tests for:
- `window.first`
- `{ first: second }`
- `first;`
- `second.replace(";"g,"")`

aleksdao:
@yurydelendik Of course! In your examples above, could you help me understand why we would not map many of those identifiers to their bindings? e.g. `window.first` => `window.first` (and not `window.n`) 

As the `mapIdentifiers` function is currently written, it picks up any `Identifier` node with `name === some_key_in_map_bindings` and replaces the original name with the mapped value. If the new test assertions indeed represent our desired output, I'll change the approach and update the PR. I'm happy to do it -- just wanted to get a confirmation. Thanks!

jasonLaster:
@aleksdao i think the rationale for `window.first => window.first` is that if it is a property then we do not expect it to be different after the mapping.

The best way to check for this is to see if the identifier is a member expression

aleksdao:
@jasonLaster Thanks! I'll continue to work off the assertions, as I think that will be the best way to ensure I'm handling every case. If you have additional inputs and expected outputs, let me know!

Both of the following are member expressions but will be handled differently.

`window.first` => `window.first`
`first[second]` => `n[u]`

jasonLaster:
correct, by the way `foo.first => foo.first` should also work :) window is not special

aleksdao:
@jasonLaster I'm having trouble differentiating between `first` => `n` and `first;` => `first;`. They seem to be represented by a nearly identical AST (the only difference being the `end` property in the `ExpressionStatement`.

`first`: [link](http://astexplorer.net/#/gist/3367194c1717033c14e5dc65a9d8404d/723a1d48b93d25994b56d98d9d987887793bc3ad)
`first;`:
 [link](http://astexplorer.net/#/gist/3367194c1717033c14e5dc65a9d8404d/f4dc678f6fa839a0b35de51b75b303f136c32fa6)

Do you have any suggestions?

jasonLaster:
I’d use that, lets see if it’s parent is an `ExpressionStatement` :slightly_smiling_face:

jasonLaster:
also join our [slack](https://devtools-html-slack.herokuapp.com) :wave:

aleksdao:
@yurydelendik Can you confirm the expected outputs for

`first` -> ?
`first;` -> ?

besides the location, the ASTs for `first` and `first;` are identical. if we need to generate different outputs for them, it would be helpful to get a tip on how to differentiate between the two inputs

thanks!

yurydelendik:
> ASTs for `first` and `first;` are identical.

If these are parsed as expressions the former is invalid one. If babel does not allow code to be parsed as expression, you can enclose that between `(` and `);` (and on the output strip these). As you can notice `(first;);` is invalid JS statement.

jasonLaster:
i pushed a small tweak to the `;` case. and am asking the babel folks for a babely way to do it

packages/devtools-map-bindings/src/utils.js
```diff
@@ -1,35 +1,33 @@
...
-  // JavaScript indetifier is a complex thing: it can contain '$', '\', ZWJ, any
-  // unicode letter or number. Simplifing that to latin character set for now.
-  // For simplicity, search only first characters of the expression to find an
-  // identifier/variable name.
-  const possibleVarNameRegex = /^([$\w]+)/;
-  const matchedVarName = possibleVarNameRegex.exec(expression);
-  if (!matchedVarName) {
-    return expression;
-  }
+  const ast = babylon.parse(expression, {
```
yurydelendik: babylon has the parseExpression method, which returns Expression tree (not Script)

jasonLaster:
yep `parseExpression` is very strict though and needs a scope or other properties.

yurydelendik:
> parseExpression is very strict though and needs a scope or other properties.

Correct, but that's what used by evaluateExpressions->evaluateExpression->hasSyntaxError

jasonLaster:
Hmm, maybe that is what we should change then :)

yurydelendik:
Okay, we need to properly define "expression" term. in JS, expression is defined as http://www.ecma-international.org/ecma-262/6.0/#sec-primary-expression . The console usage assumes that statements can be used (for, function, var, etc). Will we allow statements usage in the watch expressions? (I see it can be useful in the console though)

jasonLaster:
@yurydelendik it's the same API, so yes... and conditional breakpoints benefit from statements too

jasonLaster:
we really should let the server check for syntax errors too

jasonLaster:
lets merge this so it's done and then relax the watch expression API

PR #766, opened 2017-10-20
remove splitter preset

jasonLaster:


jasonLaster:
tested in the debugger... it's small enough for an auto-merge...

PR #770, opened 2017-10-29
Fix warnings introduced by React v15.6

joel-bentley:
Associated Issue: https://github.com/devtools-html/debugger.html/issues/4485

### Summary of Changes

In order to remove warning messages introduced in React v15.6, I removed all instances of `React.createClass`, `React.PropTypes`, and `React.DOM` (which will be deprecated in React v16.) These were replaced using the `create-react-class`, `prop-types`, and `react-dom-factories` packages for minimal changes to rest of code.

This removes some of the warnings on tests but caused others to appear. These `React.PropType` warnings were from the external packages `svg-inline-loader` and `svg-inline-react`.

There might be still some issues with this PR as I wasn't able to run `debugger.html` when linked to these files (blank browser window with error). I'm not sure if that was an issue with how I linked the packages. Does someone else want to try or have any suggestions?

### Test Plan

* I tested the affected packages which have tests using their `yarn test` scripts.
All tests pass. Tests for `devtools-components` and `devtools-reps` still had some React.PropType warnings. These look to be from the `svg-inline-loader` and `svg-inline-react` packages. I was able to remove all warnings by linking custom edited versions of `svg-inline-loader` and `svg-inline-react` with `React.PropTypes` replaced with `prop-types` package.

* I linked packages to `debugger.html` and ran tests with `yarn test` script.
All tests pass, but new warning is introduced for some reason:
```
Warning: Failed prop type: The prop `createObjectClient` is marked as required in `ObjectInspector`, but its value is `undefined`.
          in ObjectInspector
```


nchevobbe:
Thanks a lot for doing this Joel

The warning
> Warning: Failed prop type: The prop `createObjectClient` is marked as required in `ObjectInspector`, but its value is `undefined`.

is expected, since the debugger does not use the new version of the `ObjectInspector` yet (and it explains the blank screen you see).

I will try to create a new bundle and use it in mozilla-central (it is then used by the console, json viewer, dom panel, …) and see if everything is working as expected.

MikeRatcliffe:
I have been asked to close this in favor of pull request #775 where I have also upgraded dependencies and upgraded all components to ES6 classes.

Issue #769, opened 2017-10-27
Duplicate Scope Item


jasonLaster:
We show the same item twice because `jQuery` has two property names `$` and `jQuery` each with the same displayName.

<img width="1137" alt="duplicatescope" src="https://user-images.githubusercontent.com/46655/30774197-5ad73f0c-a044-11e7-8b27-eab9d76ed30e.png">

jasonLaster:
@hughugsy  is on it

nchevobbe:
I think this is fixed now

jasonLaster:
yep

PR #771, opened 2017-10-30
fixed duplicate scope item issue

hughugsy:
Associated Issue: #769 

### Summary of Changes

* Changed getFunctionName function to return propertyName when it's not equal to displayName.

### Test Plan
Catch a breakpoint and scope objects will be displayed.

### DIFF
--- a/packages/devtools-reps/src/reps/function.js
+++ b/packages/devtools-reps/src/reps/function.js
@@ -6,12 +6,7 @@
 const PropTypes = require("prop-types");
 
 // Reps
-const {
-  getGripType,
-  isGrip,
-  cropString,
-  wrapRender,
-} = require("./rep-utils");
+const { getGripType, isGrip, cropString, wrapRender } = require("./rep-utils");
 const { MODE } = require("./constants");
 const Svg = require("../shared/images/Svg");
 
@@ -47,13 +42,13 @@ function FunctionRep(props) {
     });
   }
 
-  return (
-    span({
+  return span(
+    {
       "data-link-actor-id": grip.actor,
       className: "objectBox objectBox-function",
       // Set dir="ltr" to prevent function parentheses from
       // appearing in the wrong direction
-      dir: "ltr",
+      dir: "ltr"
     },
       getTitle(grip, props),
       getFunctionName(grip, props),
@@ -66,31 +61,28 @@ function FunctionRep(props) {
 }
 
 function getTitle(grip, props) {
-  const {
-    mode
-  } = props;
+  const { mode } = props;
 
   if (mode === MODE.TINY && !grip.isGenerator && !grip.isAsync) {
     return null;
   }
 
-  let title = mode === MODE.TINY
-    ? ""
-    : "function ";
+  let title = mode === MODE.TINY ? "" : "function ";
 
   if (grip.isGenerator) {
-    title = mode === MODE.TINY
-      ? "* "
-      : "function* ";
+    title = mode === MODE.TINY ? "* " : "function* ";
   }
 
   if (grip.isAsync) {
     title = "async" + " " + title;
   }
 
-  return span({
-    className: "objectTitle",
-  }, title);
+  return span(
+    {
+      className: "objectTitle"
+    },
+    title
+  );
 }
 
 // Decodes an anonymous naming scheme that
@@ -120,27 +112,33 @@ function getFunctionName(grip, props) {
 
   let { functionName } = props;
   let end = functionName.length - 1;
+  let name;
 
-  props.functionName =
-    functionName.indexOf('"') == 0 && functionName.lastIndexOf('"') == end
+  functionName =
+    functionName.startsWith('"') && functionName.endsWith('"')
       ? functionName.substring(1, end)
       : functionName;
 
-  let name =
-    grip.displayName != undefined && grip.displayName != props.functionName
-      ? props.functionName + ": " + grip.displayName
-      : grip.userDisplayName ||
-        grip.displayName ||
-        grip.name ||
-        props.functionName ||
-        "";
+  if (
+    grip.displayName != undefined &&
+    functionName != undefined &&
+    grip.displayName != functionName
+  ) {
+    name = functionName + ": " + grip.displayName;
+  } else {
+    name =
+      grip.userDisplayName ||
+      grip.displayName ||
+      grip.name ||
+      props.functionName ||
+      "";
+  }
+
   return cropString(name, 100);
 }
 
 function renderParams(props) {
-  const {
-    parameterNames = []
-  } = props;
+  const { parameterNames = [] } = props;
 
   return parameterNames
     .filter(param => param)
@@ -157,10 +155,10 @@ function renderParams(props) {
 function supportsObject(grip, noGrip = false) {
   const type = getGripType(grip, noGrip);
   if (noGrip === true || !isGrip(grip)) {
-    return (type == "function");
+    return type == "function";
   }
 
-  return (type == "Function");
+  return type == "Function";
 }
 
 // Exports from this module
--- a/packages/devtools-reps/src/reps/function.js
+++ b/packages/devtools-reps/src/reps/function.js
@@ -102,22 +102,6 @@ const functionProperty = /([\w\d]+)[\/\.<]*?$/;
 const annonymousProperty = /([\w\d]+)\(\^\)$/;
 
 function getFunctionName(grip, props) {
-  const scenarios = [
-    objectProperty,
-    arrayProperty,
-    functionProperty,
-    annonymousProperty
-  ];
-
-  scenarios.some(reg => {
-    const match = reg.exec(name);
-    if (match) {
-      name = match[1];
-      return true;
-    }
-    return false;
-  });
-
   let { functionName } = props;
   let name;
 
@@ -140,6 +124,21 @@ function getFunctionName(grip, props) {
       props.functionName ||
       "";
   }
+  const scenarios = [
+    objectProperty,
+    arrayProperty,
+    functionProperty,
+    annonymousProperty
+  ];
+
+  scenarios.some(reg => {
+    const match = reg.exec(name);
+    if (match) {
+      name = match[1];
+      return true;
+    }
+    return false;
+  });
 
   return cropString(name, 100);
 }
--- a/packages/devtools-reps/src/reps/function.js
+++ b/packages/devtools-reps/src/reps/function.js
@@ -6,7 +6,12 @@
 const PropTypes = require("prop-types");
 
 // Reps
-const { getGripType, isGrip, cropString, wrapRender } = require("./rep-utils");
+const {
+  getGripType,
+  isGrip,
+  cropString,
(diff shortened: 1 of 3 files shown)


packages/devtools-reps/src/reps/function.js
```diff
@@ -74,11 +74,22 @@ function getTitle(grip, props) {
...
-  let name = grip.userDisplayName
-    || grip.displayName
-    || grip.name
-    || props.functionName
-    || "";
+  let { functionName } = props;
+  let end = functionName.length - 1;
+
+  props.functionName =
+    functionName.indexOf('"') == 0 && functionName.lastIndexOf('"') == end
```
nchevobbe: I think we could use `functionName.startsWith('"') && functionName.endsWith('"')`.
Also, what should we do if `functionName` is wrapped in single quotes (e.g. `'myName'`) ?
And we should handle cases where we don't have `functionName` as a prop (I expect tests to be broken with this change).

More globally, I wonder if this shouldn't be fixed on the consumer side (here I suppose the debugger): `functionName` is a props that we pass to the Rep "on purpose", as opposed to grip's properties that are defined by the protocol.
In my opinion, it makes more sense to do the `functionName` cleanup there, but I may be wrong. What do you think about it ?

packages/devtools-reps/src/reps/function.js
```diff
@@ -74,11 +74,22 @@ function getTitle(grip, props) {
...
+      ? functionName.substring(1, end)
+      : functionName;
+
+  let name =
+    grip.displayName != undefined && grip.displayName != props.functionName
```
nchevobbe: the ternary here looks a bit complex, maybe we could have : 
```js
let name;
if (grip.displayName != undefined && functionName != undefined &&  grip.displayName != functionName) {
  name = functionName + ": " + grip.displayName;
} else {
  name = grip.userDisplayName
    || grip.displayName
    || grip.name
    || functionName
    || "";
}
```

Note that I replace the `props.functionName` by `functionName`, since we do have the variable at hand.
I also added `functionName != undefined` in the first test, so we don't end up with things like `undefined: myDisplayName`.

I'm a bit worried about the `name = functionName + ": " + grip.displayName;`, can you give examples of what it should be like ?

hughugsy: I made the changes :) Thanks! :)  jQuery has two properties $ and jQuery each with the same displayName. So, instead of -> 
issue

Jason Laster proposed to make it show $: jQuery() and jQuery(). So, now after the patch it looks -> 
screenshot-2017-10-31 debugger

packages/devtools-reps/src/reps/function.js
hughugsy: I don't know where functionName becomes ""$"". I tried to find it to fix the issue there but couldn't. For single quotes, could we check with 'startsWith()' and 'endsWith()' ?

packages/devtools-reps/src/reps/function.js
```diff
@@ -22,81 +17,83 @@ const { span } = React.DOM;
...
 }
 
 function getFunctionName(grip, props) {
   let { functionName } = props;
   let end = functionName.length - 1;
+  let name;
 
-  props.functionName =
-    functionName.indexOf('"') == 0 && functionName.lastIndexOf('"') == end
+  functionName =
```
nchevobbe: will fail if props.functionName is undefined (which it can be)

hughugsy:
Hey! I wrote some test for this. How do I run the test? :smile:

nchevobbe:
Nice ! you can run them with : 
`cd packages/devtools-reps/; yarn test`

hughugsy:
expect(received).toBe(expected)
    
    Expected value to be (using ===):
      "function testName()"
    Received:
      "Invalid object"
      
      at Object.it (src/reps/tests/function.js:23:5)
          at Promise (<anonymous>)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
          at <anonymous>

all tests in tests/function.js failed like this :confused:

nchevobbe:
Mh, something must be wrong somewhere

nchevobbe:
@hughugsy they must be failing because an exception is thrown in the functionRep code, which get wrapped into an "Invalid object" component.

packages/devtools-reps/src/reps/function.js
```diff
@@ -29,62 +29,78 @@ function FunctionRep(props) {
...
 }
 
 function getFunctionName(grip, props) {
-  let name = grip.userDisplayName
-    || grip.displayName
-    || grip.name
-    || props.functionName
-    || "";
+  let { functionName } = props;
+  let end = functionName.length - 1;
```
nchevobbe: this is why you see failure on tests.
Most of the test do not pass the `functionName` props to the component (which is valid).
Because it is undefined, we get the following error : `TypeError: Cannot read property 'length' of undefined`

Here we can make things right like : 
```js
let { functionName } = props;
if (functionName) {
  let end = functionName.length - 1;
  functionName =
    functionName.startsWith('"') && functionName.endsWith('"')
      ? functionName.substring(1, end)
      : functionName;
}

if (
  grip.displayName != undefined 
  && functionName != undefined 
  && grip.displayName != functionName) {
  name = functionName + ": " + grip.displayName;
} else {
  name =
    grip.userDisplayName ||
    grip.displayName ||
    grip.name ||
    props.functionName ||
    "";
}

return cropString(name, 100);
```

hughugsy:
Hey! I am sorry had to finish midterms. Changed the way you said and all tests passed except the one I wrote :disappointed: Did I write it correctly?
`
stubs.set("ObjectProperty", {
  "type": "object",
  "class": "Function",
  "actor": "server1.conn7.obj45",
  "extensible": true,
  "frozen": false,
  "sealed": false,
  "isAync": false,
  "isGenerator": false,
  "name": "$",
  "displayName": "jQuery",
  "location": {
    "url": "debugger eval code",
    "line": 1
  }
});

describe("Function - Two properties with same displayName", () => {
  const object = stubs.get("ObjectProperty");

  it("renders object properties as expected", () => {
    expect(renderRep(object, {mode: undefined}).text())
    .toBe("function $:jQuery()");
    expect(renderRep(object, {parameterNames: []}).text())
    .toBe("function $:jQuery()");
    expect(renderRep(object, {parameterNames: ["a"]}).text())
    .toBe("function $:jQuery(a)");
    expect(renderRep(object, { parameterNames: ["a", "b", "c"] }).text())
    .toBe("function $:jQuery(a, b, c)");
    expect(renderRep(object, {
      mode: MODE.TINY,
      parameterNames: ["a", "b", "c"]
    }).text()).toBe("$:jQuery(a, b, c)");
  });
});

`

nchevobbe:
Hello @hughugsy , no worries !

So I think you forget to pass the `functionName` prop to the FunctionRep.
i.e. 
```js
expect(renderRep(object, {mode: undefined}).text())
    .toBe("function $:jQuery()");
```

should be

```js
expect(renderRep(object, {
  mode: undefined,
  functionName: "$"
}).text()).toBe("function $:jQuery()");
```
or something alike.

hughugsy:
There were some lint errors related to code I didn't write/change. Maybe because I manually reverted back prettier formating. Should I fix those lint errors?

nchevobbe:
could you fix them using eslint --fix ?

nchevobbe:
Erf, I was hoping to be able to rebase from github, but it looks like there is some conflict.
Could you try to rebase on top of master please ?

nchevobbe:
Looks like there is still a failure: 
```
Summary of all failing tests
 FAIL  object-inspector/tests/component/function.js
  ● ObjectInspector - functions › renders named function properties as expected

    expect(received).toBe(expected)
    
    Expected value to be (using ===):
      "testName()"
    Received:
      "fn:testName()"
      
      at Object.it (src/object-inspector/tests/component/function.js:39:33)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
      at process._tickCallback (internal/process/next_tick.js:109:7)
```

hughugsy:
In this case `functionName` is not equal to `displayName` but we must show `displayName` instead of `functionName: displayName`. So, our approach is wrong then?

nchevobbe:
This would work with the following patch: 
```diff
diff --git a/packages/devtools-reps/src/reps/function.js b/packages/devtools-reps/src/reps/function.js
index f2e0f0b..e0f65da 100644
--- a/packages/devtools-reps/src/reps/function.js
+++ b/packages/devtools-reps/src/reps/function.js
@@ -101,15 +101,29 @@ const arrayProperty = /\[(.*?)\]$/;
 const functionProperty = /([\w\d]+)[\/\.<]*?$/;
 const annonymousProperty = /([\w\d]+)\(\^\)$/;
 
-function getFunctionName(grip, props = {}) {
+function getFunctionName(grip, props) {
   let { functionName } = props;
-  let name =
+  let name;
+
+  if (functionName) {
+    let end = functionName.length - 1;
+    functionName =
+      functionName.startsWith('"') && functionName.endsWith('"')
+        ? functionName.substring(1, end)
+        : functionName;
+  }
+
+  if (grip.displayName != undefined && functionName != undefined &&
+    grip.displayName != functionName) {
+    name = functionName + ":" + grip.displayName;
+  } else {
+    name =
       grip.userDisplayName ||
       grip.displayName ||
       grip.name ||
       props.functionName ||
       "";
-
+  }
   const scenarios = [
     objectProperty,
     arrayProperty,
@@ -126,14 +140,6 @@ function getFunctionName(grip, props = {}) {
     return false;
   });
 
-  if (
-    grip.displayName != undefined &&
-    functionName != undefined &&
-    grip.displayName != functionName
-  ) {
-    name = functionName + ":" + grip.displayName;
-  }
-
   return cropString(name, 100);
 }
```

I wanted to push to your branch but I don't think you checked the `allow edits from maintainers` checkbox https://i.stack.imgur.com/wu4Wd.png

hughugsy:
Sorry, isn't that current patch? :slightly_smiling_face: And these are test failures...
`
● Function - Simplify name › annonymousProperty - jQuery.each(^)

    TypeError: Cannot match against 'undefined' or 'null'.
      
      at getFunctionName (src/reps/function.js:105:26)
      at Object.it (src/reps/tests/function.js:325:7)
          at Promise (<anonymous>)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
          at <anonymous>

  ● Function - Two properties with same displayName › renders object properties as expected

    expect(received).toBe(expected)
    
    Expected value to be (using ===):
      "function $:jQuery()"
    Received:
      "function jQuery()"
      
      at Object.it (src/reps/tests/function.js:335:5)
          at Promise (<anonymous>)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
          at <anonymous>
`

nchevobbe:
No, you have to apply the diff to fix the logic (and the tests) because
things changes a bit.

nchevobbe:
I can't push to your branch directly, so I created a new PR (https://github.com/devtools-html/devtools-core/pull/872). 
We can close this one since the changes will be merged there.

Issue #773, opened 2017-11-01
Provide a way to test all packages using a single command


MikeRatcliffe:
Having a way to run all tests from the root folder makes it far easier to run tests after making changes across packages.

jasonLaster:
look at the debugger-html or perf.html test-all commands :)

MikeRatcliffe:
We are using lerna in devtools-core so we can just do `"test": "lerna run test"`

PR #774, opened 2017-11-01
Provide a way to test all packages using a single command #773

MikeRatcliffe:
Associated Issue: #773

### Summary of Changes

* Provided a way to test all packages using a single command

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@
   - [Update a Package](./docs/versioning.md#update-a-package)
   - [Updating a Launchpad Dependency](./docs/versioning.md#updating-a-launchpad-dependency)
 
+### Testing
+
+From the root folder:
+
+```js
+yarn test
+```
+
 #### Syncing Assets
 
 We sync devtools assets like themes and widgets with [devtools-mc-assets][dma].
--- a/circle.yml
+++ b/circle.yml
@@ -8,7 +8,7 @@ machine:
     PATH: "${PATH}:${HOME}/.yarn/bin:${HOME}/${CIRCLE_PROJECT_REPONAME}/node_modules/.bin"
 
 test:
-  post:
+  override:
     - yarn lint-css
     - yarn lint-js
     - yarn flow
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -23,6 +23,14 @@
   - [Update a Package](./docs/versioning.md#update-a-package)
   - [Updating a Launchpad Dependency](./docs/versioning.md#updating-a-launchpad-dependency)
 
+### Testing
+
+From the root folder:
+
+```js
+npm run test
```
jasonLaster: `yarn test`

we are recommending yarn because we keep yarn.lock files for consistency

jasonLaster:
one small thing

jasonLaster:
lets also update `circle.yml`

MikeRatcliffe:
Tests failed because we took over 4G for one test... seems like 1 worker is probably enough.

nchevobbe:
Looking at https://circleci.com/gh/devtools-html/devtools-core/1498?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link , it looks like we do both `npm test` and `yarn test -- --maxWorkers 2`, and `npm test` probably does not restrict the number of workers.

MikeRatcliffe:
So let's try using npm test instead.

MikeRatcliffe:
I have moved the maxWorkers param into package.json.

Locally `"test": "jest -- --maxWorkers=2"` does not work but `"test": "jest --maxWorkers=2"` so I have used that.

nchevobbe:
I'd prefer having the option only on CI. locally I want the test to be as fast as possible

nchevobbe:
I may be wrong, but I think this is gonna fail since we execute `cd ./packages/devtools-reps; yarn test --numWorkers=2`, it should be `cd ./packages/devtools-reps; yarn test -- --numWorkers=2` I think

nchevobbe:
yep :/

MikeRatcliffe:
It may even be `-- --numWorkers 2`... we will find the correct incantation in the end.

MikeRatcliffe:
lol... you confused me with numWorkers but I have a good feeling about this one.

MikeRatcliffe:
That worked but let me squash these commits.

nchevobbe:
Yay 🎉

Issue #772, opened 2017-10-31
Upgrade to ES6 classes and silence all warnings


MikeRatcliffe:
We need to upgrade all of our components to ES6 classes as part of the path to using React 16.

createClass() is deprecated and it's replacement, createReactClass(), is only a temporary fix whereas ES6 classes are here to stay.

We have a [custom codemod](https://github.com/MikeRatcliffe/moz-codemod) that can perform the upgrade to ES6 classes so we need to run that and make any additional changes.

After we are using ES6 classes we can rebase pull request #770, which really is a great step forward.

MikeRatcliffe:
It has been easiest to upgrade to ES6 classes and address warnings all in one go. Warnings do not only come from our code... they also come from submodules and their dependencies so I have had to upgrade a bunch of dependencies to silence all warnings.

My patch will stop just short of upgrading to React 16.

MikeRatcliffe:
Enzyme was completely rewritten for Fibre so when we upgrade to react 16 somebody will need to fix all the tests or we can switch to another testing system.

Anyhow, no warnings and all ES6 classes.

brutalcrozt:
It will remove warning about babel-preset-es* right?

brutalcrozt:
@MikeRatcliffe this still open, is this still leaving some to do?
Sorry i just want to make sure :smile:

PR #775, opened 2017-11-01
Upgrade to ES6 classes and silence all warnings

MikeRatcliffe:
Associated Issue: #772

### Summary of Changes

* Always prefer ES6 classes in eslint.
* Switch all components over to ES6 classes.
* Use prop-types module instead of React.PropTypes.
* Use react-dom-factories instead of React.DOM.
* Upgraded dependencies that were causing warnings.


packages/devtools-launchpad/assets/Svg.js
```diff
@@ -1,5 +1,5 @@
 const React = require("react");
-const InlineSVG = require("svg-inline-react");
+import InlineSVG from "svg-inline-react";
```
nchevobbe: why should we do this ?
In my experience, `import` and `require` don't mix well.

packages/devtools-reps/package.json
```diff
@@ -28,15 +28,17 @@
...
+    "svg-inline-react": "^3.0.0"
   },
   "devDependencies": {
     "babel-jest": "^20.0.3",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
     "babel-preset-react": "^6.24.1",
     "devtools-config": "^0.0.15",
-    "devtools-launchpad": "^0.0.101",
-    "devtools-license-check": "^0.2.0",
+    "devtools-launchpad": "^0.0.108",
```
nchevobbe: this may cause some issues, I need to check if the reps repl still work with this patch

packages/devtools-styleguide/src/components/Variables.js
```diff
@@ -1,22 +1,18 @@
...
 function iterate(list, cbk) {
-  var index = list.length;
-  for (var i = 0; i < index; i++) {
+  let index = list.length;
+  for (let i = 0; i < index; i++) {
     cbk(list[i]);
   }
 }
 
-function map(list, cbk) {
```
nchevobbe: why was this removed ?

packages/devtools-launchpad/assets/Svg.js
MikeRatcliffe: To quote the author:

> I removed dist and es and made lib (which is ES2015 source) default, and there are several reason for it. For now use of ES2015 (and modules) is widespread, there are many tools supporting it (i.e. Webpack 2, Rollup), and you will use them anyway – **if you are using React/etc. If you want to stay in CommonJS land, please specify deps as 1.x.x. i.e) svg-inline-react: 1.x.x**

So require no longer works with this package.

The problem is that the 1.x.x versions trigger warnings so we have to use ES6 import.

packages/devtools-styleguide/src/components/Variables.js
MikeRatcliffe: Because it is not used anywhere... we can put it back if you prefer.

.eslintrc
```diff
@@ -67,7 +67,7 @@
     "react/no-unknown-property": "error",
     "react/prefer-es6-class": [
       "warn",
```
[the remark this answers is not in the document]
MikeRatcliffe: We always want to use ES6 classes in this repo and I believe this .eslintrc is only used for this repo so we should be fine.

packages/devtools-components/src/tree.js
```diff
@@ -138,7 +140,10 @@ const TreeNode = createFactory(createClass({
       renderItem(item, depth, focused, arrow, expanded)
     );
   }
-}));
+}
+
+const ArrowExpanderFactory = createFactory(ArrowExpander);
```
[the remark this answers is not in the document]
MikeRatcliffe: I didn't know that createFactory() is set to be deprecated (there is a TODO in the source).

Using new ArrowExpander() fails because you need a ReactClass and not a standard Class. The solution was to just replace createFactory with createElement calls, which is what the React folk recommend.

I have done this with all the crateFactory calls so that we don't need to change them in the future.

packages/devtools-components/src/tree.js
```diff
@@ -249,165 +254,188 @@ function oncePerAnimationFrame(fn) {
...
-    return {
+  constructor(props) {
+    super(props);
+
+    this.state = {
       seen: new Set(),
     };
-  },
+
+    this._onExpand = oncePerAnimationFrame(this._onExpand);
```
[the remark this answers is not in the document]
MikeRatcliffe: Of course, done.

brutalcrozt:
Yay it good news for me, it's really huge work ganbatte @MikeRatcliffe .  It have some conflict, i also waiting this one land.

nchevobbe:
@MikeRatcliffe could you revert https://github.com/devtools-html/devtools-core/pull/775/commits/e09aec9512e512015eedf2d1f2a2c2bf14b66967 please ? We have a pending RFC about this, and we may not end doing it :)

MikeRatcliffe:
Reverted :(

nchevobbe:
This breaks the reps repl with: 
```
 Error: Module build failed: Error: No PostCSS Config found in: /Users/nchevobbe/Projects/devtools-core/packages/devtools-reps/src/reps
    at /Users/nchevobbe/Projects/devtools-core/packages/devtools-launchpad/node_modules/postcss-load-config/index.js:51:26
    at <anonymous>
```
This is due to devtools-launchpad bump in devtools-reps.
I will file a bug to fix this

nchevobbe:
@MikeRatcliffe is this ready for review ?

MikeRatcliffe:
@nchevobbe Yes, this is ready for review.

nchevobbe:
sounds good, let's merge !

PR #780, opened 2017-11-03
Adds remapScopes utility function.

yurydelendik:
Associated Issue: https://github.com/devtools-html/debugger.html/pull/4521

### Summary of Changes

* Adds remapScopes utility function to the devtools-map-bindings package.

### DIFF
--- a/packages/devtools-map-bindings/src/remapScopes.js
+++ b/packages/devtools-map-bindings/src/remapScopes.js
@@ -1,0 +1,106 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// @flow
+
+import type {
+  Scope,
+  SourceScope,
+  MappedScopeBindings,
+  SyntheticScope
+} from "debugger-html";
+
+export type RemappedScope = {
+  scopes: SyntheticScope[],
+  start: number,
+  end: number
+};
+
+// Chunk split source scopes on function/closure boundary
+function rollupFunctionScopes(scopes) {
+  const { result } = scopes.reduce(
+    ({ isLast, result }, scope) => {
+      if (isLast) {
+        result.push([]);
+      }
+      result[result.length - 1].push(scope);
+      return {
+        isLast: scope.type === "function",
+        result
+      };
+    },
+    { isLast: true, result: [] }
+  );
+  return result;
+}
+
+function getBindingNames(summarizedScopes) {
+  return summarizedScopes.reduce(
+    (acc, { bindingsNames }) => acc.concat(bindingsNames),
+    []
+  );
+}
+
+// Performs mapping of the original parsed scopes to the locals mappings
+// based on the generated source parse and source map data.
+function remapScopes(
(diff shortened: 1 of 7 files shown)


packages/devtools-map-bindings/src/remapScopes.js
```diff
@@ -0,0 +1,96 @@
...
+        const index = searchIn.findIndex(s => name in s.bindings);
+        return index < 0 ? Math.max(index, max) : max;
+      }, 0);
+
+      // TODO double check if names were not matched/found -- the source maps
+      // and scope parsing can be broken.
+      // Moving to the function bounary (in generated scopes).
+      while (
+        foundInMax + 1 < searchIn.length &&
+        searchIn[foundInMax].type !== "function"
```
yurydelendik: `Scope.type` property is not reliable

packages/devtools-map-bindings/src/remapScopes.js
```diff
@@ -0,0 +1,96 @@
...
+): ?(RemappedScope[]) {
+  if (!scopes || scopes.length === 0) {
+    return null;
+  }
+  // Chunk original source scopes on function/closure boundary
+  const { result: scopeChunks } = scopes.reduce(
+    ({ isLast, result }, scope) => {
+      if (isLast) {
+        result.push([]);
+      }
```
jasonLaster: chunk is a pretty strange term...

I think i would prefer `functionScopes`,  perhaps we move this into a small helper function:

```js
const functionScopes = rollupFunctionScopes(scopes)
```

packages/devtools-map-bindings/src/remapScopes.js
```diff
@@ -0,0 +1,96 @@
...
+      // Process chunk of original parsed scopes: create used original names
+      // binding summary per scope and entire chunk.
+      const summarizedScopes = scopeChunk.map(({ type, bindings }) => ({
+        type,
+        bindingsNames: Object.keys(bindings)
+      }));
+      const names = summarizedScopes.reduce(
+        (acc, { bindingsNames }) => acc.concat(bindingsNames),
+        []
+      );
```
jasonLaster: it would be nice to have a small helper function: `getBindingNames` that can take our function scopes and return a list of names...

```js
const names = getBindingNames(scopeChunk)
```

packages/devtools-map-bindings/src/remapScopes.js
jasonLaster: do we need to fall back on `block` as well while we fix this?

packages/devtools-map-bindings/src/remapScopes.js
```diff
@@ -0,0 +1,96 @@
...
+      // Moving to the function bounary (in generated scopes).
+      while (
+        foundInMax + 1 < searchIn.length &&
+        searchIn[foundInMax].type !== "function"
+      ) {
+        foundInMax++;
+      }
+
+      // We found chunk of the function(s) that matches the scopeChunk names.
+      result.push({
```
jasonLaster: i'm a bit lost here

packages/devtools-map-bindings/src/remapScopes.js
yurydelendik: property reliable after all

PR #781, opened 2017-11-04
Remove resizeBreakpointGutter

jasonLaster:
Removes an unused function, that is now defined in debugger.html


jasonLaster:
merging as it's just removing an unused function

PR #785, opened 2017-11-07
Prepare devtools-map-bindings to process original scopes source maps

yurydelendik:
Associated Issue: https://github.com/devtools-html/debugger.html/pull/4613

### Summary of Changes

* Adds remapOriginalScopes to consume scope data and generate synthetic scopes
* Adds stubbed getOriginalScopes


nchevobbe:
@yurydelendik is this still relevant ? If yes, we should finish this PR

PR #789, opened 2017-11-09
Release 0.12.4

juliandescottes:
This is just a cherry pick of bc7bb952d66685c709216a675b3d674856df1c32 on top of reps v0.12.3.

It would solve https://github.com/devtools-html/debugger.html/issues/4639 which would really be nice to get in before 58 ships IMO.


juliandescottes:
Not sure how this should be merged though.

nchevobbe:
Thanks for doing this @juliandescottes .
Not sure what we should do with that too

juliandescottes:
Thanks for having a look, I guess we need to branch from 0.12.3 ?

nchevobbe:
yes, it sounds good

juliandescottes:
Let's wait to see if @jasonLaster is ok with the approach for the debugger then. After all that's the only reason why this version is needed.

jasonLaster:
ooh. I like this idea, it takes the pressure off the reduxification effort. I think we can do this as long as we merge into a 12x branch. and then NPM publish.

Thanks @juliandescottes

juliandescottes:
Cool! I created the branch, should be ready to merge and publish.

nchevobbe:
great, thanks a lot @juliandescottes  !
I'll think of doing this in the future if a similar thing happens

juliandescottes:
You're welcome :) Thanks for carefully tagging all the releases so far, would have been a nightmare without this.

Issue #749, opened 2017-10-09
Launching Firefox and Chrome both fail on Windows 10


digitarald:
Running `yarn start` in Git Bash, navigating to launchpad, I only get a promise error when clicking launch Firefox:

```
(node:28400) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Could not locate Firefox on the current system
(node:28400) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:28400) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 1)
```

Different error happens when launching Chrome:

```
Error: spawn UNKNOWN
    at _errnoException (util.js:1019:11)
    at ChildProcess.spawn (internal/child_process.js:325:11)
    at Object.exports.spawn (child_process.js:494:9)
    at handleLaunchRequest (C:\Users\micro\dev\debugger.html\node_modules\devtools-launchpad\src\server\launch.js:35:8)
    at Layer.handle [as handle_request] (C:\Users\micro\dev\debugger.html\node_modules\express\lib\router\layer.js:95:5)
    at next (C:\Users\micro\dev\debugger.html\node_modules\express\lib\router\route.js:137:13)
    at Route.dispatch (C:\Users\micro\dev\debugger.html\node_modules\express\lib\router\route.js:112:3)
    at Layer.handle [as handle_request] (C:\Users\micro\dev\debugger.html\node_modules\express\lib\router\layer.js:95:5)
    at C:\Users\micro\dev\debugger.html\node_modules\express\lib\router\index.js:281:22
    at Function.process_params (C:\Users\micro\dev\debugger.html\node_modules\express\lib\router\index.js:335:12)

```

digitarald:
Probably related to https://github.com/yarnpkg/yarn/pull/3848, as https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/src/server/launch.js#L35 tries to execute the JS file and Windows doesn't know about the shebang notation.

brutalcrozt:
For clarity, kindly remove 10 from windows 10. Since i verify it also fail on win 8. I wait pr number 775 land to work on. FF fail, perhaps the problem on the ff driver.js.

brutalcrozt:
does /claim work here? I am in progress

jasonLaster:
[...]

PR #793, opened 2017-11-10
Fix launching chrome on windows

brutalcrozt:
Associated Issue: #749

### Summary of Changes

* launch chrome with `node` as first arguments, since windows don't know about shebag notation
* catch error if selenium driver throw ff not found

### Misc
* Not bump version since, doesn't have credentials to do so
* No test

### Test Plan

Example test plan:

Still looking where the existing test :cold_sweat:

### DIFF
--- a/packages/devtools-launchpad/src/server/launch.js
+++ b/packages/devtools-launchpad/src/server/launch.js
@@ -6,7 +6,6 @@ const ps = require("child_process");
 const path = require("path");
 const { isFirefoxRunning } = require("./utils/firefox");
 const firefoxDriver = require("../../bin/firefox-driver");
-const isWindows = /^win/.test(process.platform);
 const {
   getValue
 } = require("devtools-config");
@@ -24,13 +23,6 @@ function handleLaunchRequest(req, res) {
         tcpPort: getValue("firefox.tcpPort")
       };
       if (!isRunning) {
-        process.on('unhandledRejection', (err)=> {
-          if (err.message.indexOf('Could not locate Firefox on the current system')>-1) {
-            console.error('selenium-webdriver could not locate Firefox, please launch it manually.');
-          } else {
-            throw err;
-          }
-        });
         firefoxDriver.start(location, options);
         res.end("launched firefox");
       } else {
@@ -41,11 +33,7 @@ function handleLaunchRequest(req, res) {
 
   if (browser == "Chrome") {
     const chromeDriver= path.resolve(__dirname, "../../bin/chrome-driver.js");
-    if (isWindows) {
-      ps.spawn('node', [chromeDriver, "--location", location]);
-    } else {
-      ps.spawn(chromeDriver, ["--location", location]);
-    }
+    ps.spawn('node', [chromeDriver, "--location", location]);
     res.end("launched chrome");
   }
 }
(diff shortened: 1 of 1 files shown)


brutalcrozt:
Note about launching ff:
* i already try with `try catch` `.catch` but it not work well with `seleniumDriver.build()` , 
* this leaving bad UX, since response already sent by server. i can't change client UI to explicitly as to launch manually. Please let me know if there is better way to do so. Guide me as well :smile: 
* Common failure is FF not found, i can't guessing where is it. Since we are mozzilian, Maybe someone can point it out specificly.

My first PR, can't wait how the views goes :smiling_imp:

jasonLaster:
Looks great to me.

@janodvarko mind testing on your windows machine?

packages/devtools-launchpad/src/server/launch.js
```diff
@@ -32,8 +40,12 @@ function handleLaunchRequest(req, res) {
   }
 
   if (browser == "Chrome") {
-    ps.spawn(path.resolve(__dirname, "../../bin/chrome-driver.js"),
-     ["--location", location]);
+    const chromeDriver= path.resolve(__dirname, "../../bin/chrome-driver.js");
+    if (isWindows) {
+      ps.spawn('node', [chromeDriver, "--location", location]);
```
Garbee: Wouldn't this also work on other systems? I'm thinking this entire if condition can be removed and instead we could simply always spawn node and provide the arguments in.

packages/devtools-launchpad/src/server/launch.js
```diff
@@ -23,6 +24,13 @@ function handleLaunchRequest(req, res) {
         tcpPort: getValue("firefox.tcpPort")
       };
       if (!isRunning) {
+        process.on('unhandledRejection', (err)=> {
+          if (err.message.indexOf('Could not locate Firefox on the current system')>-1) {
```
Garbee: I have a feeling that depending on the users localization this could only work on English based systems. Is there any way to do a test that isn't hard coded to the text of the message?

brutalcrozt: I don't know better way to do this, if it possible i would rather have the promise rejected instead of catching the error thrown on the process. I already mention that.

packages/devtools-launchpad/src/server/launch.js
brutalcrozt: I am not sure of that, but its agreeable when it produce more cleaner code. Let me try it when i got the time and chance.

packages/devtools-launchpad/src/server/launch.js
Garbee: Then let's look at this problem slightly differently. Is the error message really that important to rework? Instead of analyzing the message in the first place, could we instead check for error and if found display out own little fail message followed by a newline and then the original error message itself?

By not doing any checking or manipulation the error from the underlying program gets out to the user. That could end up being more useful in different error states as well rather than this one specific case being looked for. It also avoids any internationalization issues. Overall, making the code simpler, more robust, and better for end-users in more situations.

brutalcrozt: I cant argue with your point, but also i won't digging on selenium driver. Also the probability user install `custom location` is really low. So i leaving the firefox case as is.

brutalcrozt:
I hope @Garbee is not busy on real world. I already make changes, i like to ask review.

PR #812, opened 2017-11-22
Add a button to jump to a function definition.

### LINKED ISSUE #640: Add an icon/link to jump to a Function declaration
Here is the package we get for a function grip: 
```json
{
  "from": "server2.conn12.child58/consoleActor2",
  "input": "handleTestCase",
  "result": {
    "type": "object",
    "actor": "server2.conn12.child58/obj34",
    "class": "Function",
    "extensible": true,
    "frozen": false,
    "sealed": false,
    "name": "handleTestCase",
    "displayName": "handleTestCase",
    "location": {
      "url": "https://nchevobbe.github.io/demo/console-test-app.html",
      "line": 158
    }
  },
  "timestamp": 1504248246306,
  "exception": null,
  "frame": null,
  "helperResult": null,
  "notes": null
}
```

We could use the `location` object to call an `openInDebugger` prop that would look like what we already have in the console: http://searchfox.org/mozilla-central/source/devtools/client/webconsole/new-console-output/new-console-output-wrapper.js#105-107

nchevobbe:
Fixes #640.

The Function rep would accept an onViewSourceInDebugger prop, and the
icon would be shown when we do have this props as well as a location with
a non-null url (so we don't show it on prototypes for example).
Clicking on the button would call the prop with the location object.
A test is added to make sure we handle the different possible cases.

### Test Plan
Open the Reps test app and evaluate a function

### Screenshots
screen shot 2017-11-22 at 14 04 28

### DIFF
--- a/packages/devtools-reps/src/launchpad/components/Result.js
+++ b/packages/devtools-reps/src/launchpad/components/Result.js
@@ -81,6 +81,8 @@ class Result extends Component {
         releaseActor,
         mode: MODE[modeKey],
         onInspectIconClick: nodeFront => console.log("inspectIcon click", nodeFront),
+        onViewSourceInDebugger: location =>
+          console.log("onViewSourceInDebugger", {location}),
       })
     );
   }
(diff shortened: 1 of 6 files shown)


nchevobbe:
Asking @violasong for UI review :)

nchevobbe:
Let's merge this, we can always tweak the UI later :)

PR #817, opened 2017-11-28
Fix linkification for cropped URLs

### LINKED ISSUE #616: Cropped URL href is wrong
Steps to reproduce: 

Render a url string rep with a `cropLimit` prop that will crop the URL, e.g. 
```js
StringRep.rep({
  object: "http://example.com",
  cropLimit: 15
});
```

Expected results:
The URL is rendered as `"http://…ple.com"`, but as an `href` attribute of `"http://example.com"`

Actual results:
The URL is rendered as `"http://…ple.com"`, as well as the `href` attribute, which is wrong

---

The following test should pass, which it does not at the moment: 
```js
  it("renders a URL for a cropped string", () => {
    const url = "http://example.com";
    const openLink = jest.fn();
    const element = renderRep(url, {
      openLink,
      useQuotes: false,
      cropLimit: 15
    });
    expect(element.text()).toEqual(url);
    const link = element.find("a");
    expect(link.prop("href")).toBe(url);
    expect(link.prop("title")).toBe(url);

    link.simulate("click");
    expect(openLink).toBeCalledWith(url);
  });
```

nchevobbe:
Fixes #616.

### DIFF
--- a/packages/devtools-reps/src/reps/string.js
+++ b/packages/devtools-reps/src/reps/string.js
@@ -14,6 +14,7 @@ const {
   sanitizeString,
   wrapRender,
   tokenSplitRegex,
+  ELLIPSIS,
 } = require("./rep-utils");
 
 const dom = require("react-dom-factories");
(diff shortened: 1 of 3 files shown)


packages/devtools-reps/src/reps/string.js
```diff
@@ -62,47 +63,125 @@ function StringRep(props) {
...
+}
+
+/**
+ * Get an array of the elements representing the string, cropped if needed,
+ * with actual links.
+ *
+ * @param {String} text: The actual string to linkify.
+ * @param {Integer | null} cropLimit
+ * @param {Boolean} omitLinkHref: Do not create an href attribute if true.
+ * @param {Function} openLink: Function handling the link opening.
```
janodvarko: A bit off-topic question. The colon after arg name is part of JSdoc spec?
I like it and the comment clearer when it's used. Should we have a rule
for it and use it everywhere?

nchevobbe:
[JSDoc examples](http://usejsdoc.org/howto-es2015-classes.html) do use a final dot for comments.
Maybe we can enforce that with eslint, not sure though.

nchevobbe:
Opps, read too fast. No, it does not seem like they use colon, but I do like them better too.

nchevobbe:
Using Administrator privileges since Honza approved the request (but did not click the "approve" button :))

PR #820, opened 2017-11-29
Simplify function

nchevobbe:
Underlying logic borrowed from the debugger


jasonLaster:
Wow! Great work. 

You should share a screenshot

PR #824, opened 2017-12-01
Bump devtools-splitter to 0.0.6

nchevobbe:


nchevobbe:
r=me for the version bump.

Makes the splitter usable with React 16

PR #829, opened 2017-12-04
Use devtools-components-0.1.2 in reps

nchevobbe:


nchevobbe:
Dependency minor version update.
r=me

PR #832, opened 2017-12-05
Fix console message explaining how to turn on logging.

jimblandy:
The console message shows the wrong property name.


packages/devtools-launchpad/bin/firefox-proxy
```diff
@@ -14,7 +14,7 @@ function proxy(host, webSocketPort, tcpPort, logging) {
   console.log(`Listening for WS on localhost:${webSocketPort}, all traffic is proxied to ${host}:${tcpPort}`);
   if (!logging) {
     console.log("Protocol messages can be logged by enabling " +
-    "logging.firefoxProtocol in your local.json config");
+    "logging.firefoxProxy in debugger.html/configs/local.json");
```
nchevobbe: the launchpad is used in various places (console, netmonitor, reps, and debugger), so we shouldn't specify a given tool path.

PR #845, opened 2017-12-12
Add handling for a DOMException error; Fixes #682

### LINKED ISSUE #682: DomException do not show the error message
Steps to reproduce:
- Open the Reps repl or the console
- Evaluate 
```js
  try {
    var foo = document.querySelector("foo;()bar!");
  } catch (ex) {
    ex;
  }
```

Expected results:
It displays `DOMException: SyntaxError: "'foo;()bar!' is not a valid selector"`

Actual results:
It only shows `DOMException {  }`, and you have to expand the object to see the message

gregtatum:
Associated Issue: #682


nchevobbe:
CI is ✅ , let's merge !

Issue #846, opened 2017-12-13
[Launchpad] Reloading previously closed external tab causes blank screen.


lukaszsobek:
STR:
- open website in target browser
- open launchpad in debugger browser
- navigate to debugger tab in debugger browser
- close website in target browser (or whole target browser)
- reload debugger in debugger browser

ER: Debugger starts up without loaded tabs or source.

AR: Empty debugger +
index.js:195 Uncaught (in promise) TypeError: Cannot destructure propertytabof 'undefined' or 'null'. at bootstrap (index.js:195) at <anonymous>

nchevobbe:
@lukaszsobek is this still happening ?

PR #847, opened 2017-12-13
Opens launchpad on reload of previously closed external tab

lukaszsobek:
Associated Issue: #846

### Summary of Changes

- added check if debugged target exists
- if not we default to launchpad

### Test Plan

- tested on running debugger

### Screenshots/Videos 
reload

### DIFF
--- a/packages/devtools-launchpad/src/index.js
+++ b/packages/devtools-launchpad/src/index.js
@@ -180,10 +180,15 @@ async function getTabs(actions) {
 async function bootstrap(React, ReactDOM) {
   const connTarget = getTargetFromQuery();
   if (connTarget) {
-    const { tab, tabConnection } = await startDebugging(connTarget);
+    const debuggedTarget = await startDebugging(connTarget);
 
-    await updateConfig();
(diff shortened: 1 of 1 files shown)


packages/devtools-launchpad/src/index.js
```diff
@@ -180,10 +180,13 @@ async function getTabs(actions) {
 async function bootstrap(React, ReactDOM) {
   const connTarget = getTargetFromQuery();
   if (connTarget) {
-    const { tab, tabConnection } = await startDebugging(connTarget);
+    const debuggedTarget = await startDebugging(connTarget);
 
-    await updateConfig();
-    return { tab, connTarget, tabConnection };
+    if (debuggedTarget) {
```
nchevobbe: maybe we want to show something if debuggedTarget is falsy ?

lukaszsobek: Was thinking we could log an information to console that the tab has been closed because as a result of the debugged source having been closed.

nchevobbe:
Looks good, thanks @lukaszsobek

PR #849, opened 2017-12-13
Don't shrink indent

nchevobbe:


nchevobbe:
Broken indent on debugger : 
image

Fixed: 
image

Issue #684, opened 2017-09-18
CSSRules are rendered as strings


nchevobbe:
Steps to reproduce:
1. Open the Reps repl or the console
2. Evaluate `document.styleSheets`

Expected results: 
Rules are shown like : 
```
▶︎ cssRules: [CSSStyleRule, …]
```

Actual results:
Only the rule string is shown, which is weird, because it might make the user think that they are dealing with an array of strings
```
▶︎ cssRules: ["body"]
```

gregtatum:
Adding myself to this to maybe take in the future @gregtatum

bgrins:
Updated STR:

execute: `document.styleSheets[0].cssRules`

weilonge:
I would like to take this issue. Can someone help me?

nchevobbe:
Sure !

PR #852, opened 2017-12-14
Modify ObjectWithText to show the class of the object

weilonge:
Associated Issue: #684

Implement CSSRule reps for showing CSSStyleRule node.

### DIFF
--- a/packages/devtools-reps/src/reps/object-with-text.js
+++ b/packages/devtools-reps/src/reps/object-with-text.js
@@ -11,6 +11,8 @@ const {
   wrapRender,
 } = require("./rep-utils");
 
+const String = require("./string").rep;
+
 const dom = require("react-dom-factories");
 const { span } = dom;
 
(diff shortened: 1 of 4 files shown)


weilonge:
@nchevobbe Could you help to review this patch?

nchevobbe:
@weilonge sure !

So CI is not happy right now. It seems that the `supportObject` function from `css-rule` throws with some object: 
```
TypeError: Cannot read property 'class' of undefined
at Object.supportsObject (/home/ubuntu/devtools-core/packages/devtools-reps/src/reps/css-rule.js:31:16)
```
which is 
```js
function supportsObject(object, noGrip = false) {
  return object.class == "CSSStyleRule";
}
```
we should check if we have a truthy object before checking `object.class`

weilonge:
Thank you for the patch to fix the test. I am going through all errors and fixing them.

nchevobbe:
So it looks like these kind of grips (CSSStyleRule & CSSMediaRule) are handled by the ObjectWithText rep.
So we need to apply the logic we put in CSSRuleRep from this PR to ObjectWithText, and adapt the tests.
Also, the current test is only for a CSSStyleRule, we should have one for CSSMediaRule.
Here's a CSSMediaRule grip that we can put in a stub in order to test it : 
```js
{
    "type": "object",
    "actor": "server2.conn8.child17/obj30",
    "class": "CSSMediaRule",
    "extensible": true,
    "frozen": false,
    "sealed": false,
    "ownPropertyLength": 0,
    "preview": {
      "kind": "ObjectWithText",
      "text": "(min-height: 680px), screen and (orientation: portrait)"
    }
  }
```

Issue #836, opened 2017-12-07
ElementNode with LongString attributes throw


nchevobbe:
Steps to reproduce:
1. Evaluate:
```js
var div = document.createElement("div");
div.setAttribute("data-test","a".repeat(50000));
 div;
```

Expected results:
The data-test attribute is displayed with longString's initial value

Actual results:
The data-test attribute fails to render (`Invalid Object`), and there is an error in the browser console:
```
10:08:06.424 TypeError: str.replace is not a function
Stack trace:
[object Object] 1 reps.js:337:7
	wrappedFunction resource://devtools/client/shared/components/reps/reps.js:337:7
	getElements/attributeElements< resource://devtools/client/shared/components/reps/reps.js:4070:108
	getElements resource://devtools/client/shared/components/reps/reps.js:4068:29
	ElementNode resource://devtools/client/shared/components/reps/reps.js:4007:18
	wrappedFunction resource://devtools/client/shared/components/reps/reps.js:335:14
	Rep resource://devtools/client/shared/components/reps/reps.js:586:10
	renderGrip resource://devtools/client/shared/components/reps/reps.js:4991:12
	renderTreeItem resource://devtools/client/shared/components/reps/reps.js:4944:21
	render resource://devtools/client/shared/components/reps/reps.js:5240:8
	_renderValidatedComponentWithoutOwnerOrContext resource://devtools/client/shared/vendor/react-dom.js:5260:25
	_renderValidatedComponent resource://devtools/client/shared/vendor/react-dom.js:5283:27
	performInitialMount resource://devtools/client/shared/vendor/react-dom.js:4823:25
	mountComponent resource://devtools/client/shared/vendor/react-dom.js:4719:16
	mountComponent resource://devtools/client/shared/vendor/react-dom.js:11721:18
	mountChildren resource://devtools/client/shared/vendor/react-dom.js:10612:28
	_createInitialChildren resource://devtools/client/shared/vendor/react-dom.js:6343:27
	mountComponent resource://devtools/client/shared/vendor/react-dom.js:6162:7
	mountComponent resource://devtools/client/shared/vendor/react-dom.js:11721:18
	performInitialMount resource://devtools/client/shared/vendor/react-dom.js:4832:18
	mountComponent resource://devtools/client/shared/vendor/react-dom.js:4719:16
	mountComponent resource://devtools/client/shared/vendor/react-dom.js:11721:18
[...]

PR #868, opened 2017-12-19
Merge string and longstring reps

### LINKED ISSUE #550: Merge string and longString reps
Currently we have 2 reps to deal with strings.
It's unfortunate because when we add feature we have to look at 2 places (or we could forget one).
It should be doable to only have the String one and adapt it for longString.

kootoopas:
Closes #550 & #836

### Summary of Changes
*(not sure if my approach is correct but here goes)*
What I did was specify shared prop types and functions, to reduce the source of truth.

Tests are passing.


nchevobbe:
Hey @kootoopas , thanks for the PR !

I think the goal of the issue was to totally remove the LongString rep file, so we only have one file to deal with strings.
But this is a good first step to do the migration. Do you think we should do this in 2 steps ? Or does it make sense to edit your PR to only have this one file ? This is up to you

kootoopas:
I see. I can handle that in this PR. :)

kootoopas:
@nchevobbe I think that my code could be cleaner with regards to checking for `longString` within the render function and cropping, but it's a start! :)

kootoopas:
@nchevobbe Hello and Merry Christmas! :)

Were there any issues with how `LongStringRep` was merged to `StringRep`? Please don't hesitate to point them out!

nchevobbe:
Hello @kootoopas .
I'm on PTO until wednesday so I'll check when I'm back.
I set myself as a reviewer so I don't forget.

One thing I notice is that your PR needs to be rebased and you might have conflicts :)

kootoopas:
Just synced with master and resolved the conflicts. I'm assuming that's what you meant by rebasing. (since I don't see how `git rebase` would be useful in this occasion)

packages/devtools-reps/src/reps/string.js
```diff
@@ -188,13 +246,21 @@ function getCroppedString(text, offset = 0, startCropIndex, endCropIndex) {
...
+
   return getGripType(object, noGrip) == "string";
 }
 
 // Exports from this module
 
 module.exports = {
   rep: wrapRender(StringRep),
-  supportsObject,
+  supportsObject
```
nchevobbe: we should keep this comma

packages/devtools-reps/src/reps/string.js
```diff
@@ -38,34 +39,50 @@ StringRep.propTypes = {
...
+
+  const config = getElementConfig({
+    className,
+    style,
+    actor: object.actor
+  });
+
   if (!containsURL(text)) {
-    if (shouldCrop) {
+    if (!isLong && shouldCrop) {
```
nchevobbe: Maybe we can add a comment here to say that longStrings crop is handled before

packages/devtools-reps/src/reps/string.js
```diff
@@ -75,6 +92,47 @@ function StringRep(props) {
     ...getLinkifiedElements(text, shouldCrop && cropLimit, omitLinkHref, openLink));
```
nchevobbe: Now i'm not sure this would be handled like it should for long string.
Here we assume the string contains at least one URL, and we render them as actual URL's.

We should have new tests for longStrings to ensure this case is handled correctly.

packages/devtools-reps/src/reps/string.js
```diff
@@ -133,6 +149,23 @@ function getElementConfig(opts) {
...
 }
 
+function maybeCropString(opts, text) {
+  const {
+    isLong,
+    shouldCrop,
+    cropLimit,
+  } = opts;
+
+  // Cropping of LongString has been handled before formatting.
```
kootoopas: @nchevobbe I broke cropping into 2 functions for each string type (that use the same precalculated options), to emphasize that they are different from each other.

packages/devtools-reps/src/reps/string.js
kootoopas: @nchevobbe done!

packages/devtools-reps/src/reps/string.js
```diff
@@ -38,41 +39,131 @@ StringRep.propTypes = {
...
     openLink,
     omitLinkHref = true,
   } = props;
 
+  let text = object;
+
+  const isLong = isLongString(object);
+  const shouldCrop = (!member || !member.open) && cropLimit && text.length > cropLimit;
+
+  text = maybeCropLongString({
```
nchevobbe: I find it a bit confusing. Could we have the 
```js
if (isLong) {
  text = maybeCropLongString({
```

packages/devtools-reps/src/reps/stubs/long-string-with-url.js
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const urlFullText = "http://example.com?param=" + "a".repeat(2000);
+const urlFullTextLength = urlFullText.length;
+const urlInitialText = urlFullText.substring(0, 1000);
+
+const stubs = new Map();
+stubs.set("testLongUrl", {
```
kootoopas: Is there any case in which `fullText` is not accessible? 🤔 If there is, there's currently no way to keep the cropped url's `href` uncropped and valid.

nchevobbe: yes, if you log the following in the console `"a ".repeat(10000)` , we don't have the fullText in the grip. 
There should be a way then to load the entire text ( see https://bugzilla.mozilla.org/show_bug.cgi?id=1310630).

Loirooriol:
I think URLs should not be searched in longStrings. It would be bad for performance and seems unnecessary. Chrome doesn't do it.

nchevobbe:
okay, sounds reasonable

kootoopas:
Hey there @nchevobbe. :)

I think I'm moving to the right direction. I dug up and used the existing `getString` method of the client. If there's a more neat way to get it please tell me!

My modifications in `ObjectInspector.setExpanded()` are probably a bit funky too.

I also created a `Cutoff` component for the purpose of hiding the remainder of a value, the `fullText` in this case. Here's a gif of the functionality:

togglable-long-string

packages/devtools-reps/src/reps/string.js
```diff
@@ -38,41 +39,124 @@ StringRep.propTypes = {
...
+    escapeWhitespace
+  }, text);
+
+  const config = getElementConfig({
+    className,
+    style,
+    actor: object.actor
+  });
+
+  if (!containsURL(text) || isLong) {
```
Loirooriol: You can reverse the order to get short-circuit evaluation: `isLong || !containsURL(text)`. Also consider

```js
if (!isLong) {
  if (containsURL(text)) {
    return span(config, ...getLinkifiedElements(/* ... */));
  }
  text = maybeCropString(/* ... */);
}
return span(config, text);
```

kootoopas: Good point. 👆

nchevobbe:
@kootoopas thanks a lot ! There's seems to be conflicts on your PR. Could you try to update to latest master and rebase if it's not too painful ? Thanks

kootoopas:
@nchevobbe will do!

packages/devtools-reps/src/object-inspector/utils/load-properties.js
```diff
@@ -76,6 +79,10 @@ function loadItemProperties(
     loadingPromises.push(enumSymbols(getObjectClient(), start, end));
   }
 
+  if (shouldLoadFullTextProperty(item, loadedProperties)) {
+    loadingPromises.push(getLongStringFullText(getValue(item)));
```
kootoopas: This is def not ready yet, I need to pass the `getlongStringFullText(stringGrip)` redux bound action to `loadItemProperties()`.

nchevobbe: I did not realized you were looking at expanding longStrings as well, this is great !
Maybe we should this PR focused on merging longString and string rep, and you can move the work on the object inspector on a dedicated PR ?

kootoopas: Is there a way to expose `WebConsoleClient.getString` by passing in the longString's object client, as we're doing for other properties, and then calling `objectClient.getString` from there? This way I'll be able to add a `getString` util fn in `object-inspector/utils/client.js`, and structure it more consistently. 💪

kootoopas: @nchevobbe  Whoops just noticed ur comment, uum sure thing!

packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -428,12 +429,24 @@ class ObjectInspector extends Component {
     );
   }
 
+  getDefaultPropsByRep(object) {
```
nchevobbe: can changes to this file be moved to its own PR ? 
I'm not sure what we are trying to achieve here.

packages/devtools-reps/src/reps/reps.css
```diff
@@ -66,6 +66,12 @@
 
 .objectBox-string a, .objectBox-string a:visited {
   color: currentColor;
+  text-decoration: none;
+  font-style: italic;
+}
+
+.objectBox-string a:hover {
+  text-decoration: underline;
```
nchevobbe: 👍

packages/devtools-reps/src/reps/string.js
```diff
@@ -188,7 +273,15 @@ function getCroppedString(text, offset = 0, startCropIndex, endCropIndex) {
   return text;
 }
 
+function isLongString(object) {
+  return object.type === "longString";
```
nchevobbe: could we guard with object for extra safety ?
```js
return object && object.type === "longString"
```

packages/devtools-reps/src/object-inspector/index.js
kootoopas: Hmm, I was trying to set `omitLinkHref` to false so that links would be rendered. Is there a more proper way to override the default option for the launchpad instead of injecting an options layer only for the purpose of setting `omitLinkHref`, which seems pretty wrong now that I think of it. :|

nchevobbe: Let's the consumer of the ObjectInspector decude if they want to pass omitLinkHref :)
The default value here is to prevent security issue (a plain link with no way to open the link outside of the devtools toolbox).
The link is always rendered, we only don't add the href attribute

packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -15,6 +15,7 @@ const classnames = require("classnames");
 const {
   REPS: {
     Rep,
+    StringRep,
```
nchevobbe: i don't think this is needed anymore

packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -433,9 +434,11 @@ class ObjectInspector extends Component {
     props: Props
   ) {
     const object = getValue(item);
+
     return Rep(Object.assign({}, props, {
       object,
       mode: props.mode || MODE.TINY,
+      omitLinkHref: props.omitLinkHref || false,
```
nchevobbe: i'd rather remove this change completely. We need to let the caller be responsible for this, and we want the default to be handled by the grip (which set it to true, what we want).

kootoopas: Agreed, it can get pretty chaotic if you change defaults at different points in code. :)

PR #872, opened 2017-12-23
Duplicate scopes

nchevobbe:
This is a modified version of https://github.com/devtools-html/devtools-core/pull/771 green tests


nchevobbe:
Force merge since this is @hughugsy patch and I already granted the review there.

Issue #842, opened 2017-12-11
Symbol grip with actor is displayed as "Object {  }"


Loirooriol:
[Bug 1424722](https://bugzilla.mozilla.org/show_bug.cgi?id=1424722) will create actors for symbols so that they can be stored as variables.

The problem is that reps seems to consider that a grip with an actor must refer to an object, and displays `Object {  }` instead of `Symbol(name)`. The old console frontend works fine.

Cc @nchevobbe

AnshulMalik:
I can look into this :)

AnshulMalik:
I need some more inputs here please, Reps detects Symbol() and displays it as Symbol.
I might not be getting the point :(

Loirooriol:
You need to add the patch from the linked bug, https://reviewboard.mozilla.org/r/207016/diff/1#index_header. It adds an actor to the symbol grip. Right now the symbol grip has no actor so it's displayed correctly.

nchevobbe:
Anshul, this means we have to update the symbol stubs in reps code to match
what we'll have with oriol's patch. Then, we can see the test failing and
fix them.

My feeling is that here we probably default to object because supportObject
now return false.

Loirooriol:
OK, the problem is that `getGripType` is supposed to return `"symbol"` for symbol grips.

```js
function getGripType(object, noGrip) {
  let type = typeof object;
  if (type == "object" && object instanceof String) {
    type = "string";
  } else if (object && type == "object" && object.type && noGrip !== true) {
    type = object.type;
  }

  if (isGrip(object)) {
    type = object.class;
  }

  return type;
}
```

But `isGrip` returns true is there is an `actor` property:

```js
function isGrip(object) {
  return object && object.actor;
}
```

so `object.class` is returned instead, which is undefined.

As I said in some other bug, this `isGrip` function is too naive and unreliable, it should just be removed in favor of the `noGrip` parameter.

And can grips really be string objects?

I guess it should simply be something like

```js
function getGripType(object, noGrip) {
  if (!object || noGrip) {
    return typeof object;
  }
  return object.type || object.class;
}
```

nchevobbe:
@AnshulMalik do you have some time to look at this ? I think we should move quickly here since https://bugzilla.mozilla.org/show_bug.cgi?id=1424721 landed, which means Symbol will be broken in Nightly in the next days.

AnshulMalik:
I am kind of lost here @nchevobbe, trying to make sense of the code.

[...]

PR #874, opened 2017-12-26
Allow symbol grips to have an actor

Loirooriol:
Associated Issue: #842

### Summary of Changes

 - Does not assume that grips with an actor will be object grips with a class
 - Does not access `type` nor `actor` properties if `noGrip === true`
 - Removes a seemingly unnecessary case about String objects, there shouldn't be such kind of grip. And if this was necessary in the `noGrip === true` case, then I guess that Boolean, Number and Symbol objects should also be handled.

You can update stubs or whatever else once symbol actors are really added.

### Test Plan

I modified added an actor to symbol grips and applied this patch to `devtools/client/shared/components/reps/reps.js`. Symbol values were properly displayed in the console.

### DIFF
--- a/packages/devtools-reps/src/reps/rep-utils.js
+++ b/packages/devtools-reps/src/reps/rep-utils.js
@@ -341,18 +341,13 @@ function getGripPreviewItems(grip) {
  * @return {boolean}
  */
 function getGripType(object, noGrip) {
-  let type = typeof object;
-  if (type == "object" && object instanceof String) {
-    type = "string";
-  } else if (object && type == "object" && object.type && noGrip !== true) {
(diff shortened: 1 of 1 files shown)


Loirooriol:
@nchevobbe I don't know how to change a commit title. But when merging I think you can squash it into anther one with a different title.

nchevobbe:
`git commit --amend` will allow you to change the commit message.
What I refer as "title" here is the first line of the commit message

Loirooriol:
OK, done

nchevobbe:
great

nchevobbe:
Merged ! Thanks for your first contribution on devtools-core :)

Issue #878, opened 2017-12-31
Changes after adding symbol actors


Loirooriol:
In [bug 1424722](https://bugzilla.mozilla.org/show_bug.cgi?id=1424722) I plan to add a symbol actor to symbol grips. Then,

 - The symbol stubs should be updated.

   https://github.com/devtools-html/devtools-core/blob/94aae83569c2d27432008775bef7795ef66fe550/packages/devtools-reps/src/reps/stubs/symbol.js#L6-L13

   Something like `actor: "server1.conn1.child1/symbol" + someNumber,`

 - A `data-link-actor-id` should be added in `SymbolRep`

   https://github.com/devtools-html/devtools-core/blob/94aae83569c2d27432008775bef7795ef66fe550/packages/devtools-reps/src/reps/symbol.js#L30

   ```js
     return span({
       className,
       "data-link-actor-id": object.actor,
     }, `Symbol(${name || ""})`);
   ```

 - Test that with `expectActorAttribute`

   https://github.com/devtools-html/devtools-core/blob/94aae83569c2d27432008775bef7795ef66fe550/packages/devtools-reps/src/reps/tests/symbol.js#L18

   ```js
   const {
     expectActorAttribute
   } = require("./test-helpers");
   expectActorAttribute(renderedComponent, stub.actor);
   ```

Loirooriol:
Fixed in #879

PR #879, opened 2017-12-31
Add an actor id data attribute on symbol rep

Loirooriol:
This then can be used to retrieve the actor the node is representing.
For example, in the console we can use it in the context menu entry
to store an object as a global variable.

Associated Issue: #878

### Summary of Changes

 - Add an actor id data attribute on symbol rep

### Test Plan

I added an actor to symbol grips and applied #874 and this patch to devtools/client/shared/components/reps/reps.js. The symbols could be stored as global variables.

I also tried it without symbol actors (which is the current behavior) and then `object.actor` is undefined and react just ignores it. So it's harmless if this lands before [bug 1424722](https://bugzilla.mozilla.org/show_bug.cgi?id=1424722).

### DIFF
--- a/packages/devtools-reps/src/reps/stubs/symbol.js
+++ b/packages/devtools-reps/src/reps/stubs/symbol.js
@@ -5,11 +5,13 @@
 const stubs = new Map();
 stubs.set("Symbol", {
   type: "symbol",
+  actor: "server1.conn1.child1/symbol1",
   name: "foo"
 });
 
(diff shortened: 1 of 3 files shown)


nchevobbe:
@Loirooriol I hope we can land a new bundle by the end of the week with this PR inside. Will you have the time to make the suggested changes ? If not, I can do it.

Loirooriol:
OK, this fixes #878 completely. I wasn't sure about the actor ID numbers, I just used 1 and 2.

PR #859, opened 2017-12-16
Add length postfix for arraylike reps

kootoopas:
Associated Issue: #566

### Summary of Changes

Display length postfix in title of arraylike objects (Arrays, Sets, etc).

### Test Plan

Apart from syncing the unit tests, here's a screenshot of arraylike objects:

screen shot 2017-12-16 at 5 38 21 pm

Example test plan:

- Click the `array`, `map & sets` and the `promise` buttons in the reps gui to see for yourself. :)

Issue #566, opened 2017-08-07
Improve array preview


nchevobbe:
Based on a discussion with @julienw , it seems that we can do a better job at previewing arrays in Reps.

Basically it would help if one could quickly see how big an array is (either by showing it in the preview, or in a tooltip).

---

Test case: `Array(1000).fill({a: 1})`
Actual results: `Array [ {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, … ]`
Expected results: To be determined

julienw:
as a first (easy ?) step I think this is worth showing the length, either explicitely (chrome shows as a prefix) or implicitely (replacing the last `...` by `and XXX more`).

kootoopas:
@nchevobbe I'm currently working on this feature. Got the functionality down and I'm updating the `grip*` tests to be in sync with the length title postfix.

<img width="365" alt="screen shot 2017-12-15 at 2 53 01 pm" src="https://user-images.githubusercontent.com/601001/34049784-5669c20a-e1c1-11e7-9759-1d5554ac6fc9.png">

jasonLaster:
Nice @kootoopas!

nchevobbe:
this landed in #881

PR #881, opened 2018-01-05
Append length to arraylike titles

kootoopas:
*(this is a continuation of #859, since I rebased)*

Alrighty. Take `2`; action! 👏

Associated Issue: #566 

### Summary of Changes

Append length in title of arraylike objects (Arrays, Sets, etc).

screen shot 2017-12-16 at 5 38 21 pm
screen shot 2017-12-16 at 6 22 25 pm

### Test Plan

- Click the `array`, `map & sets` and`promise` buttons.


nchevobbe:
A couple of things : 
- if we do this, we should have the length for the Maps as well
- it can be a bit redundant on small arrays, e,g, `[1]` will be rendered as `Array (1) [ 1 ]` in short and long mode. It may be a bit confusing. Looking at chrome, they don't show it for 1-item long elements
- i find it a bit too prominent. But that may be personal taste. Could we drop the parens, and maybe style it with a background color and a small border-radius (to match the "repeat bubble" we have in the console, but not as strong). It would be nice to play with it a bit

kootoopas:
Oh my, I totally forgot about Maps. 😅 

I agree that it's an overkill for arrays whose length is obvious by the preview items.

As you said, we could experiment with color hues and contrast of the length bubble. After all, the devtools should attempt to break the status quo now and then!

kootoopas:
A little update on the situation:
* [x] length shows maps
* [x] length is hidden when it's easily derivable from preview items
* [ ] style bubble
* [ ] update tests

kootoopas:
Sooo, I basically replicated the console repetition bubble and trimmed the horizontal padding a bit since there's other associated text on the left and right of it:
screen shot 2018-01-13 at 6 15 38 am

Here's another one that uses `--number-color`. This one pops more:
screen shot 2018-01-13 at 6 25 23 am

By the way, wouldn't it make sense for maps to show `{...}` in tiny mode like arrays do (`[...]`)?

nchevobbe:
Thanks a lot @kootoopas . 
So, this is a bit intense to my taste :) @violasong what do you think of the screenshot from https://github.com/devtools-html/devtools-core/pull/881#issuecomment-357410597 .
cc @julienw since we was the one asking for better display of Array length.

---

> By the way, wouldn't it make sense for maps to show {...} in tiny mode like arrays do ([...])?

If we do add the length next to the title, I don't think it is necessary (it may not for Sets either). 
The TINY mode needs to be the more terse possible (while being useful). Here, `Map 2000 { ... }` does not add any value over `Map 2000`. I agree it would be the case with the current status (i.e. no length indicator).

julienw:
I think the view from the initial description is lighter and less intrusive.
Maybe we don't need it in the "long" view though, as we have the "length" property already. Although I wouldn't mind.

I still like my previous proposal: `["1", "1", "1", and 997 more]` with "997" standing out. This feels more "human". But knowing directly the absolute length is useful as a pragmatic way.

kootoopas:
@nchevobbe It's too intense indeed!
@julienw I think that not breaking the `--object-color` scheme is the right approach too.

So with the following I think we're getting somewhere. Since it's more subtle and also in the outlinish style of the devtools UI (🤞 🙏):
screen shot 2018-01-15 at 6 51 10 pm

screen shot 2018-01-15 at 6 54 20 pm

>The TINY mode needs to be the more terse possible (while being useful). Here, Map 2000 { ... } does not add any value over Map 2000.

@nchevobbe I can work on making the tiny mode terse by omitting `{...}` from Sets and Maps and the likes of them when it makes sense, but first we need to resolve if we're going to have a length bubble at all. :)

violasong:
I'm finding that because the Console output is so code-like, the first version with parentheses seems the most obvious to me — it's similar to the syntax for initializing an array with certain length — but maybe this is a point against it, since it isn't actually code. The nice "badge"-looking versions make me think there's some other kind of activity going on. Curious what you all think of this.

kootoopas:
I personally don't mind the no-badge traditional length depiction either, but it seems like a nice use-case to experiment a bit since badge elements are used to represent all sorts of counts (although mainly notifications).

nchevobbe:
okay, let's see if we can go with the parens then :) @kootoopas could you try it, and apply the same color we have for the "__proto__" node (grey) ?

kootoopas:
@nchevobbe Done :)

In tiny mode, should we omit the `[...]` from non-array arraylikes (like Sets) since the title w/ the length bubble (f.e. `Set(3)`) provide sufficient info?

nchevobbe:
Yes, it would be nice :) Thanks a lot !

Le 20 janv. 2018 07:06, "Fotis Koutoupas" <notifications@github.com> a
écrit :

kootoopas:
@nchevobbe Alright. All that's left to do now is sync the tests and this will probably be done. :D

By the way, should I color the length's parens with `--object-color`? Curly and normal brackets are colored as such.

screen shot 2018-01-22 at 5 38 57 pm
screen shot 2018-01-22 at 5 40 19 pm
screen shot 2018-01-22 at 5 40 32 pm

kootoopas:
@nchevobbe I actually prefer it as is too, since it emphasizes distinction between the object name and its length.

violasong:
Thank you @kootoopas! Looks good!

packages/devtools-reps/src/reps/grip-array.js
```diff
@@ -229,4 +266,5 @@ module.exports = {
   supportsObject,
   maxLengthMap,
   getLength,
+  getMoreEllipsisElement,
```
nchevobbe: why do we need to export this function ? It's already imported from utils.

packages/devtools-reps/src/reps/grip-map.js
```diff
@@ -105,14 +129,10 @@ function entriesIterator(props, object, max) {
...
-    entries.push(span({
-      key: "more",
-      className: "more-ellipsis",
-      title: "more…"
-    }, "…"));
+    entries.push(getMoreEllipsisElement());
   }
 
-  return unfoldEntries(entries);
+  return unfoldEntries(entries, props.mode);
```
nchevobbe: we make no use of the mode in `unfoldEntries`. Do I miss something ?

packages/devtools-reps/src/reps/grip-map.js
```diff
@@ -63,10 +80,17 @@ function GripMap(props) {
 }
 
 function getTitle(props, object) {
-  let title = props.title || (object && object.class ? object.class : "Map");
+  const title = props.title || (object && object.class ? object.class : "Map");
   return span({
-    className: "objectTitle",
-  }, title);
+      className: "objectTitle"},
```
nchevobbe: nit: indent

packages/devtools-reps/src/reps/rep-utils.js
```diff
@@ -387,6 +387,19 @@ function isURL(token) {
   }
 }
 
+/**
+ * Gets the ellipsis element.
+ *
+ * @return React.Element
+ */
+function getMoreEllipsisElement() {
```
nchevobbe: Thanks for putting this in a util function, the code looks better.
Maybe we could even have the element in a constant ? Like

```js
const ellipsisElement = span({
  title: `more${ELLIPSIS}`
}, ELLIPSIS);

module.exports = {
  ...,
  ellipsisElement
}
```

Since the element is always the same, we save a bit of time by not having to re-create it for each Reps that need it.

packages/devtools-reps/src/reps/tests/grip-array.js
```diff
@@ -57,18 +59,23 @@ describe("GripArray - max props", () => {
   it("renders as expected", () => {
     const renderRep = (props) => shallowRenderRep(object, props);
 
-    const defaultOutput = `Array [ 1, "foo", {} ]`;
+    let length = getGripLengthBubbleText(object);
```
nchevobbe: i'd rather not use the `getGripLengthBubbleText` function and hardcode the bubble in each test.
Here, if at some point `getGripLengthBubbleText` returns `""`, the test will pass although it should fail.

packages/devtools-reps/src/reps/tests/grip-array.js
```diff
@@ -412,22 +471,24 @@ describe("GripArray - Set", () => {
     component = renderRep({ mode: MODE.TINY });
     // FIXME: Tiny mode should render Set as "Set" and not "[…]".
```
nchevobbe: not in your patch, but here I think we can remove the comment. This has been fixed.

packages/devtools-reps/src/reps/tests/grip-array.js
```diff
@@ -494,15 +559,15 @@ describe("GripArray - WeakSet", () => {
     component = renderRep({ mode: MODE.TINY });
     // FIXME: Tiny mode should render WeakSet as "WeakSet" and not "[…]".
```
nchevobbe: we can remove this comment, the issue has been fixed

packages/devtools-reps/src/reps/tests/grip-map-entry.js
```diff
@@ -48,41 +49,61 @@ describe("GripMapEntry - createGripMapEntry", () => {
 describe("GripMapEntry - complex", () => {
   it("Handles complex objects as key and value", () => {
     let stub = gripArrayStubs.get("testBasic");
+    let length = getGripLengthBubbleText(stub);
     let entry = createGripMapEntry("A", stub);
-    expect(renderRep(entry, MODE.TINY).text()).toEqual("A → []");
-    expect(renderRep(entry, MODE.SHORT).text()).toEqual("A → Array []");
-    expect(renderRep(entry, MODE.LONG).text()).toEqual("A → Array []");
+    expect(renderRep(entry, MODE.TINY).text()).toEqual(`A → []`);
+    expect(renderRep(entry, MODE.SHORT).text()).toEqual(`A → Array${length} []`);
```
nchevobbe: same comment, hardcode the bubble instead of using `getGripLengthBubbleText`

packages/devtools-reps/src/reps/tests/grip-map.js
```diff
@@ -13,6 +14,7 @@ const stubs = require("../stubs/grip-map");
 const {
   expectActorAttribute,
   getSelectableInInspectorGrips,
+  getMapLengthBubbleText,
```
nchevobbe: ditto about not using `getMapLengthBubbleText`

packages/devtools-reps/src/reps/tests/grip.js
```diff
@@ -15,6 +16,7 @@ const gripArrayStubs = require("../stubs/grip-array");
 const {
   expectActorAttribute,
   getSelectableInInspectorGrips,
+  getGripLengthBubbleText
```
nchevobbe: Do not use `getMapLengthBubbleText`

packages/devtools-reps/src/reps/tests/promise.js
```diff
@@ -13,7 +13,8 @@ const { MODE } = require("../constants");
 const stubs = require("../stubs/promise");
 const {
   expectActorAttribute,
-  getSelectableInInspectorGrips
+  getSelectableInInspectorGrips,
+  getGripLengthBubbleText
```
nchevobbe: Do not use `getMapLengthBubbleText`

packages/devtools-reps/src/reps/grip-array.js
kootoopas: there's no need actually, it just used to be defined in `grip-array`. I'll remove this.

packages/devtools-reps/src/reps/grip-map.js
kootoopas: at some point i tried to calculate sth in there. Will remove.

packages/devtools-reps/src/reps/rep-utils.js
kootoopas: I thought of that too actually but I went with playing safe! :)

kootoopas: I think that we should keep `key` and `className`, since it's often used as an array item and its styling is pretty standard across the codebase, resulting to:
```javascript
const ellipsisElement = span({
    key: "more",
    className: "more-ellipsis",
    title: `more${ELLIPSIS}`
}, ELLIPSIS);
```

packages/devtools-reps/src/reps/tests/grip-array.js
kootoopas: @nchevobbe It may be kind of confusing at first sight, but to preemptively print the empty `""` length text makes the tests more future-proof. What if we change the `visibilityThreshold` prop for certain grips? What if we decide to change the `maxLengthMap` of `grip-array`/`grip-map` in the future? We'd need to update all tests by hand and omit or add `${length}` in certain expected outputs.

If we leave it as (calculate `${length}` for each mode of each stub regardless of empty text), we're having more dynamic test-cases, although not as straightforward as your proposal would make 'em. 🙂 

For reference, here are the current conditions that return `""`:
```javascript
  const isEmpty = length === 0;
  const isObvious = [MODE.SHORT, MODE.LONG].includes(mode) &&
    length <= maxLengthMap.get(mode) &&
    length <= visibilityThreshold;
  if (isEmpty || isObvious) {
    return "";
  }
```

kootoopas: please tell me if I'm missing something! 😅

nchevobbe: I was saying return `""` as an example, if something changes in the function that might be unwanted.
But I see your point. So I think I will be okay with that if we add tests for `getGripLengthBubbleText`. That way, we can still use it in test, and if something goes wrong, the test will fail.

packages/devtools-reps/src/reps/tests/grip-map.js
```diff
@@ -31,14 +33,14 @@ describe("GripMap - empty map", () => {
 
   it("renders as expected", () => {
     const renderRep = (props) => shallowRenderRep(object, props);
-    const defaultOutput = "Map {  }";
+    const defaultOutput = "Map {}";
```
nchevobbe: let's be consistent and have empty maps rendered as `Map(0)`

packages/devtools-reps/src/shared/grip-length-bubble.js
```diff
@@ -0,0 +1,43 @@
...
+  getLength: PropTypes.func.isRequired,
+  mode: ModePropType,
+  visibilityThreshold: PropTypes.number
+};
+
+function GripLengthBubble(props) {
+  const {
+    object,
+    mode = MODE.SHORT,
+    visibilityThreshold = 5,
```
nchevobbe: i think the threshold should be 2, because some objects can be quote lengthy and hard to count (e.g. `new Map([[document, location]])`)

packages/devtools-reps/src/shared/grip-length-bubble.js
```diff
@@ -19,17 +19,19 @@ function GripLengthBubble(props) {
   const {
     object,
     mode = MODE.SHORT,
-    visibilityThreshold = 5,
+    visibilityThreshold = 2,
```
kootoopas: ✅ Visibility changed

jasonLaster:
Yay! Great work @kootoopas!

kootoopas:
@jasonLaster @nchevobbe Anytime guys, anytime... 😎 

@nchevobbe Thanks for guiding me throughout this!

PR #884, opened 2018-01-09
[WIP] Move the logic to load properties of a node to util file.

nchevobbe:
This will allow a consumer (e.g. the debugger) to preload properties of a given root
and use them as it wishes (e.g., for the debugger, use all the direct properties as
root for the ObjectInspector of the Popup preview).

### DIFF
--- a/packages/devtools-reps/src/object-inspector/utils/node.js
+++ b/packages/devtools-reps/src/object-inspector/utils/node.js
@@ -29,12 +29,21 @@ const NODE_TYPES = {
   PROTOTYPE: Symbol("__proto__"),
 };
 
+const {
+  enumEntries,
+  enumIndexedProperties,
+  enumNonIndexedProperties,
+  getPrototype,
+  enumSymbols,
+} = require("./client");
+
 import type {
   CachedNodes,
   GripProperties,
   LoadedProperties,
   Node,
   NodeContents,
+  ObjectClient,
   RdpGrip,
 } from "../types";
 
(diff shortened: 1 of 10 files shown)


nchevobbe:
WIP because I need to write the debugger code (which I'll do right away) that would make use of that and validate my assumptions.

cc @jasonLaster

packages/devtools-reps/src/object-inspector/utils/node.js
```diff
@@ -762,6 +771,69 @@ function getClosestNonBucketNode(item: Node) : Node | null {
   return getClosestNonBucketNode(parent);
 }
 
+function loadItemProperties(
```
jasonLaster: this function could probably be in its own module. It is sufficiently interesting in terms of dependencies and logic

```diff
  
+ const {
+   enumEntries,
+   enumIndexedProperties,
+   enumNonIndexedProperties,
+   getPrototype,
+   enumSymbols,
+ } = require("./client");
+ 

```

nchevobbe: what do you mean by own module ?

jasonLaster: it's own file

nchevobbe: sure

nchevobbe:
okay, this is working in the debugger, even if it needs a bit too much work to my taste.

PR #889, opened 2018-01-10
Update babel-eslint and bump launchpad to 0.0.114

nchevobbe:
This is likely to fix https://github.com/devtools-html/debugger.html/issues/5068 (see https://github.com/eslint/eslint/issues/9767#issuecomment-353840674)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "dependencies": {
     "babel-cli": "^6.7.5",
     "babel-core": "^6.17.0",
-    "babel-eslint": "^7.1.0",
+    "babel-eslint": "^8.2.1",
     "babel-plugin-transform-async-to-generator": "^6.16.0",
     "babel-plugin-transform-flow-strip-types": "^6.14.0",
     "eslint": "^3.12.0",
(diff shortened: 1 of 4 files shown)


nchevobbe:
CI is failing because of devtools-license-check complaining about pako being MIT **and** Zlib.
pako is not requested directly and is probably a dependency of babel-eslint.

I'm unsure what should be done here.

nchevobbe:
[Zlib license](https://spdx.org/licenses/Zlib.html) seems pretty good for us. Should we add it to devtools-license-check valid licences ? @captainbrosset do you have an opinion on this ?

codehag:
good question -- im not too sure how to review this. pbro and @jasonLaster probably have a good idea though. I don't know if we need to clear this with any legal agents... would be good to know!

captainbrosset:
If this is a devDependency, then we should be fine, no need to check anything, we're not shipping or embedding code anywhere. And looking at the mentioned issue, I guess this is or should be a devDependency.

It it's not, then that means we're distributing Pako with our software and should therefore make sure its license is a compatible one.
It looks like it is, so feel free to open a PR against the license-checker and ask Tom for a quick review.

nchevobbe:
So it is a direct dependency at the moment, in the devtools-launchpad package (which we publish).
@jasonLaster is there any reason to keep it as a dependency and not a dev one ?
Also, can't we have this dependency as a devDependency on the debugger instead ?

jasonLaster:
we can make it a devDependency.

The debugger now has devDependencies for the eslint plugins

nchevobbe:
the dependency was added in debugger and now everything is fine (after releasing a new version of devtools-licenes-check to support Zlib).
Let's put that aside for now

PR #891, opened 2018-01-10
Add ThreadClient methods for rewinding and reverse-stepping a thread.

bhackett1024:
Rewinding can be used with tabs that are replaying via Web Replay, mozilla bug 1422587.


bhackett1024:
What else needs to be done to merge these changes?  The automated tests have failed but the failures looks unrelated to this patch.

jasonLaster:
CC @nchevobbe -- looks like reps tests

When we merge this I'll bump the version and land it in debugger.html

nchevobbe:
Yep, they all have the same error : `Cannot find module 'asn1'`

nchevobbe:
looks like everything is fine on CI now :)

PR #892, opened 2018-01-10
Add Zlib into valid licenses.

nchevobbe:
Zlib is used by the packo packages, which is a dependency of babel-eslint,
of which we need the latest version on the debugger.
The [Zlib license](https://spdx.org/licenses/Zlib.html) seems pretty safe to
add in the list of valid licenses here.


nchevobbe:
Thanks @tromey . Should we publish a new package version when this PR is merged ?

tromey:
> Should we publish a new package version when this PR is merged ?

Yes, I think that would make sense.

PR #895, opened 2018-01-10
Update devtools-components to 0.1.4

nchevobbe:


nchevobbe:
r=me for simple dependency bump

PR #896, opened 2018-01-11
Expose all ObjectInspector utils.

nchevobbe:
This is needed for the debugger to be able to fetch properties on its own.


nchevobbe:
thanks @codehag

PR #898, opened 2018-01-11
Prevent expanding an object when a string is selected.

### LINKED ISSUE #885: Do not toggle object inspector nodes when selecting text
1. Open the console
2. Enter `inspect({a: () => {}})`
```
▼ {…}
  ▶ a: function a()
​  ▶ __proto__: Object { … }
```
3. Select `fun`

Before you can e.g. copy it, the property is expanded and `fun` disappears:
```
▼ {…}
  ▼ a: a()
     length: 0
     name: "a"
​  ▶ __proto__: Object { … }
```

weilonge:
Fix #885

This patch is going to fix an unnecessary object expanding when a string is selected.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/index.js
+++ b/packages/devtools-reps/src/object-inspector/index.js
@@ -368,6 +368,10 @@ class ObjectInspector extends Component {
         }),
         onClick: e => {
           e.stopPropagation();
+          const selection = getSelection();
+          if (selection && selection.toString()) {
+            return;
+          }
           if (isPrimitive === false) {
             this.setExpanded(item, !expanded);
           }


weilonge:
I need to fix `ReferenceError: getSelection is not defined` in test and add a test to verify the new behavior.

weilonge:
I am also thinking if there is any other case needs this fix. e.g. `onLabelClick`

Loirooriol:
Why `selection.toString()` instead of `selection.isCollapsed`? You just want a boolean. And is `toString` even standard? I can't find it in the Selection API spec.

nchevobbe:
Good point about onLabelClick
I don't think it is used somewhere but since we offer this ability, it should be bug free :)

nchevobbe:
Closing in favor of https://github.com/devtools-html/devtools-core/pull/925

PR #900, opened 2018-01-12
Update README.md links

ncalexan:
I used relative links, so there seems no advantage to pinning the branch.


nchevobbe:
Thanks @ncalexan !

PR #905, opened 2018-01-18
Flush nodes cache when roots prop is updated.

### LINKED ISSUE #904: ObjectInspector is not updated if roots changed but kept same path
Steps to reproduce:
1. Set an ObjectInspector with a `roots` object
2. Expand it
3. Update the `roots` object to another object, but keep the same path value

Expected results:
The ObjectInspector is updated and shows the new `roots` children

Actual results:
The ObjectInspector is not updated.

This is due to the fact that we cache nodes, and never clear the cache. The `getChildren` then uses this cache and find matching nodes since the path stayed the same.

---

This issue is resolved when the following test pass: 
```diff
diff --git a/packages/devtools-reps/src/object-inspector/tests/component/basic.js b/packages/devtools-reps/src/object-inspector/tests/component/basic.js
index bf27ca0..80800e8 100644
--- a/packages/devtools-reps/src/object-inspector/tests/component/basic.js
+++ b/packages/devtools-reps/src/object-inspector/tests/component/basic.js
@@ -261,4 +261,40 @@ describe("ObjectInspector - renders", () => {
     }]});
     expect(formatObjectInspector(oi)).toMatchSnapshot();
   });
+
+  it("updates when the root changes but has same path", () => {
+    let oi = mount(ObjectInspector(generateDefaults({
+      roots: [{
+        path: "root",
+        name: "root",
+        contents: [{
+          name: "a",
+          contents: {
+            value: 30,
+          }
+        }, {
+          name: "b",
+          contents: {
+            value: 32,
+          }
+        }]
+      }],
+      mode: MODE.LONG,
+    })));
+    oi.find(".node").at(0).simulate("click");
+
+    const oldTree = formatObjectInspector(oi);
+    oi.setProps({roots: [{
+        path: "root",
+        name: "root",
+        contents: [{
+          name: "c",
+          contents: {
+            value: "i'm the new node",
+          }
+        }]
+    }]});
+
+    expect(formatObjectInspector(oi)).not.toBe(oldTree);
+  });
 });

```

nchevobbe:
Fixes #904.
Add a test to make sure we don't regress this.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/index.js
+++ b/packages/devtools-reps/src/object-inspector/index.js
@@ -162,8 +162,15 @@ class ObjectInspector extends Component {
       expandedPaths,
       loadedProperties
     } = this.state;
-    return this.props.roots !== nextProps.roots
-      || expandedPaths.size !== nextState.expandedPaths.size
+
+    if (this.props.roots !== nextProps.roots) {
+      // Since the roots changed, we assume the properties did as well. Thus we can clear
+      // the cachedNodes to avoid bugs and memory leaks.
+      this.cachedNodes.clear();
+      return true;
+    }
+
+    return expandedPaths.size !== nextState.expandedPaths.size
       || loadedProperties.size !== nextState.loadedProperties.size
       || [...expandedPaths].some(key => !nextState.expandedPaths.has(key));
   }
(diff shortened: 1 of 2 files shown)


nchevobbe:
@jasonLaster thanks to this change, I now have only 1 failing mochitest on https://github.com/devtools-html/debugger.html/pull/4915

PR #906, opened 2018-01-18
Introduce getStackTrace method for webconsole client (sync with m-c)

janodvarko:
This PR add `getStackTrace` method for fetching stack-trace info from the back-end.  This method exists in m-c for some time, but wasn't synced with devtools-connection package.

Honza


janodvarko:
The original `WebConsoleClient.getStackTrace` method has been introduced here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1416201
Honza

PR #911, opened 2018-01-25
Allow to disable postCSS loader

nchevobbe:
This is useful for projects using the launchpad but with no need for postCSS (devtools-reps, console, netmonitor, …)

### DIFF
--- a/packages/devtools-launchpad/webpack.config.js
+++ b/packages/devtools-launchpad/webpack.config.js
@@ -7,7 +7,6 @@ require("babel-register");
 const path = require("path");
 
 const webpack = require("webpack");
-const { NormalModuleReplacementPlugin } = require("webpack");
 
 const ExtractTextPlugin = require("extract-text-webpack-plugin");
 const {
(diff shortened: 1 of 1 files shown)


packages/devtools-launchpad/webpack.config.js
```diff
@@ -93,27 +93,29 @@ module.exports = (webpackConfig, envConfig, options = {}) => {
...
+      loader: "style-loader"
+    }, {
+      loader: "css-loader",
+      options: {
+        importLoaders: 1,
+        url: false
+      }
+    }];
+
+    if (options.disablePostCSS !== true) {
```
jasonLaster: NIT, i'm not sure if we have a consensus here, but I'd prefer `!!` when it's either a boolean or undefined. Other cases where the type is overloaded should be more verbose

`if (options.disablePostCSS !== true) {` => `if (!!options.disablePostCSS) {`

nchevobbe: done

PR #919, opened 2018-01-31
Fix indentation of non-expandable nodes

MikeRatcliffe:
Associated Issue: https://github.com/devtools-html/debugger.html/issues/5250

The indentation of non-expandable nodes was too large.

### Summary of Changes

```diff
 .tree-node[data-expandable="false"] .tree-indent:last-of-type {
-  margin-inline-end: 15px;
+  margin-inline-end: 4px;
 }
```

### DIFF
--- a/packages/devtools-components/src/tree.css
+++ b/packages/devtools-components/src/tree.css
@@ -44,7 +44,7 @@
 
 /* Align with expandables siblings (where we have the arrow) */
 .tree-node[data-expandable="false"] .tree-indent:last-of-type {
-  margin-inline-end: 15px;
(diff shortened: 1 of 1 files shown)


nchevobbe:
let's look how this looks in storybook.
Note that the issue in debugger might be because it is not using the arrow the tree provides :/

nchevobbe:
here's the difference: 
before/after

I think the current indent is working well (at least for the object inspector), so we should apply the fix either on the debugger CSS or change the ObjectInspector css if we make this change here.

MikeRatcliffe:
Okay, let's do it in the debugger.

PR #921, opened 2018-02-02
Changed fetchSourceMap to reject instead of resolve when sourceMapURL is not found

AlexisDeschamps:
This is a result of the discussion here: https://github.com/devtools-html/debugger.html/pull/1937#issuecomment-330414299.

The end goal is to better allow failures when loading of source maps to clear the sourceMapURL in order to enable pretty printing: Relied on by: https://github.com/devtools-html/debugger.html/pull/5296.

Associated Issue: https://github.com/devtools-html/debugger.html/issues/1661

### Summary of Changes

* Changed fetchSourceMap to reject instead of resolve when sourceMapURL is not found

### DIFF
--- a/packages/devtools-source-map/src/utils/fetchSourceMap.js
+++ b/packages/devtools-source-map/src/utils/fetchSourceMap.js
@@ -64,7 +64,7 @@ function fetchSourceMap(generatedSource: Source) {
   }
 
   if (!generatedSource.sourceMapURL) {
-    return Promise.resolve(null);
+    return Promise.reject();
   }
 
   // Fire off the request, set it in the cache, and return it.


AlexisDeschamps:
@tromey Do you think this could be a valid change?

tromey:
I am not sure about this patch.  I think maybe I don't understand the rationale for it.

Landing this would require patching the toolbox as well somehow.  See https://dxr.mozilla.org/mozilla-central/rev/35dfa0882568c06f2d81b6749179815cd4f82886/devtools/client/framework/toolbox.js#602-611

In this case, rejecting would be a non-error case, which seems weird.  Why wouldn't checking for a null return work?  Or checking the input for a missing sourceMapURL before even calling this at all?

I read https://github.com/devtools-html/debugger.html/pull/5296/files but I still don't understand, because I don't see what exactly this change would affect there.  I guess I'd expect the promise to resolve to null, then the null result would be returned by `loadOriginalSourceUrls`; and then wouldn't `loadSourceMap` do the right thing?  Since the `sourceMapURL` is missing anyway?

Also in that patch I wonder if the error-handling behavior in the toolbox is going to make this not work in M-C.

jasonLaster:
Thanks @tromey, i'll take a look at the `null` handling on the debugger.html side and see how this is affected by the toolbox

AlexisDeschamps:
OK, thanks. Closing this for now as the change should not be needed.

tromey:
Sorry if I wasn't clear.  I'm not actually sure that the change is not needed -- I just didn't understand why it was.

PR #925, opened 2018-02-08
Prevent expanding an object when a string is selected

### LINKED ISSUE #885: Do not toggle object inspector nodes when selecting text

nchevobbe:
_Previously at https://github.com/devtools-html/devtools-core/pull/898 , finishing it since Sean might not have the time to do so_

---

Fix #885 

Prevent expanding an object when text is selected.
Add tests to make sure we don't expand a node, or trigger `onLabelClick`, when the
user selected some text.
Add a mock for `window.getSelection` since it is not defined in nodejs.

### Test Plan

- In `devtools-core/packages/devtools-reps` do `yarn install && yarn start` 
- open localhost:8000.
- Launch a new Firefox window
- select the tab in the launchpad
- evaluate `location`
- select the location. **It is important your mouseup is fired on the text itself**
- assert that the node wasn't expanded
- click on the node
- assert that the node was expanded

### DIFF
--- a/packages/devtools-reps/src/object-inspector/index.js
+++ b/packages/devtools-reps/src/object-inspector/index.js
@@ -384,6 +384,10 @@ class ObjectInspector extends Component {
         }),
         onClick: e => {
           e.stopPropagation();
+          const selection = getSelection();
+          if (selection && selection.toString()) {
+            return;
+          }
           if (isPrimitive === false) {
             this.setExpanded(item, !expanded);
           }
--- a/packages/devtools-reps/src/object-inspector/index.js
+++ b/packages/devtools-reps/src/object-inspector/index.js
@@ -384,8 +384,8 @@ class ObjectInspector extends Component {
         }),
         onClick: e => {
           e.stopPropagation();
-          const selection = getSelection();
-          if (selection && selection.toString()) {
+          // If the user selected text, bail out.
+          if (Utils.selection.documentHasSelection()) {
             return;
           }
           if (isPrimitive === false) {
@@ -411,6 +411,12 @@ class ObjectInspector extends Component {
             onClick: onLabelClick
               ? event => {
                 event.stopPropagation();
+
+                // If the user selected text, bail out.
+                if (Utils.selection.documentHasSelection()) {
+                  return;
+                }
+
                 onLabelClick(item, {
                   depth,
                   focused,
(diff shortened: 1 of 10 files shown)


nchevobbe:
@Loirooriol  do you mind having a look since you commented on the previous PR ?
I used Selection.type since it's the most supported property from what I saw on MDN

Loirooriol:
Comparing http://help.dottoro.com/ljitmswc.php and http://help.dottoro.com/ljlflvjl.php I'm not sure `Selection.type` is more supported, but at least it appears in the spec.

I get various errors like `Module not found: Error: Can't resolve 'react/lib/getNextDebugID'` when I use `yarn install && yarn start` so I can't test, but the patch seems OK to me.

packages/devtools-reps/src/object-inspector/index.js
```diff
@@ -407,6 +412,13 @@ class ObjectInspector extends Component {
             onClick: onLabelClick
               ? event => {
                 event.stopPropagation();
+
+                const selection = getSelection();
+                // If the user selected text, bail out.
+                if (selection && selection.type === "Range") {
+                  return;
+                }
+
```
janodvarko: nit: you might introduce a new `this.getSelection` method that wraps the logic.

packages/devtools-reps/src/object-inspector/tests/component/events.js
```diff
@@ -80,4 +80,32 @@ describe("ObjectInspector - properties", () => {
 
     expect(onLabelClick.mock.calls.length).toBe(1);
   });
+
+  it("does calls the onLabel prop function when the user selected text", () => {
```
janodvarko: typo: does call the ...

nchevobbe: fixed

packages/devtools-reps/src/object-inspector/index.js
nchevobbe: good idea, I created a util for that (`documentHasSelection`)

PR #930, opened 2018-02-09
force a state change when changing the initial measurements

ericawright:
In conjunction with this PR to debugger: https://github.com/devtools-html/debugger.html/pull/5383

### Summary of Changes

* force a state change when changing the initial measurements

This is to ensure that when we render another SplitBox in the same location, it actually get initialized as desired.
Another solution could be to use react's `key` property, perhaps passed into props, so that react can tell that it is a different element.

### DIFF
--- a/packages/devtools-splitter/src/SplitBox.js
+++ b/packages/devtools-splitter/src/SplitBox.js
@@ -81,6 +81,14 @@ class SplitBox extends Component {
     if (this.props.vert !== nextProps.vert) {
       this.setState({ vert: nextProps.vert });
     }
+    if (this.props.initialSize !== nextProps.initialSize
+      || this.props.initialWidth !== nextProps.initialWidth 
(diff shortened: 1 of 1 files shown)


ericawright:
@nchevobbe by "story" do you mean a comment explaining it?

nchevobbe:
Oops, sorry, I should have been more explicit.

So in this project, and some others, we have a utility called [storybook](https://github.com/storybooks/storybook), which allow us to render the components with different props, so we have a visual way of checking the components.

The way you start it is by doing: 
```bash
cd packages/devtools-splitter
yarn install
yarn storybook
```

and then navigating to the url you're prompted.

You can see the stories here: https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-splitter/stories/index.js

I just tried to add the comment here, and it looks broken, so don't bother for now.
I'll try to fix it soon and I'll add the story, or I'll try to add jest tests.

Anyway, thanks for the patch Erica !

PR #932, opened 2018-02-14
Move the url opened when clicking the launch firefox button to config

nchevobbe:
It was hardcoded in launch.js and was well suited for the debugger,
but other tools might want to see other default pages.


nchevobbe:
@jasonLaster since you're around, does that sounds good to you ? https://github.com/devtools-html/devtools-core/pull/932

jasonLaster:
yep - seems reasonable :)

PR #935, opened 2018-02-17
Add arrow spacing in right pane for Chrome

tomodwyer:
Associated Issue: https://github.com/devtools-html/debugger.html/issues/5457

### Summary of Changes

* Chrome doesn't support `margin-line-end` so `-webkit-margin-end: 5px;` was added to `.arrow svg`

### Test Plan

- Start debugger.html
- Open in Chrome
- Open a new tab in Firefox
- Attach debugger.html to Firefox
- View sidebar arrow with correct spacing

### Screenshot

screen shot 2018-02-17 at 15 41 19

### DIFF
--- a/packages/devtools-components/src/tree.css
+++ b/packages/devtools-components/src/tree.css
@@ -75,6 +75,7 @@
   transition: transform 0.125s ease;
   width: 10px;
   margin-inline-end: 5px;
+  -webkit-margin-end: 5px;
   transform: rotate(-90deg);
 }
 
--- a/packages/devtools-components/src/tree.css
+++ b/packages/devtools-components/src/tree.css
@@ -15,10 +15,6 @@
 }
 
 .tree.noselect {
-  -webkit-user-select: none;
-  -moz-user-select: none;
-  -ms-user-select: none;
-  -o-user-select: none;
   user-select: none;
 }
 
@@ -75,7 +71,6 @@
   transition: transform 0.125s ease;
   width: 10px;
   margin-inline-end: 5px;
-  -webkit-margin-end: 5px;
   transform: rotate(-90deg);
 }
 
--- a/packages/devtools-components/postcss.config.js
+++ b/packages/devtools-components/postcss.config.js
@@ -3,4 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // Dumb export so launchpad does not throw.
-module.exports = () => ({});
+module.exports = () => {
+  return {
+    plugins: [require("autoprefixer")]
+  };
+};


packages/devtools-components/src/tree.css
```diff
@@ -75,6 +75,7 @@
   transition: transform 0.125s ease;
   width: 10px;
   margin-inline-end: 5px;
+  -webkit-margin-end: 5px;
```
jasonLaster: we shouldnt need -webkit because we're using postcss autoprefix

nchevobbe: so, if the debugger is using postCSS, what's the issue here ?

tomodwyer: Doing some more testing (hopefully this is useful):

If I add a `postcss.config.js` file:  

```js
module.exports = ({ file, options, env }) => {
  return {
    plugins: [
      require("autoprefixer")({
        browsers: ["last 2 Firefox versions", "last 2 Chrome versions"],
        flexbox: false,
        grid: false
      })
    ]
  };
};
```

into `debugger.html/node_modules/devtools-components/src` and then delete the  
`debugger.html/node_modules/devtools-components/postcss.config.js` and restart the app, PostCSS does indeed add the -webkit prefix.

nchevobbe: so, it's about the config in the debugger not targeting CSS in node_modules right ?
Could this be resolved on the debugger side ?

tomodwyer:
So I did some testing this weekend and these are my findings:

If you remove the [PostCSS config](https://github.com/devtools-html/devtools-core/blob/b83e8d0759130f6d46e593c59530abfacd556cf4/packages/devtools-components/postcss.config.js) from devtools-components - the following error is shown:
```sh
ERROR in ./node_modules/css-loader?{"importLoaders":1,"url":false}!./node_modules/postcss-loader/lib!../devtools-core/packages/devtools-components/src/tree.css
Module build failed: Error: No PostCSS Config found in: /Users/[redacted]/Code/devtools-core/packages/devtools-components/src
```

There is an [included comment](https://github.com/devtools-html/devtools-core/blob/b83e8d0759130f6d46e593c59530abfacd556cf4/packages/devtools-components/postcss.config.js#L5) which explains this is launchpad blowing up:
`// Dumb export so launchpad does not throw.`

Therefore, if you add [Autoprefixer to the PostCSS config](https://github.com/tomodwyer/devtools-core/blob/0547209a9a062e11406edb0f3c4e0d8a61f8763e/packages/devtools-components/postcss.config.js#L8) for devtools-components, debugger.html adds in the right browser prefixes and everything seems to work.

I've therefore added [another commit](https://github.com/devtools-html/devtools-core/pull/935/commits/0547209a9a062e11406edb0f3c4e0d8a61f8763e) which adds that to the config and removed the unnecessary prefixed properties in tree.css

nchevobbe:
let's fix this at the debugger level

PR #937, opened 2018-02-19
Reps code cleanup

nchevobbe:
I initiated this work in order to find areas where we could use fragments, but it seems like we are already careful on what we create.
There were only a few places where I was able to remove unnecessary nodes.
I took this as an opportunity to clean up the code a bit.

### DIFF
--- a/packages/devtools-reps/src/reps/rep-utils.js
+++ b/packages/devtools-reps/src/reps/rep-utils.js
@@ -387,6 +387,22 @@ function isURL(token) {
   }
 }
 
+/**
+ * Returns a new array in which `char` are interleaved between the original items.
+ *
+ * @param {Array} items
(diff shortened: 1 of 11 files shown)


packages/devtools-reps/src/reps/rep-utils.js
```diff
@@ -387,13 +387,30 @@ function isURL(token) {
...
+ * @returns Array
+ */
+function interleave(items, char) {
+  return items.reduce((res, item, index) => {
+    if (index !== items.length - 1) {
+      return res.concat(item, char);
+    }
+    return res.concat(item);
+  }, []);
+}
```
jasonLaster: This way works, I've seen `zip` used here `zip(items, delimiters)` where del is one less than items.

Issue #452, opened 2017-06-14
Apply sourcemaps and formatting to the stacks on the Error rep 


gregtatum:
<img width="983" alt="image" src="https://user-images.githubusercontent.com/1588648/27146615-c660da6c-50ee-11e7-9568-d84dd04e01f8.png">

When I throw a `console.error` at an Error object, the Error rep shows the raw text of the stack trace taken from an Error object. This stack is raw text, and does not include source maps, and we can't click into sources in that text. This gives me the impression that source maps is not working, and I can't get access to the information I need in my app.

So it seems like we should take that text, parse it, display it nicely, and spit out the sourcemapped functions and locations.

/cc @clarkbw @bgrins

nchevobbe:
Kind of related to https://github.com/devtools-html/devtools-core/issues/448 .
We should be able to parse the error string to recreate the callstack, and then use the usual Frame component, which handle sourcemap.

tromey:
The parsing approach won't work in all cases, because a script can be GC'd and then the source map URL won't be available.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1360252

tromey:
You can probably ignore my earlier comment, since that's just a generic problem that doesn't impact this project any more than it does anything else.

`error.js` mentions that it can't use localization due to https://bugzilla.mozilla.org/show_bug.cgi?id=1317038.  It seems like that's likely to be a problem with moving the stack and frame components to devtools-core as well.

nchevobbe:
The simplest way we can go right now would be to pass the [StackTrace component](https://searchfox.org/mozilla-central/source/devtools/client/shared/components/StackTrace.js) to the Reps when we create them.
In the console, it is done in the [GripMessageBodyComponent](https://searchfox.org/mozilla-central/rev/9d920555ec81f1c9d4b7fa3b08e23eb88efb60e1/devtools/client/webconsole/new-console-output/components/GripMessageBody.js#85-125).

We could then get it in the Error Rep and use it.
We'll only need to parse the `preview.stack` property to something usable for the StackTrace component.

nchevobbe:
Closing this as we have a better plan in https://github.com/devtools-html/devtools-core/projects/3

PR #942, opened 2018-02-20
Add prettier formatting for error rep's stacks

gregtatum:
image

image

### DIFF
--- a/packages/devtools-reps/src/launchpad/components/Console.css
+++ b/packages/devtools-reps/src/launchpad/components/Console.css
@@ -38,6 +38,11 @@ main {
   margin-bottom: 0.5rem;
 }
 
+.rep-element {
+  white-space: pre-wrap;
+  word-wrap: break-word;
+}
+
 .rep-element + .rep-element {
   margin-left: 0.5rem;
 }
--- a/packages/devtools-reps/src/launchpad/samples.js
+++ b/packages/devtools-reps/src/launchpad/samples.js
@@ -137,6 +137,36 @@ let samples = {
   symbol: [
     "Symbol('foo')",
     "Symbol()"
+  ],
+
+  errors: [
+    "throw new Error('This is a simple error message.');",
+    `
+      var error = new Error('Complicated error message');
+      error.stack =
+        "unserializeProfileOfArbitraryFormat@http://localhost:4242/2969802751c9e11c0c2d.bundle.js:26705:11\\n" +
+        "_callee7$@http://localhost:4242/2969802751c9e11c0c2d.bundle.js:27948:27\\n" +
+        "tryCatch@http://localhost:4242/2969802751c9e11c0c2d.bundle.js:64198:37\\n" +
+        "invoke@http://localhost:4242/2969802751c9e11c0c2d.bundle.js:64432:22\\n" +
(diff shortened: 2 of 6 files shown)


packages/devtools-reps/src/launchpad/components/Console.css
```diff
@@ -38,6 +38,11 @@ main {
   margin-bottom: 0.5rem;
 }
 
+.rep-element {
```
gregtatum: This aligns the styling to the webconsole's. The newlines were being ignored before.

jasonLaster:
Nice work @gregtatum!

couple of questions
1. what does it look like if you expand?
2. should we flip the pink/blue. i believe pink is secondary... This looks good, just not sure
3. did you try getFunctionName?

nchevobbe:
That's great Greg, thank you so much for working on this.
Regarding colors, we should ensure that they work well on a red background (which can happen if we do `console.error(err)`)

I was thinking, could we simply prevent user to expand it ?
It is an inspectable object, but I think users don't care much about it ?

For the custom formatters work, I was thinking we could have an "enhanced" icon. Basically it would be something that would tell you that a special treatment was applied to the object, and let you switch to the "raw" version if you want to inspect the object.

So here we would have the cool output 
```
✨ Error: "A complicated message"
      λ a   script.js:23
      λ b   script.js:3
```

And if you hover the ✨ you'll have a "this is an enhanced version of the logged object. Click to see it raw" (need better wording).

And then, on click, we give the user: 
```
▶︎ ✨ Error: "A complicated message"
```
which only show the error class and message. The user can then expand it to access columnNumber, stack, …

And I think we should force this view when the Error object is in another object (array, object, map, …). Not sure if relying only on MODE would be sufficient.

tromey:
>  So here we could add an onClick on the location element, calling onViewSourceInDebugger.

If you're going this route, how much harder would it be to just fix #452?

nchevobbe:
> If you're going this route, how much harder would it be to just fix #452?

Not so hard, we need to plug in the sourcemap service. But I think this has to be passed from the consumer, and that would require a bit of work to re-render the stack when the sourcemap is ready.

I'd say we can do it in another PR, soon after merging this one (and we would make both part of the same Reps bundle)

jasonLaster:
Yep - I'm looking at the service for a perf bug. Happy to consult, but it's pretty straight forward.

nchevobbe:
For the colors, here's what we are using in the console: 

function name: var(--console-output-color, currentColor)
file name: --frame-link-source
line and column numbers: --frame-link-line-color

they were carefully chosen to not clash with the different background color we can have in both light and dark mode (red, yellow, white/black)

nchevobbe:
re embedded errors: 

image

This looks good to me. Stack is not shown in TINY mode, so it won't appear in previews, and when expanded, it takes more vertical space, but this is fine to me

gregtatum:
I didn't get to this today, I see you approved this, what all from the feedback would you like addressed before being merged?

nchevobbe:
Let's merge it now and address the comments in follow ups

nchevobbe:
@gregtatum I filed a handful of follow up issues. Feel free to tackle some, or not.
I think that we should have at least clickable frames for the next reps bundle (and I'll probably work on this today).

nchevobbe:
See https://github.com/devtools-html/devtools-core/projects/3

PR #948, opened 2018-02-22
Expand node when arrow is clicked

nchevobbe:
The issue was caused because of the documentHasSelection helper
we introduced recently.
When an inline svg document is clicked, the getSelection does return
a Range type.
Here we set an explicit if for the arrow click to fix the issue.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/index.css
+++ b/packages/devtools-reps/src/object-inspector/index.css
@@ -25,3 +25,7 @@
 .object-inspector .object-delimiter {
   color: var(--theme-comment);
 }
+
+.tree.object-inspector .arrow svg {
+  pointer-events: none;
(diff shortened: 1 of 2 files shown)


Loirooriol:
`closest()` is an Element method but you only check that `focusNode` is not a text node. It could still be another kind of node.

Anyways I think `focusNode.closest(".arrow")` is just a hack. Consider something like

    isCollapsed || focusNode === anchorNode && focusNode.nodeType == 1 && !focusNode.hasChildNodes();

Or maybe just add `.arrow svg { pointer-events: none }`

nchevobbe:
now testing the nodeType as Oriol suggested. Couldn't remove the `closest` hack since the arrow element do have child nodes.

PR #955, opened 2018-02-26
Add native mappings in devtools-launchpad babel and webpack configs

nchevobbe:


jasonLaster:
What does this do?

nchevobbe:
this is needed for the launchpad to work in mc world

nchevobbe:
Looks like we don't need the babel thing. Removed.

PR #966, opened 2018-03-01
Update devtools-reps to React 16.2.

nchevobbe:
Update devtools-* packages as well.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/component/classnames.js
+++ b/packages/devtools-reps/src/object-inspector/tests/component/classnames.js
@@ -20,13 +20,13 @@ function generateDefaults(overrides) {
   }, overrides);
 }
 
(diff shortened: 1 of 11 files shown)


packages/devtools-reps/src/object-inspector/tests/component/classnames.js
```diff
@@ -20,31 +20,31 @@ function generateDefaults(overrides) {
   }, overrides);
 }
 
+function mounObjectInspector(props) {
```
MikeRatcliffe: nit: shouldn't this be `mountObjectInspector`?

MikeRatcliffe: nit: Shouldn't this be `mountObjectinspector`?

PR #967, opened 2018-03-01
License improvements

jasonLaster:
### Summary of Changes

These changes are needed for some dependency upgrades in the debugger.html

https://github.com/devtools-html/debugger.html/issues/5560
https://github.com/devtools-html/debugger.html/issues/5465

I included a prettier commit, but the 2nd commit has the relevant changes.

https://creativecommons.org/publicdomain/zero/1.0/
https://en.wikipedia.org/wiki/BSD_licenses


tromey:
FWIW this looks reasonable to me.

PR #975, opened 2018-03-05
Tree's aria-level should be depth + 1

### LINKED ISSUE #963: [Tree] aria-level should be depth + 1 (according to spec values should be >= 1).
See https://bugzilla.mozilla.org/show_bug.cgi?id=1438837

nchevobbe:
(according to spec values should be >= 1).

Fixes #963.


nchevobbe:
Synced what happened in [Bug 1438837](https://bugzilla.mozilla.org/show_bug.cgi?id=1438837)

nchevobbe:
CI fails but it's because lerna use direct dependencies instead of relying on published npm version.

PR #973, opened 2018-03-05
Strip browser-loader part in frames url for the Error rep stacktrace

nchevobbe:
Fixes #971.

Extract stacktrace creation into its own function and strip base-loader part if needed

PR #977, opened 2018-03-05
 Better function name in Error stacktrace. 

### LINKED ISSUE #959: Use `getFunctionName` from Function rep in the Error rep stacktrace
`getFunctionName` is an helper that get rid of cryptic `/<` symbols in the function name.
We should call it in the Error rep stacktrace to have nicer function names.

Since it takes a grip, and not a string, it should be split into 2 functions so we can handle strings.

nchevobbe:
Fixes #959.

Built on top of https://github.com/devtools-html/devtools-core/pull/973, only https://github.com/devtools-html/devtools-core/commit/4c4accc965f9a9f233ce2837b956aaed10600d0f is relevant here.

- Display `<anonymous>` for unnamed functions.
- Call cleanFunctionName on it.

### DIFF
--- a/packages/devtools-reps/src/reps/tests/__snapshots__/error.js.snap
+++ b/packages/devtools-reps/src/reps/tests/__snapshots__/error.js.snap
@@ -15,7 +15,9 @@ exports[`Error - Eval error renders with expected text for an EvalError 1`] = `
     <span
       className="objectBox-stackTrace-fn"
       key="fn0"
-    />
+    >
+      &lt;anonymous&gt;
+    </span>
     <span
       className="objectBox-stackTrace-location"
       key="location0"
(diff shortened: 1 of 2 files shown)


jasonLaster:
Nice

packages/devtools-reps/src/reps/tests/__snapshots__/error.js.snap
```diff
@@ -15,7 +15,9 @@ exports[`Error - Eval error renders with expected text for an EvalError 1`] = `
     <span
       className="objectBox-stackTrace-fn"
       key="fn0"
-    />
+    >
```
gregtatum: I believe you landed the better snapshots, thanks for easier reading here.

packages/devtools-reps/src/reps/tests/error.js
```diff
@@ -311,3 +311,24 @@ describe("Error - DOMException", () => {
...
+
+describe("Error - base-loader.js", () => {
+  const stub = stubs.get("base-loader Error");
+
+  it("renders as expected without mode", () => {
+    const renderedComponent = shallow(ErrorRep.rep({
+      object: stub
+    }));
+
+    expect(renderedComponent).toMatchSnapshot();
```
gregtatum: Snapshots work for me!

PR #978, opened 2018-03-06
Guard isRoot on node value truthiness

nchevobbe:
It was causing an issue in the debugger with scopes "Block" nodes


nchevobbe:
yes, I'll add it

nchevobbe:
added a test that was throwing without the fix.
@codehag does it look good to you ?

codehag:
Looks great!

PR #979, opened 2018-03-06
Add a new BLOCK node type in the ObjectInspector.

### LINKED ISSUE #940: Allow ObjectInspector nodes to have custom classname
In https://github.com/devtools-html/debugger.html/issues/5426, we are discussing having separate styling for "Block" nodes in the Scopes panel of the debugger.
The Scopes panel uses the ObjectInspector, and those "Block" nodes are mixed with the other nodes in the data structure passed to the ObjectInspector.

So here, we would need to have a way to specify, from the node data, a specific classname which would be applied to the `.tree-node` element.

nchevobbe:
Fixes #940.

Visual is still discussed at https://github.com/devtools-html/debugger.html/issues/5426#issuecomment-366620776 and might change.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/test-utils.js
+++ b/packages/devtools-reps/src/object-inspector/tests/test-utils.js
@@ -34,7 +34,12 @@ function formatObjectInspector(wrapper: Object) {
       } else {
         arrowStr = "  ";
       }
-      return `${indentStr}${arrowStr}${getSanitizedNodeText(node)}`;
+
+      let icon = "";
+      if (node.find(".node").first().hasClass("block")) {
+        icon = "☲ ";
+      }
+      return `${indentStr}${arrowStr}${icon}${getSanitizedNodeText(node)}`;
     })
     .join("\n");
 }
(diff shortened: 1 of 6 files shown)


packages/devtools-reps/src/object-inspector/tests/test-utils.js
```diff
@@ -34,7 +34,12 @@ function formatObjectInspector(wrapper: Object) {
       } else {
         arrowStr = "  ";
       }
-      return `${indentStr}${arrowStr}${getSanitizedNodeText(node)}`;
+
+      let icon = "";
+      if (node.find(".node").first().hasClass("block")) {
```
codehag: maybe this can be a ternary? Unless there is a chance that this will have further types...

```javascript
const icon = node.find(".node").first().hasClass("block")
  ? "☲ "
  : "";
```

nchevobbe: ✅

nchevobbe:
Thanks for the review yulia. I'm going to merge this, and we'll revisit the visual side with victoria

PR #981, opened 2018-03-06
Use devtools-components-0.3.0 in devtools-reps.

nchevobbe:
Some snapshots are updated since the tree was modified (aria-level),
and formatObjectInspector is updated to address this.


nchevobbe:
thanks Jason :)

Issue #651, opened 2017-09-07
Get rid of __proto__ magic


Loirooriol:
Reps is messing with `__proto__` properties. See e.g. [bug 1390027](https://bugzilla.mozilla.org/show_bug.cgi?id=1390027) and [bug 1397772](https://bugzilla.mozilla.org/show_bug.cgi?id=1397772).

 - Stop ignoring `__proto__` enumerable own properties in object previews.
 - Stop ignoring `__proto__` own properties in object inspector.
 - Stop adding misleading `__proto__` own properties in object inspector. To reference the [[Prototype]] use a `<prototype>` property or something like that.

nchevobbe:
See similar concerns in https://bugzilla.mozilla.org/show_bug.cgi?id=866334

Loirooriol:
Fixed by #987 and #982. Also see #1000.

PR #982, opened 2018-03-06
Have consistent pseudo-nodes in ObjectInspector.

nchevobbe:
`__proto__` is renamed to `<prototype>` and
`[default properties]` is renamed to `<default properties>` so
all our pseudo nodes share the same syntax, i.e. `<${name}>`.


nchevobbe:
cc @Loirooriol

Loirooriol:
Yes yes yes! This fixes #651, right? `<prototype>` is not an identifier so such a user-defined property is quoted as `"<prototype>"` and shouldn't collide with the internal `<prototype>`.

nchevobbe:
Thanks for the review @codehag

PR #987, opened 2018-03-08
Remove Object.assign calls in favor of the spread operator

### LINKED ISSUE #986: Replace Object.assign with spread properties
`Object.assign({}, obj)` is used all over the place. But this calls setters!

This is harmful in case `obj` contains an own `__proto__` property.

You should use `{...obj}` ([spread properties](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/Object_initializer#Spread_properties)) instead. Compare:

```js
let obj = {["__proto__"]: null};
Reflect.ownKeys(Object.assign({}, obj)); // []
Reflect.ownKeys({...obj});               // ["__proto__"]
```

Specifically in https://github.com/devtools-html/devtools-core/blob/b6aa4ab9787d519c9f468af91f4755760a9265b1/packages/devtools-reps/src/reps/grip.js#L156

it's the cause of [bug 1390027](https://bugzilla.mozilla.org/show_bug.cgi?id=1390027).

nchevobbe:
As found by @Loirooriol , this fix an issue with objects containing a `__proto__` property (`Object.assign` calls setters and strips the __proto__ property).

Took this as an opportunity to remove `Object.assign` calls where it could be done.

Fixes #986.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 7.7.2
+    version: 8.9.4
   environment:
     DOWNLOADS_PATH: "$HOME/downloads"
     YARN_PATH: "$HOME/.yarn"
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -267,9 +267,10 @@ class ObjectInspector extends Component {
       return {
         label: this.renderGrip(
           item,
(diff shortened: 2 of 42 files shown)


nchevobbe:
Mhh, circle does not like the spread operator

nchevobbe:
yay, circle's happy with correct Node version

PR #991, opened 2018-03-09
Add cursor:pointer on string links

nchevobbe:

PR #992, opened 2018-03-09
[WIP] Release: devtools-reps 0.21.0

nchevobbe:
WIP because we follow the release process

PR #993, opened 2018-03-11
Fix object spread babel error

kootoopas:
I tried running #991 and I was getting this error:
```
ERROR in ./src/reps/prop-rep.js
Module build failed: SyntaxError: Unexpected token (67:6)

  65 |   } else {
  66 |     key = Rep({
> 67 |       ...props,
     |       ^

```

Then I added `babel-plugin-transform-object-rest-spread` to `detools-reps` and it worked like a charm.

### DIFF
--- a/packages/devtools-reps/.babelrc
+++ b/packages/devtools-reps/.babelrc
@@ -2,6 +2,7 @@
   "presets": ["react", "flow"],
   "plugins": [
     "transform-es2015-modules-commonjs",
-    "transform-flow-strip-types"
+    "transform-flow-strip-types",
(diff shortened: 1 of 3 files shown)


nchevobbe:
Hello @kootoopas , thanks for trying to fix this.
The issue was resolved in https://github.com/devtools-html/devtools-core/pull/992 .
We don't want to transform the spread operator because it causes an issue (See https://github.com/devtools-html/devtools-core/issues/986#issue-303304522).

Hence I'm going to close this, hope you don't mind  :/

PR #995, opened 2018-03-13
WIP - Update source-map to 0.7.2 (#5598)

calebstdenis:
This PR addresses devtools-html/debugger.html#5598 for updating `source-map` to `0.7.2`. The new version uses WebAssembly to encode/decode mappings.  💥 

This PR is marked as WIP for a couple reasons:

* [One of the source mapping Jest tests](https://github.com/calebstdenis/devtools-core/blob/41d6b01c581415b9fd22fdde260ffc4b26f688fb/packages/devtools-source-map/src/tests/locations.js#L80) gives a surprising failure:
image
    * I don't see how this could be correct. Hoping someone from the source-map team can shed some light on this.

* WebAssembly is [loaded differently](https://github.com/mozilla/source-map/blob/43b2687e6e2f3cf6c22926aa473d5031fb7df89d/lib/read-wasm.js#L1) in Node as opposed to the web. (Jest tests run in a Node-like environment.) The way I handled that is almost certainly not how we want to do it. See my comments below for more details.

### DIFF
--- a/packages/devtools-source-map/jest-setup.js
+++ b/packages/devtools-source-map/jest-setup.js
@@ -1,5 +1,6 @@
 // TODO: Remove this file after upgrading to Jest 22.4.1+
 // More info: https://github.com/facebook/jest/issues/687
+// and https://github.com/devtools-html/devtools-core/pull/995#discussion_r174317667
 console.debug = console.log
 
 
--- a/packages/devtools-source-map/jest-setup.js
+++ b/packages/devtools-source-map/jest-setup.js
@@ -1,0 +1,5 @@
+// TODO: Remove this file after upgrading to Jest 22.4.1+
+// More info: https://github.com/facebook/jest/issues/687
+console.debug = console.log
+
+
--- a/packages/devtools-source-map/src/source-map.js
+++ b/packages/devtools-source-map/src/source-map.js
@@ -11,6 +11,9 @@
 
 const { networkRequest } = require("devtools-utils");
 const { SourceMapConsumer, SourceMapGenerator } = require("source-map");
+SourceMapConsumer.initialize({
+  "lib/mappings.wasm": "https://unpkg.com/source-map@0.7.2/lib/mappings.wasm"
+})
 
 const assert = require("./utils/assert");
 const { fetchSourceMap } = require("./utils/fetchSourceMap");
--- a/packages/devtools-source-map/package.json
+++ b/packages/devtools-source-map/package.json
@@ -21,7 +21,7 @@
     "jest": "^19.0.2",
     "md5": "^2.2.1",
     "regenerator-runtime": "^0.10.3",
-    "source-map": "^0.6.1"
+    "source-map": "^0.7.2"
   },
   "devDependencies": {
     "devtools-launchpad": "^0.0.116"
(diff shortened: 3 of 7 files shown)


packages/devtools-source-map/jest-setup.js
```diff
@@ -0,0 +1,5 @@
+// TODO: Remove this file after upgrading to Jest 22.4.1+
+// More info: https://github.com/facebook/jest/issues/687
+console.debug = console.log
```
calebstdenis: In server environments, source-map calls `console.debug`, which does not exist in the version of Jest that this repo is currently running. Some other problems prevented me from simply upgrading Jest. I didn't think it was the right time to be going down that rabbithole.

packages/devtools-source-map/src/source-map.js
```diff
@@ -11,6 +11,9 @@
 
 const { networkRequest } = require("devtools-utils");
 const { SourceMapConsumer, SourceMapGenerator } = require("source-map");
+SourceMapConsumer.initialize({
+  "lib/mappings.wasm": "https://unpkg.com/source-map@0.7.2/lib/mappings.wasm"
```
calebstdenis: It does work, I must say! The relevant documentation is [here](https://github.com/mozilla/source-map#sourcemapconsumerinitializeoptions). The relevant code is [here](https://github.com/mozilla/source-map/blob/43b2687e6e2f3cf6c22926aa473d5031fb7df89d/lib/read-wasm.js#L1).

I'm not sure what is the ideal way to do the wasm init context of a Webpack app. Would it work just to use the Node way of loading wasm? @fitzgen would love to get your input on this :)

jasonLaster:
CC @tromey Would you mind taking a look at this?

packages/devtools-source-map/src/source-map.js
tromey: I don't know the answer here either, but this is going to need to be configurable so that this can also work in Mozilla Central.  There, we'll want to ship the wasm in Firefox and we'll need to use a resource URL or something.

jasonLaster: yep, we'll need to provide something when we start the worker...

tromey:
I looked at the test failure a bit.  I am not sure this is a valid test.

The issue is that the source map contains two mappings for the `O[5,4]` location mentioned in the test case.  One does map to `G[8,0]` -- but another maps to `G[8,114]`.

I tend to think that the source map implementation has some freedom to decide which result it returns in this case.  (Or perhaps it should return both.)

I'm not 100% sure though, so CC @fitzgen on this as well.

tromey:
I forgot to mention - I read this patch and, aside from the URL configurability thing, it is pretty much what I thought this patch should look like.  So assuming the webpack and test case issues are worked out, r+ from me.  Thank you for doing this!

jasonLaster:
Thanks @tromey 

Lets sort out the symmetry thing. I'm good with anything as long as I understand what is going on.

I'm happy to fix this up so we can get the mappings.wasm in m-c and pulled in in the launchpad.

calebstdenis:
>I tend to think that the source map implementation has some freedom to decide which result it returns in this case.
If this is the case and debugger.html doesn't make any assumptions to the contrary, it should be OK to work around the failing tests like so: 1ed76bbd278ddb0ccc6b98ec3aeeed404e291a7a

loganfsmyth:
I've posted an updated version of this in https://github.com/devtools-html/debugger.html/pull/7071, so going to close this.

PR #998, opened 2018-03-16
Don't inline SVG images in devtools-reps

### LINKED ISSUE #988: Switch to CSS SVGs
Referencing https://github.com/devtools-html/debugger.html/issues/4350

We would like to switch from using react-based inline SVGs to CSS-based SVGs. The primary advantage is performance. SVGs created by React are slow to render. CSS drawn SVGs have a significant perf benefit.

TODO:
- [ ] convert Object Inspector arrows
- [ ] convert "Open in inspector" icon
- [ ] convert "Jump to definition" icon

nchevobbe:
This causes warning in the browser console AND can cause performance issue.
Using those images as background images means we can remove InlineSVG dependency
and helpers.
It also means we need to have a postCSS config to be able to transform those
images URL for development in launchpad, and also for the copy-assets command.

Some tests needed some adjustment because of the new markup.

Fixes #988


packages/devtools-reps/postcss.config.js
```diff
@@ -0,0 +1,35 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const mapUrl = require("postcss-url-mapper");
+const MC_PATH = "chrome://devtools/skin/images/devtools-reps/";
+const EXPRESS_PATH = "/devtools-reps/images/";
+
+function mapUrlProduction(url, type) {
+  const newUrl = url
+    .replace("/images/open-inspector.svg", MC_PATH + "open-inspector.svg")
+    .replace("/images/jump-definition.svg", MC_PATH + "jump-definition.svg");
```
jasonLaster: can this be more general

nchevobbe: Yes, it can, but we can think about it later.
Also, with the plan to migrate all this under the debugger repo, I guess there will be changes anyway, so I'd rather not spend too much time on this at the moment, since there is a pretty small amount of images.

PR #1006, opened 2018-03-21
Use onViewSourceInDebugger prop when clicking on the Error rep stacktrace location

nchevobbe:
Fixes #942.


jasonLaster:
Thanks!

PR #1008, opened 2018-03-23
Do not create an ObjectInspector element for primitive grips.

nchevobbe:
When we have a single primitive root, there's no advantage rendering
them inside the ObjectInspector component. On the contrary, it adds
some overhead that can harm performances.
This patch allow us to branch early so we don't go through the ObjectInspector
element if we don't need to.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/utils/should-render-roots-in-reps.js
+++ b/packages/devtools-reps/src/object-inspector/tests/utils/should-render-roots-in-reps.js
@@ -1,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const Utils = require("../../utils");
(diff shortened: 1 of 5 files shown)


nchevobbe:
This shows some nice [improvements in DAMP](https://treeherder.mozilla.org/perf.html#/compare?originalProject=try&originalRevision=1b4a868711d05b7532129d26be87dfba5fcdb1ba&newProject=try&newRevision=7b4feacf2c7333c7ea17b6db7a8854f00add0596&framework=1)

packages/devtools-reps/src/object-inspector/tests/utils/should-render-roots-in-reps.js
```diff
@@ -0,0 +1,74 @@
...
+    }])).toBeFalsy();
+  });
+
+  it("returns false for Arrays", () => {
+    expect(shouldRenderRootsInReps([{
+      name: "label",
+      contents: {value: gripArrayStubs.get("testMaxProps")}
+    }])).toBeFalsy();
+  });
+});
```
janodvarko: I like the tests!

PR #1010, opened 2018-03-26
Fix Error Rep with longString stacktrace.

### LINKED ISSUE #1009: Nested error representation is weird
Renders as invalid object if not expanded:
screen shot 2018-03-22 at 17 18 32

When expanded it renders normally I guess:
screen shot 2018-03-22 at 17 08 05

The error has to do with a wrong import name if you're curious. :)

nchevobbe:
This was throwing because we are calling split on preview.stack,
as we thought having a string. But if this is a longString, we get
an object and thus we have to pick longString's initial value.
This can be considered as a hotfix since we are not properly rendering
the whole stacktrace in this case. This will be handled in a separate
issue, maybe directly on the server side.

Fixes #1009


packages/devtools-reps/src/reps/error.js
```diff
@@ -82,7 +82,16 @@ function ErrorRep(props) {
  */
 function getStacktraceElements(props, preview) {
   const stack = [];
-  preview.stack
+  if (!preview.stack) {
+    return stack;
+  }
+
+  const isStacktraceALongString = preview.stack.type === "longString";
```
janodvarko: Could we somehow utilize `isLongString()` from reps.js?

packages/devtools-reps/src/reps/error.js
```diff
@@ -147,6 +156,12 @@ function getStacktraceElements(props, preview) {
       }, location));
     });
 
+  if (isStacktraceALongString) {
+    // Remove the last frame which is certainly incomplete.
+    // Can be removed when https://bugzilla.mozilla.org/show_bug.cgi?id=1448833 is fixed.
+    stack.splice(-2);
```
janodvarko: The comment says "Remove the last frame", but using -2 means last two frames are removed, is that correct?

packages/devtools-reps/src/reps/tests/error.js
```diff
@@ -334,6 +334,18 @@ describe("Error - base-loader.js", () => {
...
 
+describe("Error - longString stacktrace", () => {
+  const stub = stubs.get("longString stack Error");
+
+  it("renders as expected", () => {
+    const renderedComponent = shallow(ErrorRep.rep({
+      object: stub
+    }));
+
+    expect(renderedComponent).toMatchSnapshot();
```
janodvarko: The test should also cover case where the stack is complete, ie it's string not long string

packages/devtools-reps/src/reps/error.js
nchevobbe: sure

packages/devtools-reps/src/reps/error.js
nchevobbe: yes, it's unclear here, but stack has 2 items for each frame (function name + location). I'll change that comment to make it more obvious.

packages/devtools-reps/src/reps/tests/error.js
janodvarko: Ok, excellent.

Thanks!

Honza

PR #1011, opened 2018-03-26
Don't use ObjectInspector for Error grips.

### LINKED ISSUE #960: Do not use the ObjectInspector for Error rep
When a user log an Error object, I'm pretty sure that what they want 99% of the time is to have the error message and a meaningful and actionable stacktrace.

Thus, I think we should not render Error wrapped in an ObjectInspector unless the user deliberately asks for it.

For the custom formatters work, I was thinking we could have an "enhanced" icon. Basically it would be something that would tell you that a special treatment was applied to the object, and let you switch to the "raw" version if you want to inspect the object.

So here we would have the nice stacktrace 
```
✨ Error: "A complicated message"
      λ a   script.js:23
      λ b   script.js:3
```

And if you hover the ✨ you'll have a "this is an enhanced version of the logged object. Click to see it raw" (need better wording).

And then, on click, we give the user: 
```
▶︎ ✨ Error: "A complicated message"
```
which only show the error class and message. The user can then expand it to access columnNumber, stack, …

```
▼ ✨ Error: "A complicated message"
 |      columNumber: 23
 |      fileName: "script.js"
 |      lineNumber: 23
 |      stack: "…"
```

And I think we should force this view when the Error object is in another object (array, object, map, …).

Here the downside is that the ✨ is right next to  the expand arrow, and I guess it could cause mis-clicks.

@violasong what do you think of this ?

nchevobbe:
Fixes #960.
Since the ErrorRep already provides all the error informations
in a nice way (message and stacktrace), there's no benefits
being able to inspect the object. Furthermore, when in the console,
throwing an Error object gives you a double arrow (first to expand
the stack we get from the exception, second is the objectInspector
arrow), which is clearly not ideal.
Using Rep directly solves these issues.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/utils/should-render-roots-in-reps.js
+++ b/packages/devtools-reps/src/object-inspector/tests/utils/should-render-roots-in-reps.js
@@ -11,6 +11,7 @@ const undefinedStubs = require("../../../reps/stubs/undefined");
 const gripStubs = require("../../../reps/stubs/grip");
 const gripArrayStubs = require("../../../reps/stubs/grip-array");
 const symbolStubs = require("../../../reps/stubs/symbol");
+const errorStubs = require("../../../reps/stubs/error");
 
 describe("shouldRenderRootsInReps", () => {
   it("returns true for a string", () => {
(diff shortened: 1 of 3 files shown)


nchevobbe:
Last point in favor of this move: this will bring us at parity with Chrome, which also do not allow user to inspect an error object.

Issue #1003, opened 2018-03-19
Slashes are not escaped in paths


Loirooriol:
1. Run `inspect({a: {b: {}}, "a/b": []})`
2. Expand `a`
3. Expand `b` (inside `a`)

Result: `a/b` is also expanded.

Either escape slashes in paths, or consider using symbols instead of strings to avoid collisions.

nchevobbe:
Node paths were refactored to use Symbols in https://github.com/devtools-html/devtools-core/pull/1013 , so we don't have path clashes anymore

Issue #1002, opened 2018-03-19
SAFE_PATH_PREFIX is not safe at all


Loirooriol:
1. Run `inspect(Object.assign(Promise.resolve({}), {"##-value": []}))`
2. Expand `<value>`

Result: both `<value>` and `"##-value"` are expanded.

This happens for all pseudo-nodes except `<prototype>`, which has its own problem (#1001).

Either prefix normal properties with something different than `##-`, or use symbols instead of strings for paths. The latter probably requires more work but seems the proper approach IMO.

jasonLaster:
Yes, we discussed symbols early on and decided that the `##-` prefix was worth the risk for the time being. We should address this hack at some point.

nchevobbe:
I agree of the issue, even if it's unlikely to happen.
We can't use just symbol for path, since each path is built upon the previous one.
But we could make the path an array which would be something like
```js
[
  "/root"
  "a",
  Symbol(NODE_TYPE.PROTOTYPE)
]
```
and use Sets and Maps in the state when we want to refer to them.
I think it shouldn't be too hard to do

nchevobbe:
`SAFE_PATH_PREFIX` does not exist anymore (removed in https://github.com/devtools-html/devtools-core/pull/1013)

PR #1013, opened 2018-03-27
 Use symbols for nodes path

### LINKED ISSUE #1001: Conflict between <prototype> and "<prototype>"
1. Run `inspect({"<prototype>": null})`
2. Expand `<prototype>`
3. Collapse `<prototype>`
4. Expand `<prototype>`

Result: the properties of the `<prototype>` also appear below `"<prototype>"`.

screencast

nchevobbe:
Fixes #1001, #1002 & #1003.

The path property on a node is very important as it serve
as an identifier to know if a node is expanded, and what
its properties are. Currently, this path was a simple string
which could led to 2 nodes having the same path, which could
have unwanted effect on user interaction.
To avoid this, the path is now a symbol, which has the advantage
of being unique.
Because of this change, we needed to adapt our code so we always
have the same path reference in the state.
With this patch, we'll always dispatch the nodePropertiesLoaded action,
even if the node don't have any properties. We'll store null in the state,
so we can then check if a node was properly handled.
As a result, we change the shouldComponentUpdate of the component
to reflect this new behavior.
In order to have always the same reference, we always add created node
to the cacheNodes property, which we can do because of the change in
loadedProperties.
This should also give us some performance improvements as we don't have
to recreate some nodes twice.
We take this as an opportunity to change the signature of the createNode
function to be more explicit (it now expect an object of options), and
we also create the path of the node from its parent when the path option
is not supplied.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/actions.js
+++ b/packages/devtools-reps/src/object-inspector/actions.js
@@ -39,7 +39,10 @@ function nodeExpand(
       type: "NODE_EXPAND",
       data: {node}
     });
-    dispatch(nodeLoadProperties(node, actor, loadedProperties, createObjectClient));
+
+    if (!loadedProperties.has(node.path)) {
+      dispatch(nodeLoadProperties(node, actor, loadedProperties, createObjectClient));
+    }
   };
 }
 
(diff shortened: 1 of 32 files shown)


nchevobbe:
Apologies for the lengthy review @janodvarko , but it was hard splitting it up since we are changing a core piece of the component.

nchevobbe:
Thanks a lot for the review Honza ! Fixed the typos

Issue #941, opened 2018-02-20
We should be able to expand longStrings in the ObjectInspector


nchevobbe:
**Steps to reproduce**
1. Evaluate `({long: "ab ".repeat(10000)})`

**Expected results**
I can expand the `long` property to see the whole string content

**Actual results**

I don't have any way to see the whole string content

---

@kootoopas I am assigning this to you since you started to work in it in another PR, but feel free to not take it if you don't want to/don't have the time :)

kootoopas:
Oh hey @nchevobbe. Cool, I'm currently working on this anyway. :)

How do you think it's the best way to expand it?
* The clickable existing trailing ellipsis option seems that it can go unnoticed to me.
* If we were to make the whole string rep clickable, wouldn't that be an overkill?
* I like the idea of an expand button at the end of the string... Maybe label it with `…` and remove the existing in-string trailing ellipsis?

nchevobbe:
So, I think we should align with what we are doing in the json viewer: if you go to [this URL](data:application/json,{"string":"aaa","longString":"aaaaaasdfsdfsdolijfsdklfjsdlkfjsdlfkjsdflksdjflskdjfsdlkjfsdlkjfhsdkjfhsdkjfhdfkgjhdfkgjhdfkgjhdfgkjdfhgkdfjhgkdf"}), you'll see that you can collapse the "longString" property, and re-expand it.
I think it should behave the same in the ObjectInspector so we are consistent.

Anything that would be placed at the end of the string can be unnoticed I think, and I like the idea of a consistent way to expand a node. What do you think ?

kootoopas:
I prefer the consistent approach too. I didn't think about checking out the json viewer at all!

So, If I understand correctly:
* prepend a togglable arrow
* abbreviate the string as in the json viewer
* style the abbreviated string, if it's a standalone string, or property key, if it's an object property, like a link (pointer & underline)

nchevobbe:
Fixed by https://github.com/devtools-html/devtools-core/pull/1015

PR #1015, opened 2018-03-28
[ObjectInspector] Add fullText expansion for longString grips

kootoopas:
Associated Issue: #941

@nchevobbe This is a work-in-progress but I wanted to share it anyway to let you know that
it's on the way. :)

### DIFF
--- a/packages/devtools-reps/src/launchpad/index.js
+++ b/packages/devtools-reps/src/launchpad/index.js
@@ -4,6 +4,7 @@
 
 // globals window, document
 
+
 require("../reps/reps.css");
 
 const React = require("react");
@@ -11,7 +12,7 @@ const ReactDOM = require("react-dom");
 
 const { bootstrap, renderRoot } = require("devtools-launchpad");
 
-const { nodeIsLongString } = require("../object-inspector/utils/node");
+const { isLongString } = require("../reps/string");
 const RepsConsole = require("./components/Console");
 const { configureStore } = require("./store");
 
@@ -35,8 +36,8 @@ function onConnect(connection) {
     },
 
     getObjectClient: function (grip) {
-      if (nodeIsLongString(grip)) {
-        return connection.tabConnection.debuggerClient.getString(grip);
+      if (isLongString(grip)) {
+        return connection.tabConnection.tabTarget.activeConsole.getString(grip);
       }
 
       return connection.tabConnection.threadClient.pauseGrip(grip);
--- a/packages/devtools-reps/src/launchpad/index.js
+++ b/packages/devtools-reps/src/launchpad/index.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // globals window, document
+
 require("../reps/reps.css");
 
 const React = require("react");
@@ -10,6 +11,7 @@ const ReactDOM = require("react-dom");
 
 const { bootstrap, renderRoot } = require("devtools-launchpad");
 
+const { nodeIsLongString } = require("../object-inspector/utils/node");
 const RepsConsole = require("./components/Console");
 const { configureStore } = require("./store");
 
@@ -19,6 +21,8 @@ function onConnect(connection) {
   if (!connection) {
     return;
   }
+
+  console.log(connection.tabConnection.debuggerClient);
 
   const client = {
     clientCommands: {
@@ -31,6 +35,10 @@ function onConnect(connection) {
     },
 
     getObjectClient: function (grip) {
+      if (nodeIsLongString(grip)) {
+        return connection.tabConnection.debuggerClient.getString(grip);
+      }
+
       return connection.tabConnection.threadClient.pauseGrip(grip);
     },
 
--- a/packages/devtools-reps/src/launchpad/index.js
+++ b/packages/devtools-reps/src/launchpad/index.js
@@ -4,7 +4,6 @@
 
 // globals window, document
 
-
 require("../reps/reps.css");
 
 const React = require("react");
--- a/packages/devtools-reps/src/launchpad/index.js
+++ b/packages/devtools-reps/src/launchpad/index.js
@@ -22,8 +22,6 @@ function onConnect(connection) {
     return;
   }
 
-  console.log(connection.tabConnection.debuggerClient);
-
   const client = {
     clientCommands: {
       evaluate: input => new Promise(resolve => {
--- a/packages/devtools-reps/src/launchpad/index.js
+++ b/packages/devtools-reps/src/launchpad/index.js
@@ -11,7 +11,6 @@ const ReactDOM = require("react-dom");
 
 const { bootstrap, renderRoot } = require("devtools-launchpad");
 
-const { isLongString } = require("../reps/string");
 const RepsConsole = require("./components/Console");
 const { configureStore } = require("./store");
 
@@ -33,12 +32,11 @@ function onConnect(connection) {
     },
 
     getObjectClient: function (grip) {
-      if (isLongString(grip)) {
-        // Use `WebConsoleClient.getString`.
-        return connection.tabConnection.tabTarget.activeConsole;
-      }
-
       return connection.tabConnection.threadClient.pauseGrip(grip);
+    },
+
+    getLongStringClient: function (grip) {
+      return connection.tabConnection.tabTarget.activeConsole.longString(grip);
     },
 
     releaseActor: function (actor) {
@@ -47,7 +45,7 @@ function onConnect(connection) {
   };
 
   let store = configureStore({
-    makeThunkArgs: (args, state) => Object.assign({}, args, { client }),
+    makeThunkArgs: (args, state) => ({ ...args, client }),
     client,
   });
   renderRoot(React, ReactDOM, RepsConsole, store);
--- a/packages/devtools-reps/src/launchpad/index.js
+++ b/packages/devtools-reps/src/launchpad/index.js
@@ -36,7 +36,8 @@ function onConnect(connection) {
 
     getObjectClient: function (grip) {
       if (isLongString(grip)) {
-        return connection.tabConnection.tabTarget.activeConsole.getString(grip);
+        // Use `WebConsoleClient.getString`.
+        return connection.tabConnection.tabTarget.activeConsole;
       }
 
       return connection.tabConnection.threadClient.pauseGrip(grip);
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -175,6 +175,7 @@ class ObjectInspector extends Component {
 
     const {
       createObjectClient,
+      createLongStringClient,
       loadedProperties,
       nodeExpand,
       nodeCollapse,
@@ -189,7 +190,8 @@ class ObjectInspector extends Component {
         return rootValue && rootValue.actor === value.actor;
       });
       const actor = isRoot || !value ? null : value.actor;
-      nodeExpand(item, actor, loadedProperties, createObjectClient);
+      nodeExpand(item, actor, loadedProperties, createObjectClient,
+        createLongStringClient);
     } else {
       nodeCollapse(item);
     }
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -299,7 +299,7 @@ class ObjectInspector extends Component {
         repProps.mode = MODE.TINY;
       }
 
-      repProps.showFullText = nodeIsLongString(item) && expanded;
+      repProps.showFullText = nodeHasFullText(item) && expanded;
 
       return {
         label,
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -285,11 +285,8 @@ class ObjectInspector extends Component {
     if (nodeIsLongString(item)) {
       const repProps = {
         ...this.props,
-        member: { open: expanded }
+        showFullText: true,
       };
-      item.contents.value.fullText = expanded && nodeHasFullText(item)
-        ? this.getItemChildren(item).fullText
-        : null;
 
       return {
         value: Utils.renderRep(item, repProps)
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -210,7 +210,7 @@ class ObjectInspector extends Component {
     value?: string | Element,
     label?: string
   }) {
-    let label = item.name;
+    const label = item.name;
     const isPrimitive = nodeIsPrimitive(item);
 
     if (nodeIsOptimizedOut(item)) {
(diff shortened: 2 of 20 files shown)


packages/devtools-reps/src/object-inspector/utils/client.js
```diff
@@ -80,6 +80,19 @@ async function getPrototype(
   return objectClient.getPrototype();
 }
 
+async function getFullText(
+  objectClient: ObjectClient,
+  stringGrip: Node,
+) : ?Promise<{prototype?: Object}> {
```
nchevobbe: I think this should be something like `: ?Promise<{_fullText?: string}>`

packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -155,7 +156,7 @@ class ObjectInspector extends Component {
   }
 
   setExpanded(item: Node, expand: boolean) {
-    if (nodeIsPrimitive(item)) {
+    if (nodeIsPrimitive(item) && !nodeIsLongString(item)) {
```
nchevobbe: I wonder if we should return false in nodeIsPrimitive for long strings 🤔 
It would make sense since now we don't render top-level primitives using the launchpad.
But I don't know if there would be unwanted side effects doing that

packages/devtools-reps/src/launchpad/index.js
```diff
@@ -31,6 +35,10 @@ function onConnect(connection) {
     },
 
     getObjectClient: function (grip) {
+      if (nodeIsLongString(grip)) {
```
nchevobbe: i'm not sure we should use nodeIsLongString here. We are dealing with a grip, and nodeIsLongString expect a ObjectInspector node, so I'm unsure this is correct.
We could use https://github.com/devtools-html/devtools-core/blob/b6819f649da535d42005ee06403bd44efe38633a/packages/devtools-reps/src/reps/string.js#L270-L272 which seems like a safe bet here.

packages/devtools-reps/src/object-inspector/utils/client.js
```diff
@@ -80,6 +80,19 @@ async function getPrototype(
...
+  if (typeof objectClient.getString !== "function") {
```
nchevobbe: It doesn't seem like we have the `getString` function on the [ObjectClient](https://searchfox.org/mozilla-central/source/devtools/shared/client/object-client.js)
But we do have it on the [WebConsoleClient](https://searchfox.org/mozilla-central/rev/028cd8342899357d80fafba2d528a0fd5819e316/devtools/shared/webconsole/client.js#709-745).
Another way of getting a longString content is to use the [LongStringClient](https://searchfox.org/mozilla-central/source/devtools/shared/client/long-string-client.js)

So here, either we can pass to the ObjectInspector a `createLongStringClient` function, like we do for `createObjectClient`, which would create the LongStringClient and call substring on it, or a more straightforward  `getLongStringFullText` function (which could use `WebConsoleClient.getString` in caller code).

What do you think would be the best ?

nchevobbe:
There seems to have some failures in tests: 
```
Error: Uncaught [TypeError: isLongString is not a function]
          at nodeIsLongString (/home/ubuntu/devtools-core/packages/devtools-reps/src/object-inspector/utils/node.js:190:10)
          at ObjectInspector.setExpanded (/home/ubuntu/devtools-core/packages/devtools-reps/src/object-inspector/component.js:147:35)
          at Object.onExpand (/home/ubuntu/devtools-core/packages/devtools-reps/src/object-inspector/component.js:367:30)
```

maybe we don't require `isLongString` as we should ?

packages/devtools-reps/src/object-inspector/utils/client.js
kootoopas: I think I'll go with `WebConsoleClient.getString` since it seems to be kind of optimized and avoids redundant queries by immediately returning the cached `_fullText`. It is also more straightforward to use since it gets it through 1 public method call.

packages/devtools-reps/src/object-inspector/utils/client.js
kootoopas: 👍

packages/devtools-reps/src/object-inspector/component.js
kootoopas: Omitting longstrings from primitives would influence:
- `shouldRenderRootsInReps`, making longStrings not rendered as reps directly, which is what we want since we want them within ObjectInspector:
  ```javascript
  if (shouldRenderRootsInReps(roots)) {
    return renderRep(roots[0], props);
  }
  return new OI(props);
  ```
- `shouldLoadItemPrototype`, loading prototypes for them
- `shouldLoadItemSymbols`, loading symbols for them

kootoopas: If I understand correctly, we don't want any symbols and prototype for it but we want it in the OI.

In the OI it shouldn't have any children and in `renderTreeItem` we probably need to add a `renderTreeItemValue` that renders if roots is longString, since we don't need labels for them.

Lastly `setExpanded` should should toggle value.member.open if it is root and longString.

nchevobbe: Thanks for digging into that @kootoopas ! So, looks like we should keep longString a part of nodeIsPrimitive and add if statements in the places that need so.

nchevobbe:
> There seems to have some failures in tests:

Seems fixed now :)

nchevobbe:
@kootoopas is this ready for review or is there more things to do ?

packages/devtools-reps/src/object-inspector/component.js
kootoopas: Hmm, 🤔  maybe we shouldn't. We can add checks to skip prototype, symbols or whatever unwanted loadable property for a longString while treating it as a non primitive.

kootoopas:
@nchevobbe It's incomplete! I'll inform you once it's ready.

kootoopas:
@nchevobbe Hey there! I've finally created the togglable longString, although the following change breaks a lot of tests (see circleci), probably because I'm waiting for properties to load before expanding the node:
```diff
diff --git a/packages/devtools-reps/src/object-inspector/actions.js b/packages/devtools-reps/src/object-inspector/actions.js
index 8728601..f6d0280 100644
--- a/packages/devtools-reps/src/object-inspector/actions.js
+++ b/packages/devtools-reps/src/object-inspector/actions.js
@@ -35,11 +35,11 @@ function nodeExpand(
   createObjectClient : (RdpGrip) => ObjectClient
 ) {
   return async ({dispatch} : ThunkArg) => {
+    await dispatch(nodeLoadProperties(node, actor, loadedProperties, createObjectClient));
     dispatch({
       type: "NODE_EXPAND",
       data: {node}
     });
-    dispatch(nodeLoadProperties(node, actor, loadedProperties, createObjectClient));
   };
 }

```
My issue is that I can't run jest and progress. Seems like `jest@22.4.2` causes the issue. I'm getting this message for every suite on `yarn test`:
```
  ● Test suite failed to run

    TypeError: environment.setup is not a function
      
      at ../node_modules/jest-runner/build/run_test.js:71:23

```

Seems that it's an unresolved jest issue. Other people have already stumbled upon this:
https://github.com/facebook/jest/issues/5119
https://github.com/facebook/jest/issues/5294

Are you guys using a workaround for this?

nchevobbe:
Thanks a lot @kootoopas 

> I've finally created the togglable longString, although the following change breaks a lot of tests (see circleci), probably because I'm waiting for properties to load before expanding the node

Yes, you are probably right, that would impact a lot of tests. By the way, this is likely to change in https://github.com/devtools-html/devtools-core/pull/1013, so I'd recommend you not to make this change for now.

> My issue is that I can't run jest and progress. Seems like jest@22.4.2 causes the issue.

Did you try to run `yarn nom` from the root folder and then try to run the tests again ? I encountered errors as well and this command seems to do the trick to me.

kootoopas:
>Yes, you are probably right, that would impact a lot of tests. By the way, this is likely to change in #1013, so I'd recommend you not to make this change for now.

👍 

Also: yep, `yarn nom` did it for me too.

nchevobbe:
https://github.com/devtools-html/devtools-core/pull/1013 landed, you might want to rebase your patch on top of master. Hope it won't be too hard !

kootoopas:
Aaaaand it's done.

Ready to be reviewed whenever, @nchevobbe :) . I hope we met 61's deadline...

packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -284,21 +286,26 @@ class ObjectInspector extends Component {
...
       }
 
+      repProps.member = {
+        open: nodeHasFullText(item) && expanded
+      };
+
       return {
         label,
-        value: Utils.renderRep(item, repsProp)
+        value: Utils.renderRep(item, repProps)
```
kootoopas: Should we also pass a small `cropLimit` so that the `initial` string can be displayed in a single line, to mirror JSONViewer's way of displaying it?

nchevobbe: Not sure about that.

packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -284,21 +286,26 @@ class ObjectInspector extends Component {
...
+        repProps.mode = this.props.mode === MODE.LONG
           ? MODE.SHORT
           : MODE.TINY;
       }
       if (expanded) {
-        repsProp.mode = MODE.TINY;
+        repProps.mode = MODE.TINY;
       }
 
+      repProps.member = {
```
nchevobbe: maybe we should wrap this in a `nodeIsLongString` if ?

packages/devtools-reps/src/object-inspector/utils/load-properties.js
```diff
@@ -88,6 +94,10 @@ function mergeResponses(responses: Array<Object>) : Object {
     if (response.prototype) {
       data.prototype = response.prototype;
     }
+
+    if (typeof response === "string") {
```
nchevobbe: This looks brittle to me, I guess we could have in the future more than 1 response being a string.
In l.76, could we put into the `promises` array something more predictable ? Something like
```js
if (shouldLoadFullText(item, loadedProperties)) {
  promises.push(
    getFullText(getObjectClient(), getValue(item))
      .then(fullString => ({fullString}))
  );
}
```

This way we can test the `fullString` property existence here.

packages/devtools-reps/src/object-inspector/utils/node.js
```diff
@@ -233,10 +249,12 @@ function nodeHasEntries(
     return false;
   }
 
-  return value.class === "Map"
-    || value.class === "Set"
-    || value.class === "WeakMap"
-    || value.class === "WeakSet";
+  return [
```
nchevobbe: could we revert this change ? This doesn't seem related to longString

packages/devtools-reps/src/object-inspector/utils/load-properties.js
kootoopas: I didn't like the generic string check either but I haven't thought of postprocessing the promise to be honest!

packages/devtools-reps/src/object-inspector/component.js
kootoopas: 👍

packages/devtools-reps/src/object-inspector/component.js
kootoopas: 👍

packages/devtools-reps/src/object-inspector/utils/node.js
kootoopas: Oh okay. I just thought it would read easier I guess. `¯\_(ツ)_/¯`

packages/devtools-reps/src/object-inspector/utils/client.js
```diff
@@ -80,6 +81,17 @@ async function getPrototype(
   return objectClient.getPrototype();
 }
 
+async function getFullText(
+  objectClient: ObjectClient,
+  object: NodeContents,
+) : Promise<?string | {}> {
+  if (typeof objectClient.getString !== "function") {
```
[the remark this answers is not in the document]
kootoopas: I guess it's a hack to treat `WebConsoleClient` as an `ObjectClient` and only care about the `getString` method then:

```javascript
    // packages/devtools-reps/src/launchpad/index.js
    getObjectClient: function (grip) {
      if (isLongString(grip)) {
        // Use `WebConsoleClient.getString`.
        return connection.tabConnection.tabTarget.activeConsole;
      }
```

I thought we could avoid passing a longString specific function and increase the OI's and the loadProperties functions' arity, since all property loaders are owned by `ObjectClient` (`enumEntries`, `enumProperties`, `getPrototype`, ...), but if that's confusing and the current `WebConsoleClient.getString` lacks something we can go with your approach.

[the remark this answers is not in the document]
nchevobbe: I think it's a but confusing. The webConsoleClient, as its name indicates, is specific to the webconsole. Here the objectInspector can be used in a variety of tools (debugger, inspector, …), and the way they can get a longString client might differ from the console.
If we go this way and let the responsability of the consumer to branch on longString to provide a specific "ObjectClient", I'd rather call it something else so we can't confuse it with the actual ObjectClient class (e.g. ObjectPropertyFetcher, or something with a better name :D). 

I'm still unsure which way we should go for now. I'd say that letting the ObjectInspector chose which client to create and how to fetch data makes the code easier to reason about and use from a consumer standpoint, whereas if we need to branch on the consumer on specific node types, the ObjectInspector delays some of its responsabilities. 

I'm not trying to impose something here and I'm genuinely hesitating between the two approoaches. What do you think @kootoopas ?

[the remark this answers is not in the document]
kootoopas: I hope I got the functionality and your words right. I will attempt to give some more arguments and a conclusion.

The `ObjectPropertyFetcher` seems more reasonable if we're gonna go with the single fetcher  function (now `getObjectClient`) approach, but to be honest I'd rather delegate picking the right client for the grip to the `OI` since it will make `OI` initialization more straightforward in dependents. Else, if I understand correctly, we're forcing dependents to implement a complicated `getObjectClient` (or `getObjectPropertyFetcher`):
https://github.com/kootoopas/devtools-core/blob/03ebe023945283302de8aa984fc700747f0c7ca2/packages/devtools-reps/src/launchpad/index.js#L35-L42

So basically, I'm voting for the addition of a `createLongStringClient` prop in `ObjectClient` that will result to easier consumer usage! (given that my args are valid 🤞)

nchevobbe:
Yes, you understood me well. Let's give `createLongStringClient` a shot then

Le lun. 9 avr. 2018 à 18:57, Fotis Koutoupas <notifications@github.com> a
écrit :

kootoopas:
This works but lacks tests. They're on the way!

packages/devtools-reps/src/object-inspector/utils/client.js
```diff
@@ -80,6 +81,28 @@ async function getPrototype(
   return objectClient.getPrototype();
 }
 
+async function getFullText(
+  longStringClient: LongStringClient,
+  object: NodeContents,
+) : Promise<?Object> {
```
kootoopas: Whoops, need to change this to `Promise<{fullText?: string}>`

nchevobbe:
@kootoopas Do you mind if we merge this PR and add tests in a follow up ?

kootoopas:
@nchevobbe Yesss... 🎉 I'm very happy to hear that it was easy is to integrate! It surely was a nice challenge!

Sure thing let's merge it.

PR #1021, opened 2018-04-03
Fix expanding node with selected text.

### LINKED ISSUE #1017: Can't interact with reps' "expand" icon if some text is selected
__From [Bug 1449865](https://bugzilla.mozilla.org/show_bug.cgi?id=1449865)__

STR:
1. Run this code in the console:

```js
var map = new Map(); 
for (var i = 0; i < 20; i++) { 
  map.set('foo' + i, 'bar' + i); 
} 
map;
```

2. select a word in the console, eg the "map" in the code above
3. try to expand the reps using the "arrow" icon

=> It's not possible.
We first need to unselect by simply clicking elsewhere.
Note that expanding clicking on "Map" seems to work.

nchevobbe:
Fixes #1017.

Since clicking on the arrow does NOT unselect potential selected
text, we need to check the target of the click and see if it is
the node arrow. In such case, we don't block the expansion.

---

A mochitest will be added in mozilla-central to be sure it does work with "true" text selection (in Node we can only rely on a mock)

### DIFF
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -372,8 +372,16 @@ class ObjectInspector extends Component {
       onClick: e => {
         e.stopPropagation();
 
-        // If the user selected text, bail out.
-        if (Utils.selection.documentHasSelection()) {
+        // If this click happened because the user selected some text, bail out.
+        // Note that if the user selected some text before and then clicks here,
+        // the previously selected text will be first unselected, unless the user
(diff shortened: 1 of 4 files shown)


nchevobbe:
is there a bug on Github ? I can't see your comments on this PR @julienw  🤔

julienw:
I just put a comment in a wrong place and then deleted it :) now you should
see the full review that I pushed.

Le mar. 10 avr. 2018 18:00, Nicolas Chevobbe <notifications@github.com> a
écrit :

PR #1029, opened 2018-04-12
Allow batched worker tasks and generated sourcemap range lookup

loganfsmyth:
Associated Issue: https://github.com/devtools-html/debugger.html/issues/5561

### Summary of Changes

* Allow multiple synchronous calls to an async task to be batched into a single postMessage
* Add a new function for getting the generated ranges for a given original location

### DIFF
--- a/packages/devtools-utils/src/worker-utils.js
+++ b/packages/devtools-utils/src/worker-utils.js
@@ -34,55 +34,83 @@ WorkerDispatcher.prototype = {
     this.worker = null;
   },
 
-  task(method) {
-    return (...args: any) => {
+  task(method, { queue = false } = {}) {
+    const calls = [];
+    const push = (args: Array<any>) => {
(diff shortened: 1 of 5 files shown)


packages/devtools-utils/src/worker-utils.js
```diff
@@ -34,55 +34,83 @@ WorkerDispatcher.prototype = {
...
-    } catch (error) {
-      // Error can't be sent via postMessage, so be sure to convert to
-      // string.
-      self.postMessage({ id, error: error.toString() });
-    }
+    })).then(results => {
+      self.postMessage({ id, results });
+    });
   };
 }
```
loganfsmyth: None of this updates `streamingWorkerHandler` below here because as far as I can tell it is already incompatible with the existing `.task` function in the general case. `.task` only listens for a single message, so it can't be used alongside `streamingWorkerHandler` as is.

PR #1030, opened 2018-04-12
Fix ObjectInspector keyboard navigation

nchevobbe:
The fix here is mainly to call the nodeFocus action when needed in order to changed the `focusedItem` state value.
Also, we let the Tree component handle the call to setExpanded (through `onExpand` and `onCollapse` props), and only stop the propagation of the click event on a tree node for the text selection case.

The objectInspector test formatter was modified in order to reflect whether a node is selected or not, and as a result snapshots were updated.
A simple keyboard navigation test is added to test the basic. I didn't added to many case since this is also tested in the Tree component, and I will add keyboard navigation tests in consumers (at least for the webconsole, and we should also do it in the debugger)

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/test-utils.js
+++ b/packages/devtools-reps/src/object-inspector/tests/test-utils.js
@@ -23,6 +23,7 @@ const {WAIT_UNTIL_TYPE} = require("../../shared/redux/middleware/waitUntilServic
  *
  */
 function formatObjectInspector(wrapper: Object) {
+  const hasFocusedNode = wrapper.find(".tree-node.focused").length > 0;
   return wrapper.find(".tree-node")
     .map(node => {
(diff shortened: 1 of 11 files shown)


packages/devtools-reps/src/object-inspector/tests/test-utils.js
```diff
@@ -38,7 +39,13 @@ function formatObjectInspector(wrapper: Object) {
...
         ? "☲ "
         : "";
-      return `${indentStr}${arrowStr}${icon}${getSanitizedNodeText(node)}`;
+      let text = `${indentStr}${arrowStr}${icon}${getSanitizedNodeText(node)}`;
+      if (!hasFocusedNode) {
+        return text;
+      }
+      return node.hasClass("focused")
+        ? `[ ${text} ]`
+        : `  ${text}`;
```
jasonLaster: this is great

PR #1031, opened 2018-04-12
Wrap text representation of ObjectInspector in new lines.

nchevobbe:
This way tree nodes in snapshot are well aligned.


nchevobbe:
r=me for snapshots small change

PR #1033, opened 2018-04-13
Wrap text representation of Tree in new lines.

nchevobbe:
This way tree nodes in snapshot are well aligned.


nchevobbe:
r=me for trivial snapshot change.

PR #1035, opened 2018-04-17
Cleanup redux state when roots prop changes.

nchevobbe:
The roots prop isn't handled by the redux state, but is passed from
upper-level components (e.g. console, inspector, ...). There might be
cases when the roots prop changes (e.g., in the console, the object sidebar
is already opened, and the user choose to put another object in it).
In such cases, there are a few things that should be taken care of:
- we need to clear the node cache to start anew
- we need to clean the redux state of the old roots-related properties,
  like expandedPaths, loadedProperties, …
- we need to release the actors related to the old roots.
- we need to make sure a possible new focusedItem is handled as expected.

In order to do that, we create a rootsChanged action that's handled in
a middleware where we reinitialize the state and release actors.
We don't want to render before the state is re-initialized, so we add a
new state property, forceUpdate, that we can then use as a flag that we
set when the roots prop changes and reset when the component was updated.

This isn't perfect but I wasn't able to find a better way of handling all
those requirements.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -110,9 +110,22 @@ class ObjectInspector extends Component {
     } = this.props;
 
     if (roots !== nextProps.roots) {
-      // Since the roots changed, we assume the properties did as well. Thus we can clear
-      // the cachedNodes to avoid bugs and memory leaks.
+      // Since the roots changed, we assume the properties did as well, so we need to
+      // cleanup the component internal state.
+
+      // We can clear the cachedNodes to avoid bugs and memory leaks.
       this.cachedNodes.clear();
+      // The rootsChanged action will be handled in a middleware to release the actors
+      // of the old roots, as well as cleanup the state properties (expandedPaths,
+      // loadedProperties, …).
+      this.props.rootsChanged(nextProps);
+      // We don't render right away since the state is going to be changed by the
+      // rootsChanged action. The `state.forceUpdate` flag will be set to `true` so we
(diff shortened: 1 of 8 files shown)


packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -110,9 +110,22 @@ class ObjectInspector extends Component {
...
+
+      // We can clear the cachedNodes to avoid bugs and memory leaks.
       this.cachedNodes.clear();
+      // The rootsChanged action will be handled in a middleware to release the actors
+      // of the old roots, as well as cleanup the state properties (expandedPaths,
+      // loadedProperties, …).
+      this.props.rootsChanged(nextProps);
+      // We don't render right away since the state is going to be changed by the
+      // rootsChanged action. The `state.forceUpdate` flag will be set to `true` so we
+      // can execute a new render cycle with the cleaned state.
```
janodvarko: If I understand correctly, the rendering is postponed to avoid side-bar paint flashing?

nchevobbe: That's the idea. Also, it doesn't only prevent flashing but some bizarre issues that might happen since we are mixing new roots with old state properties.

nchevobbe:
> 'forceUpdate' looks a bit non standard but landable. I am curious if there is a better/standard way how to solve this kind of situation. The rest of the path looks great.

I didn't find anything, and it might be related to the fact that usually you only have a redux app as a whole. Here the ObjectInspector is then called from the outside, which makes it a bit weird in some cases.

PR #1040, opened 2018-04-18
Remove lodash usage from ObjectInspector util.

nchevobbe:
Lodash functions were used in an hot function and was taking significant amount of
time (~80ms). Using plain js does not make the code too bad and shaves of dozens of
millisecond (~25ms).

### DIFF
--- a/packages/devtools-reps/src/object-inspector/utils/node.js
+++ b/packages/devtools-reps/src/object-inspector/utils/node.js
@@ -4,7 +4,6 @@
 
  // @flow
 
-const { get, has } = require("lodash");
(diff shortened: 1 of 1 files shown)


kootoopas:
Darn it, if I had some more confidence with this I would have attempted to change this myself... At a point I noticed that only `get` and `has` was used from `lodash`. 😄 Cool perf boost!

nchevobbe:
@kootoopas yes, the import isn't too bad itself since lodash is vendored-in firefox, but get and has have some costs that we don't really benefit from :) 
Also, if you see something weird in the code, go ahead ;)

PR #1044, opened 2018-04-18
Don't pass onFocus prop to Tree if disabledFocus is true

nchevobbe:
In the end, `this.focusItem` checks `disabledFocus`, so it's not too bad. But the Tree component might use the presence of `onFocus` to do extra work. Let's avoid that.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -222,13 +222,13 @@ class ObjectInspector extends Component {
 
   focusItem(item: Node) {
     const {
-      disabledFocus,
+      focusable = true,
       focusedItem,
       nodeFocus,
       onFocus,
     } = this.props;
 
-    if (!disabledFocus && focusedItem !== item) {
+    if (focusable && focusedItem !== item) {
       nodeFocus(item);
       if (focusedItem !== item && onFocus) {
         onFocus(item);
@@ -476,7 +476,7 @@ class ObjectInspector extends Component {
     const {
       autoExpandAll = true,
       autoExpandDepth = 1,
-      disabledFocus,
+      focusable = true,
       disableWrap = false,
       expandedPaths,
       focusedItem,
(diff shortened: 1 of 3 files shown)


packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -492,7 +492,7 @@ class ObjectInspector extends Component {
 
       onExpand: item => this.setExpanded(item, true),
       onCollapse: item => this.setExpanded(item, false),
-      onFocus: this.focusItem,
+      onFocus: disabledFocus ? null : this.focusItem,
```
kootoopas: Hm, I think it might be clearer if we reversed the boolean logic and turned `disabledFocus` to `focusable` since it'll put the `null` expression second in this case, plus on line 231 it would turn a negated negation (has `not` `dis`abled focus) to a plain statement (can be focused):
```diff
diff --git a/packages/devtools-reps/src/object-inspector/component.js b/packages/devtools-reps/src/object-inspector/component.js
index 900b9bf..89d6c06 100644
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -222,13 +222,13 @@ class ObjectInspector extends Component {
 
   focusItem(item: Node) {
     const {
-      disabledFocus,
+      focusable,
       focusedItem,
       nodeFocus,
       onFocus,
     } = this.props;
 
-    if (!disabledFocus && focusedItem !== item) {
+    if (focusable && focusedItem !== item) {
       nodeFocus(item);
       if (focusedItem !== item && onFocus) {
         onFocus(item);
```

Also, on line 482, `disabledFocus` is passed but not used nor defined as a a prop in `Tree` of `devtools-components`.

 I took the liberty of appending said change with passing tests at my copy of this branch over at https://github.com/kootoopas/devtools-core/tree/onFocus. So if what I'm saying seems valid to you, it's ready to be applied! 😎

nchevobbe: Thanks Footis, it looks like a nice change, can you open a PR with your patch ?
I noticed some snapshots changed, but I think it's because before disabledFocus was false by default, so we need to make focusable true by default.

nchevobbe:
followed @kootoopas recommendation of renaming the prop

PR #1047, opened 2018-04-19
Add an onCmdCtrlKey prop to the ObjectInspector.

nchevobbe:
This will be called with the item an data about it when
the user Ctrl + Click (Cmd on OSX) on a node.
This will be useful in the console in order to have a shortcut
to put objects in the sidebar.

### DIFF
--- a/packages/devtools-reps/src/launchpad/components/Result.js
+++ b/packages/devtools-reps/src/launchpad/components/Result.js
@@ -83,10 +83,15 @@ class Result extends Component {
         createLongStringClient,
         releaseActor,
         mode: MODE[modeKey],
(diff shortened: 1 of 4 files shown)


packages/devtools-reps/src/launchpad/components/Result.js
```diff
@@ -83,6 +83,9 @@ class Result extends Component {
         createLongStringClient,
         releaseActor,
         mode: MODE[modeKey],
+        onCmdCtrlClick: (node, { depth, event, focused, expanded }) => {
+          console.log("CmdCtrlClick", {node, depth, event, focused, expanded});
```
gregtatum: Could you add a comment here explaining that the console.log is for verifying that this functionality works in launchpad? Otherwise I think it's a bit confusing why it's there.

PR #1051, opened 2018-04-19
Bump launchpad 119

jasonLaster:


packages/devtools-launchpad/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "devtools-launchpad",
-  "version": "0.0.117",
+  "version": "0.0.119",
```
nchevobbe: why do we skip 118 ?

jasonLaster: i missed

jasonLaster: forgot to do it

PR #1052, opened 2018-04-19
Speed up containsURL function.

nchevobbe:
The function was splitting up a string and testing if each parts was an actual URL,
until one was encountered.
The split function was showing up in some profile I did today (nothing too big, but
bigger that it should be).
Here switching to a simple regex test works and is faster to run.

### DIFF
--- a/packages/devtools-reps/src/reps/rep-utils.js
+++ b/packages/devtools-reps/src/reps/rep-utils.js
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // Dependencies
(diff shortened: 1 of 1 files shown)


packages/devtools-reps/src/reps/rep-utils.js
```diff
@@ -359,12 +359,12 @@ function getGripType(object, noGrip) {
  *         Whether the grip is a string containing a URL.
  */
 function containsURL(grip) {
-  if (typeof grip !== "string") {
+  // An URL can't be shorter than 5 char.
```
janodvarko: Maybe an example of such short URL would be nice to have in the comment.

PR #1063, opened 2018-05-04
Create devtools-environment

jasonLaster:
This dramatically simplifies devtools-config so that it is just launchpad config stuff like the port, favicon, and connection logging.

This introduces a simple runtime environment check


packages/devtools-environment/index.js
```diff
@@ -0,0 +1,30 @@
...
+const flag = require("./test-flag");
+
+function isBrowser() {
+  return typeof window == "object";
+}
+
+export function isDevelopment() {
+  if (isBrowser()) {
+    const href = window.location ? window.location.href : "";
+    return href.match(/^file:/) || href.match(/localhost:/);
```
wldcordeiro: I know you just copied over but couldn't this be achieved with 1 match call since regex supports or?

jasonLaster: true... but i think this is easier to read

packages/devtools-environment/index.js
```diff
@@ -0,0 +1,34 @@
...
+function isBrowser() {
+  return typeof window == "object";
+}
+
+function isNode() {
+  return process && process.release && process.release.name == 'node'
+}
+
+export function isDevelopment() {
+  if (!isNode && isBrowser()) {
```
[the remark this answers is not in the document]
jasonLaster: oh oops...

PR #1064, opened 2018-05-04
Only render the SplitBox draggable when nothing is collapsed.

lijunle:
Associated Issue:

The debugger is showing scrollbar when collapse the panes.

image

Besides, currently, when the pane is collapsed, hover on the edge (dragger), the cursor is still showing as "draggable", which is confusing.

### Summary of Changes

* It should not render the dragger when any pane is collapsed.

### Test Plan

I FAILED TO RUN ANY TEST. THE DEV IS VERY NOT FRIENDLY. 😞 


jasonLaster:
Thanks. I'll take a look!

PR #1066, opened 2018-05-18
Create DevTools Services

jasonLaster:
### Summary of Changes

Services maps to Services.jsm in MC and we want to be able to use the real module in MC. 

We currently do the mapping in [webpack.config.devtools.js][wc]. You can see how [vendors.js][v] pulls in devtools-modules, but excludes Services.

It might be nice to simplify the process a bit by creating a package for it in core, we can require it directly in the debugger and reps and then at release time map it to the real thing. I think this system is marginally better, but perhaps not needed...

[wc]: https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/webpack.config.devtools.js#L28-L32
[v]: https://searchfox.org/mozilla-central/source/devtools/client/debugger/new/dist/vendors.js


packages/devtools-services/README.md
```diff
@@ -0,0 +1,13 @@
+## Services
+
+[![Npm version](https://img.shields.io/npm/v/devtools-services.svg)](https://npmjs.org/package/devtools-services)
+
+Shims [Services.jsm](https://searchfox.org/mozilla-central/source/toolkit/modules/Services.jsm)
+
+* `prefs`
+* `Services`
```
nchevobbe: I think we only export the method of `Services` and not `{Services, prefs}`

nchevobbe:
and circle is ✅

packages/devtools-services/README.md
jasonLaster: that makes sense

jasonLaster:
merge at will

PR #1071, opened 2018-06-22
Upgrade event emitter

jasonLaster:
### Summary of Changes

* We should try and upgrade the devtools-connection event emitter so that we are consistent w/ m-c. 
* This patch has some release candidate versions, but does mostly work with the debugger

#### What Doesn't Work

For some reason, `client.fetchWorkers` does not work. But if you comment it out everything seems to work fine

```diff
diff --git a/src/actions/debuggee.js b/src/actions/debuggee.js
index 6a52fae..a93ea2b 100644
--- a/src/actions/debuggee.js
+++ b/src/actions/debuggee.js
@@ -8,7 +8,7 @@ import type { Action, ThunkArgs } from "./types";

 export function updateWorkers() {
   return async function({ dispatch, client }: ThunkArgs) {
-    const { workers } = await client.fetchWorkers();
-    dispatch(({ type: "SET_WORKERS", workers }: Action));
+    // const { workers } = await client.fetchWorkers();
+    // dispatch(({ type: "SET_WORKERS", workers }: Action));
   };
 }
(END)
```

#### How to test

1. run `yarn add -W devtools-launchpad@0.0.126-rc9; yarn;` in the debugger
2. run `yarn start`

everything should just work at this point. There might be other edge cases like `fetchWorkers` but i couldnt find them, and the mochitests will just work because they use mc.

### DIFF
--- a/packages/devtools-connection/src/utils/event-emitter.js
+++ b/packages/devtools-connection/src/utils/event-emitter.js
@@ -2,138 +2,259 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-/**
- * EventEmitter.
- */
+"use strict";
 
-var EventEmitter = function EventEmitter() {};
-module.exports = EventEmitter;
+const BAD_LISTENER = "The event listener must be a function, or an object that has " +
+                     "`EventEmitter.handler` Symbol.";
 
-const promise = require("./promise");
+const eventListeners = Symbol("EventEmitter/listeners");
+const onceOriginalListener = Symbol("EventEmitter/once-original-listener");
+const handler = Symbol("EventEmitter/event-handler");
 
-/**
- * Decorate an object with event emitter functionality.
- *
- * @param Object aObjectToDecorate
- *        Bind all public methods of EventEmitter to
- *        the aObjectToDecorate object.
- */
-EventEmitter.decorate = function EventEmitter_decorate(aObjectToDecorate) {
-  let emitter = new EventEmitter();
-  aObjectToDecorate.on = emitter.on.bind(emitter);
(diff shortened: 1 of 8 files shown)


jasonLaster:
@wldcordeiro this will need some review, and we should make sure that it works in the debugger before landing

packages/devtools-connection/src/utils/event-emitter.js
```diff
@@ -2,138 +2,259 @@
...
+                     "`EventEmitter.handler` Symbol.";
+
+const eventListeners = Symbol("EventEmitter/listeners");
+const onceOriginalListener = Symbol("EventEmitter/once-original-listener");
+const handler = Symbol("EventEmitter/event-handler");
+
+const isEventHandler = (listener) =>
+  listener && handler in listener && typeof listener[handler] === "function";
+
+class EventEmitter {
```
codehag: we should add a comment that this file is adapted from https://searchfox.org/mozilla-central/source/devtools/shared/event-emitter.js lines 0 - 259, and that it should be kept in sync with it. We can also add that logging and devtools.dump.emit functionality has been removed

packages/devtools-connection/src/utils/event-emitter.js
```diff
@@ -2,138 +2,259 @@
...
+        resolve(first);
+      };
+
+      newListener[onceOriginalListener] = listener;
+      EventEmitter.on(target, type, newListener);
+    });
+  }
+
+  static emit(target, type, ...rest) {
+    if (!(eventListeners in target)) {
```
codehag: Logging is disabled by default, but I'm wondering if it makes sense to keep it in. The code doesn't look too intrusive. Would make keeping this file in sync easier, as then it is almost a direct copy. The only thing that stands out in that case is services. If we decide to keep it out, lets add to the comment that logging has been removed.

PR #1074, opened 2018-07-06
Upgrade to React 16.4.1

jasonLaster:


jasonLaster:
> react-dom-factories

I think we'll likely just move to jsx pretty soon. Babel should already be parsing it. 

> storybook

We're not using storybook. This was setup because last year we were using percy.io. I removed it becasue storybook was pulling down another version of react and it seemed heavy.

jasonLaster:
oh, some context for react-dom-factories, they are no longer supported. I'd keep it if i could

jasonLaster:
oh... maybe i was wrong :)

PR #1075, opened 2018-07-06
Allow Configuration of Packages to Process

ahuth:
### Summary of Changes

For https://github.com/devtools-html/debugger.html/pull/6595, we need babel to still compile react-aria-components.

As a part of that, this slightly refactors how the regex doing the matching is constructed.

### DIFF
--- a/packages/devtools-launchpad/webpack.config.js
+++ b/packages/devtools-launchpad/webpack.config.js
@@ -41,7 +41,7 @@ module.exports = (webpackConfig, envConfig, options = {}) => {
 
       // Some of the excluded items need to be re-included for processing.
       const reincludePaths = ["devtools-", "react-aria-components"];
-      const reincludeRe = new RegExp(`node_modules(\/|\\)${reincludePaths.join("|")}`);
+      const reincludeRe = new RegExp(`node_modules(\\/|\\\\)${reincludePaths.join("|")}`);
 
       return excluded && !request.match(reincludeRe);
     },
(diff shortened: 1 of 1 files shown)


packages/devtools-launchpad/webpack.config.js
```diff
@@ -38,7 +38,12 @@ module.exports = (webpackConfig, envConfig, options = {}) => {
         // If the tool defines an additional exclude regexp for Babel.
         excluded = excluded || !!request.match(options.babelExcludes);
       }
-      return excluded && !request.match(/node_modules(\/|\\)devtools-/);
+
+      // Some of the excluded items need to be re-included for processing.
+      const reincludePaths = ["devtools-", "react-aria-components"];
```
jasonLaster: Lets make this a parameter so it is easy for others to extend in the future

jasonLaster:
I just pushed a change that adds a new option for included react-aria-components. 
It's published here at `devtools-launchpad@0.0.127-rc2` so you can test it out

debugger.html

```diff
diff --git a/webpack.config.js b/webpack.config.js
index 8638d95..72c2a77 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -76,6 +76,8 @@ function buildConfig(envConfig) {

     extra.excludeMap = mozillaCentralMappings;

+    extra.included = ["react-aria-components"];
+
     mappings.forEach(([regex, res]) => {
       webpackConfig.plugins.push(new NormalModuleReplacementPlugin(regex, res));
     });
```

PR #1082, opened 2018-08-22
Change the splitter's variable name since m-c has changed these name.

mantaroh:
Associated Issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1465644

### Summary of Changes

* Change the size of SplitBox min-height or min-width.
  - I added 1px extra width to the splitter of m-c and renamed the related variable.

* Change the background color of the SplitBox.
* Add hover style to the splitter.

This commit changes the style only.


jasonLaster:
Thanks!

PR #1019, opened 2018-03-30
removed --theme-search-overlays-semitransparent to fix #5566

irrationalRock:
Associated Issue: https://github.com/devtools-html/debugger.html/issues/5566

### Summary of Changes

* I removed --theme-search-overlays-semitransparent in Root.css.

PR #1084, opened 2018-08-27
Remove unused theme-search-overlays-semitransparent

juliandescottes:
There is still an unresolved comment on #1019 so let's move this to another PR


juliandescottes:
Steps after merging this:
[ ] release new launchpad on npm
[ ] update launchpad dependency on debugger
[ ] during the release of debugger that includes this new launchpad version, the test browser_parsable_css.js will fail and will need to be updated (just remove the mention of the variable deleted here)

PR #1087, opened 2018-09-13
feat: Add accelerator support for panel

jbhoosreddy:
Associated PR: https://github.com/devtools-html/debugger.html/pull/6970
### Summary of Changes

* Add support for accelerator in-panel

### Test Plan

Example test plan:

- [x] Command-P opens the panel
- [x] Clicking “+” opens the panel
- [x] Clicking a source navigates to the source
- [x] Clicking "x" closes the panel


packages/devtools-contextmenu/package.json
```diff
@@ -9,6 +9,10 @@
   "author": "Jason Laster",
   "license": "MPL-2.0",
   "dependencies": {
-    "devtools-modules": "^0.0.38"
+    "autoprefixer": "^9.1.5",
+    "devtools-modules": "^0.0.38",
+    "postcss-bidirection": "^2.7.0",
+    "postcss-class-namespace": "^0.1.0",
+    "postcss-url-mapper": "^1.2.0"
```
jasonLaster: why do we need to add these css tools?

jbhoosreddy: @jasonLaster if I remove the postcss tools debugger fails to build. Can we let them in?

PR #1094, opened 2018-09-21
Popupset element for context-menu doesn't have to exist (see also bug…

janodvarko:
Related to XUL removal.

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1476097

Honza


jasonLaster:
@janodvarko could you look into why this is causing test failures?

PR #1101, opened 2018-10-12
Update devtools-connection to 0.0.13

nchevobbe:


nchevobbe:
self r+ for version bump

Issue #1095, opened 2018-09-22
TypeError: appRoot is null


darkwing:
When changing the theme between light and dark, then dark to light, after the 2nd or 3rd change the following error is shown:

```
TypeError: appRoot is null[Learn More] index.js:41:4 
```

...and the theme will no longer change.

artemruts:
@darkwing I'd like to work on this.
Is it issue claimable?

PR #1112, opened 2018-12-15
Fixes launchpad theme switching logic #1095.

artemruts:
Associated Issue: #1095

### Summary of Changes

When Launchpad theme is switched only the one theme class (`theme-dark`, `theme-light`, ...) is switched on the `root` and `appRoot` elements.

This way we don't replace the whole class string and get to keep other classes that were set on the element before.

E.g. Currently `html` class disappears after the theme is switched.
https://github.com/devtools-html/devtools-core/blob/c4fa6097a69c7b5d2156070532cc5b9374713040/packages/devtools-launchpad/src/index.js#L57

### Notes

Not the best solution and can be 100% improved if 😢 
1. there was a unique theme prefix (`active-theme-` or something like that) then we could just replace the class without filtering;
2. or even better `classList` wasn't manipulated directly and we had a `Theme` react component / provider.

I can take a stab at one of those ideas (preferably number 2) if the fix is too ugly.

### Test Plan

- [x] reproduced the error
- [x] made code changes
- [x] switched launchpad theme multiple times without an error 
- [x] checked the DOM and verified that theme classes are correct

### Screenshots/Videos

themes

### DIFF
--- a/packages/devtools-launchpad/src/index.js
+++ b/packages/devtools-launchpad/src/index.js
@@ -16,6 +16,11 @@ const { showMenu, buildMenu } = require("devtools-contextmenu");
 
 setConfig(DebuggerConfig);
 
+// List of theme class names to ignore when replacing a theme.
+// Certain nodes may contain multiple "theme-" classes and some of them we must
+// preserve.
+const IGNORE_THEME_CLASS_NAMES = ["theme-body"];
+
 // Set various flags before requiring app code.
 if (getValue("logging.client")) {
   // DevToolsUtils.dumpn.wantLogging = true;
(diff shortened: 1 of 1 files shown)


AnshulMalik:
Looks good to me

packages/devtools-launchpad/src/index.js
```diff
@@ -31,17 +36,35 @@ if (process.env.TARGET !== "firefox-panel") {
...
+  // If one of the arguments is missing - bail out.
+  if (!targetNode || !newThemeClassName) {
+    return;
+  }
+
+  const {classList} = targetNode;
+
+  // Attempt to find the first theme class, assuming there's only one and
+  // ignoring class names from IGNORE_THEME_CLASS_NAMES array.
+  const activeThemeClassName = [...classList].find(cls =>
```
bomsy: nice work!  Doing a `find` might be too much. This should do the switch
```suggestion
  const activeThemeClassName = newThemeClassName == "light" ? "dark" : "light";
  // the class might almost always exist and if the class does 
  // not exist remove wont throw and error so should be fine
  classList.remove(`theme-${activeThemeClassName}`); 
  classList.add(`theme-${newThemeClassName}`)
```

packages/devtools-launchpad/src/index.js
```diff
@@ -31,17 +36,35 @@ if (process.env.TARGET !== "firefox-panel") {
...
+    classList.add(newThemeClassName);
+  }
+}
+
 function updateTheme(className) {
   if (process.env.TARGET !== "firefox-panel") {
-    const theme = getValue("theme");
     const root = document.body.parentNode;
     const appRoot = document.querySelector(".launchpad-root");
+    const newThemeClassName = `theme-${getValue("theme")}`;
```
bomsy:
```suggestion
    const newThemeClassName = getValue("theme");
```

bomsy:
@artemruts Nice work! left some suggestions!

packages/devtools-launchpad/src/index.js
artemruts: Thanks, for the feedback.
I used `find` because if somebody will add a new theme (like `camo`) a simple ternary will not work there. Not sure if we're going to support more than a few themes.

bomsy: Good point...I was working with the assumption we’ll always have light or dark, since we’ve only had those, but u r right to support future themes.
Nice work! 👍🏾

bomsy:
LGTM!

PR #1122, opened 2019-01-15
Added + to formatKeyShortcut to fix issue #7652 in Debugger.html

bamanczak:
Associated Issue: [#7652 (in debugger.html)](https://github.com/devtools-html/debugger.html/issues/7652)

### Summary of Changes

* added + in formatKeyShortcut method for Windows shorcuts formatting

### Test Plan
Manual tests

- [x] Verify if "acceltext" property contains proper shortcut (i.e. with "+" on Windows)


jasonLaster:
Thanks @bamanczak and sorry for the late reply


We should try and make sure we're the same as what's in firefox.
https://searchfox.org/mozilla-central/source/devtools/client/debugger/new/src/utils/text.js#39

It looks like what is there is more modern...

bamanczak:
To be honest I thought of removing the method referenced by you in the next step, and leave the devtools-core one, as currently we simply duplicate the method for formatting the shortcuts (and on top of that there is an error with adding spaces on MacOS, Apple is formatting shortcuts with spaces).

But I assume that apart from the spaces on MacOS (which in my view is a bug) you are refering to the localisation method for ctrl: `{L10N.getStr("ctrl")}` ?
Any idea how to implement that in devtools-core? I do not see such a mechanism right now :(

jasonLaster:
@bamanczak lets discuss in slack https://devtools-html-slack.herokuapp.com/

i'm a bit lost...

bamanczak:
Hi! @jasonLaster I updated the Pull Request as per your comment. I hope I moved the function correctly. I verified that the fix was working correctly.

jasonLaster:
Thanks!

PR #1126, opened 2019-02-04
Update client for changes in bug 990137, bug 1520993, bug 1520972.

bhackett1024:
bug 990137, bug 1520993, and bug 1520972 made changes to the m-c/devtools/shared clients.  This PR ports these changes to the client implementation in devtools-connection.

### Test Plan

I still need to test this patch manually.


bhackett1024:
With the above fix normal breakpoints can be set in the debugger.  Conditional breakpoints and logpoints will not work until the commands.js changes from bug 1520972 are ported to debugger.html from m-c.

PR #1148, opened 2019-05-06
Synchronize context menu changes needed for Bug 1543940

juliandescottes:
Menu API should now take a document argument and there is additional logic to hide the context menu when its parent window is hidden.

The corresponding bugzilla record is https://bugzilla.mozilla.org/show_bug.cgi?id=1543940
It is a mandatory change to run DevTools in frame with type=content (https://bugzilla.mozilla.org/show_bug.cgi?id=1539979) which is required for the overall fission effort.

### DIFF
--- a/packages/devtools-contextmenu/menu.js
+++ b/packages/devtools-contextmenu/menu.js
@@ -92,12 +92,12 @@ function showMenu(evt, items) {
     });
 
   if (inToolbox()) {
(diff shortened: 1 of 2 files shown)


jasonLaster:
I'm happy to do a new release with this change tomorrow.

jasonLaster:
okay - feel free to bump to launchpad 152 in the debugger directory

juliandescottes:
thanks @jasonLaster !

PR #1151, opened 2019-05-10
Use createXULElement to create popup items

juliandescottes:
Sync with https://bugzilla.mozilla.org/show_bug.cgi?id=1550827


juliandescottes:
let's abandon this.

PR #1152, opened 2019-05-14
Require Menu and MenuItem with full path to override for mc bundle

juliandescottes:
This will allow us to add two mappings to mozilla-central-mappings:
```javascript
    "devtools-modules/src/menu": "devtools/client/framework/menu",
    "devtools-modules/src/menu-item": "devtools/client/framework/menu-item",
```

After this change, the debugger will use the regular menu implementation when running in the toolbox.
There is a lot of cleanup possible between devtools-contextmenu and devtools-modules/src/menu after that, since they are only used when running in content, but I'd leave that for later.


juliandescottes:
@jasonLaster any chance you can review this?

PR #1153, opened 2019-05-17
Fix inToolbox helpers to detect about:devtools-toolbox URLs

juliandescottes:
Sync with https://bugzilla.mozilla.org/show_bug.cgi?id=1552444

"about:devtools-toolbox" is only valid for the default toolbox. We can also have additional parameters to modify the target, the runtime etc...


juliandescottes:
Sync with https://bugzilla.mozilla.org/show_bug.cgi?id=1552444

PR #1155, opened 2019-05-23
fix typo in path to MenuItem

jfo:
Looks like the path was truncated in https://github.com/firefox-devtools/devtools-core/commit/b63607474ea8edbc3d3c2ebfdc78cff4354453db

Noticed on webpack failure while building debugger.html.


jasonLaster:
oh snap. This was changed last week, but not merged.

PR #1156, opened 2019-05-23
add isDevelopment and isEnabled to feature.js

jfo:
These seemed to be missing and the associated tests were failing.

Returning booleans.


jasonLaster:
nice catch

PR #1195, opened 2020-06-30
Add a few more aliases for MPL2 and Apache 2

julienw:
I found these (invalid) values in some existing packages that we use, some of them transitively, in the profiler server. This would be a bit painful to make them all update...


nchevobbe (approved):
That looks good to me, thanks Julien :)

PR #1209, opened 2020-08-20
Bump version to 0.0.2 and fix URLs in package.json

nchevobbe:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "devtools-core",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
julienw, line 3: devtools-core itself isn't deployed anywhere, I don't think this is necessary to update this version. But I don't mind :)

julienw (approved):

package.json
nchevobbe: reverted :)

PR #1217, opened 2020-09-29
Remove devtools-license-check

nchevobbe:
The package was moved to its own repo: https://github.com/firefox-devtools/devtools-license-check


julienw (approved):
thanks for the cleanup :+1:

